PR #14, opened 2018-08-03
Add job resource

wslulciuc:
This PR adds the following endpoints:

* `POST` /jobs -creates a job
* `GET` /jobs - returns a list of jobs (option to filter by `created_at`)
* `GET` /jobs/**:job_name** - retrieves a job
* `PUT` /jobs/**:job_name** - update a job


src/main/resources/db/migration/V1__create_tables.sql
```diff
@@ -57,7 +57,7 @@ CREATE TABLE job_run_states (
 
 CREATE TABLE datasets (
   id              SERIAL PRIMARY KEY,
-  name            VARCHAR(64) NOT NULL,
+  name            VARCHAR(64) UNIQUE NOT NULL,
```
sshah-wework: is `name` meant to be fairly free-form, or will it get standardized to something like NAMESPACE.NAME

sshah-wework: another question for thought, should the uniqueness of a dataset include it's db as well? <db_id, name>. Or if namespace is more explicit: <db_id, db_namespace, name>. Just ideas :)

src/main/java/marquez/db/dao/JobDAO.java
```diff
@@ -1,14 +1,64 @@
...
+  OwnershipDAO createOwnershipDAO();
+
+  default void insert(final Job job) {
+    try (final Handle handle = getHandle()) {
+      handle.useTransaction(
+          h -> {
+            final int jobId =
+                h.createUpdate(
+                        "INSERT INTO jobs (name, nominal_time, category, description)"
+                            + " VALUES (:name, :nominalTime, :category, :description)")
```
sshah-wework: Will this have update behavior instead of insert if a row with the same primary key is already present?

src/main/java/marquez/resources/JobResource.java
```diff
@@ -1,26 +1,55 @@
...
+    jobDAO.insert(job);
+    return Response.created(buildLocation(Job.class, job.getName())).build();
+  }
 
-  public JobResource(JobDAO dao) {
-    this.dao = dao;
+  @GET
+  @Timed
+  public Response getAll() {
+    return Response.ok(jobDAO.findAll()).build();
```
sshah-wework: Not something to do now, but what do you think about pagination on this in the future? Could this response get pretty large?

src/main/java/marquez/resources/JobResource.java
```diff
@@ -1,26 +1,55 @@
...
   }
 
   @GET
-  public List<Job> listJobs() {
-    return dao.findAll();
+  @Path("/{jobName}")
+  @Timed
+  public Response get(@PathParam("jobName") final String jobName) {
+    return Response.ok(jobDAO.findByName(jobName)).build();
```
sshah-wework: Could there be an error case to handle here? For instance, if there's no job found should it be an HTTP 404?

sshah-wework: For this and other endpoints?

src/main/java/marquez/db/dao/JobDAO.java
```diff
@@ -1,14 +1,64 @@
...
+                    .findOnly();
+            final int ownershipId =
+                createOwnershipDAO()
+                    .insert(
+                        new Ownership(
+                            new Timestamp(System.currentTimeMillis()),
+                            Optional.empty(),
+                            jobId,
+                            ownerId));
+            updateOwnership(ownershipId, jobId);
```
sshah-wework: how do you feel about moving the above SELECT + update into an `updateOwnershipByName` method? that way the JobDAO has very little knowledge of the owners data.

src/main/resources/db/migration/V1__create_tables.sql
wslulciuc: > is name meant to be fairly free-form ...

I think simple rules around name validation would make sense (ie allowed characters). We'll want dataset names to be human-readable / fairly free-form, but leaving general naming conventions up to the owner of the dataset.

> or will it get standardized to something like NAMESPACE.NAME ...

We'll want to standardize on some internal representations of datasets. Let's say we have an owner `analytics`. There should be only a single owner registered with marquez under that name. But datasets names shouldn't be globally unique, though unique to the owner.

**Valid**
*  `analytics.room_bookings`
*  `datascience.room_bookings`

sshah-wework: Cool, sounds good. 👍

wslulciuc: > another question for thought, should the uniqueness of a dataset include it's db as well? <db_id, name>. Or if namespace is more explicit: <db_id, db_namespace, name>

Agreed. Defining a tuple that uniquely identifies DBs totally makes sense (or at least some variation of what you are proposing).

src/main/java/marquez/db/dao/JobDAO.java
wslulciuc: For convenience, I added `JobDAO.updateOwnership()`, but it probably doesn't make sense for a job to update it's ownership. We can simply have the `OwnershipDAO` update `job.current_ownership.` and handle all the logic around changing ownership over time.

src/main/java/marquez/db/dao/JobDAO.java
```diff
@@ -1,14 +1,64 @@
...
+            h.commit();
+          });
+    } catch (Exception e) {
+      LOG.error(e.getMessage());
```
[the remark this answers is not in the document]
wslulciuc: We'll want to define API-specific exceptions. For example, `APIException` as a base API exception class, `JobException`, `InvalidRequestException` and so on

src/main/java/marquez/resources/JobResource.java
wslulciuc: Yep, it sure can. I'll add an option to limit results (with a reasonable default). So something like:

```
GET /jobs?limit=25
```

src/main/java/marquez/resources/JobResource.java
wslulciuc: Yeah, we'll want to come up with a strategy to add general exception handling across all endpoints.

src/main/java/marquez/db/dao/JobDAO.java
wslulciuc: +1

sshah-wework: cool, I'll add that in my PR for the ownership DAO

sshah-wework:
Sounds great. Love seeing the codecov report!

PR #15, opened 2018-08-03
first DAO test with h2 db setup

julienledem:


build.gradle
```diff
@@ -16,9 +16,6 @@ repositories {
   mavenCentral()
 }
 
-if (JavaVersion.current() != JavaVersion.VERSION_1_8) {
-  throw new GradleException("This build must be run with Java 8")
-}
```
julienledem: do we need this? I have java 9

wslulciuc: Actually, we can remove. This check was more for guarding against compatibility changes and ensure we build with a stable version (but java 9 is fairly stable).

build.gradle
```diff
@@ -31,6 +28,7 @@ dependencies {
   testCompile 'io.dropwizard:dropwizard-testing:1.3.5'
   testCompile 'junit:junit:4.12'
   testCompile 'org.mockito:mockito-core:2.19.0'
+  testCompile 'com.h2database:h2:1.4.197'
```
wslulciuc: Did we want to test alternative DB types that we currently don't support? The DB [schema](https://github.com/MarquezProject/marquez/blob/master/src/main/resources/db/migration/V1__create_tables.sql) is currently defined using postgres. We also config [postgres](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/MarquezApplication.java#L56) at runtime and don't allow for other DB types (at least currently).

src/test/java/marquez/db/dao/TestOwnerDAO.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import marquez.api.Owner;
+import marquez.db.dao.fixtures.DAOSetup;
+
+public class TestOwnerDAO {
+
+	@ClassRule
+	public static final DAOSetup daoSetup = new DAOSetup();
```
wslulciuc: I'm wondering if we can simplify the app setup with:

```java
   @ClassRule
   static final DropwizardAppRule<MarquezConfiguration> RULE =
     new DropwizardAppRule<MarquezConfiguration>(
       MarquezApplication.class,
       ResourceHelpers.resourceFilePath("config.yaml"));

```

source: http://dropwizard.readthedocs.io/en/stable/manual/testing.html

build.gradle
wslulciuc: Ahh, overlooked `testCompile`. What are you're thoughts on using [testcontainers](https://www.testcontainers.org/) instead of H2?

src/test/java/marquez/db/dao/TestOwnerDAO.java
wslulciuc: If we decided to use testcontainers, the rule would be defined simply as:

```java
 @Rule
public PostgreSQLContainer postgres = new PostgreSQLContainer();
```

source: [`SimplePostgreSQLTest`](https://github.com/testcontainers/testcontainers-java/blob/master/modules/jdbc-test/src/test/java/org/testcontainers/junit/SimplePostgreSQLTest.java)

build.gradle
julienledem: I was hoping that h2 would be compatible enough. I did not know about testcontainers, but as long as the setup is integrated in the test (with a  Rule) then it looks great. We just need to verify it what dependencies are (docker installed etc). The fewer the better.

src/test/java/marquez/db/dao/TestOwnerDAO.java
```diff
@@ -0,0 +1,21 @@
+package marquez.db.dao;
+
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import marquez.api.Owner;
+import marquez.db.dao.fixtures.DAOSetup;
+
+public class TestOwnerDAO {
```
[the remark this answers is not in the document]
julienledem: SGTM.

src/test/java/marquez/db/dao/TestOwnerDAO.java
julienledem: The DAOSetup class is trying to simplify further the DropwizardAppRule, since we know the Config and App classes, have a config file etc. I started with the DARule.

src/test/java/marquez/db/dao/TestOwnerDAO.java
```diff
@@ -0,0 +1,21 @@
...
+
+	@ClassRule
+	public static final DAOSetup daoSetup = new DAOSetup();
+
+	final OwnerDAO ownerDAO = daoSetup.onDemand(OwnerDAO.class);
+
+	@Test
+	public void testUpdateOwner() {
+		ownerDAO.insert(new Owner("Amaranta"));
```
ashulmanWeWork: We should verify the result.

build.gradle
sshah-wework: Should we be using Java 9 with this project going forward?

sshah-wework:
Is the idea that we merge this as the empty test suite and we can work on filling in with tests in future PRs?

src/test/java/marquez/db/dao/TestOwnerDAO.java
julienledem: agreed :) The DAO doesn't have a "get" yet so I stopped there. It doesn't fail which mean that the db is setup properly. I wanted to have the DB setup to bootstrap DAO testing.

julienledem:
Yes, the idea was to bootstrap DAO test by setting up the DB

sshah-wework:
Ok perfect.

build.gradle
wslulciuc: Sure. Java 9 is fairly stable anyways. Also comes with convenient factory methods for collections!

src/test/java/marquez/db/dao/fixtures/DAOSetup.java
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	/**
+	 * @see Jdbi#onDemand(Class)
+	 */
+	public <T> T onDemand(Class<T> classT) {
+		return jdbi.onDemand(classT);
+	}
+}
+
```
wslulciuc: 💯

build.gradle
wslulciuc: There is a dependency on docker, which does add _one more thing_ to get up and running. That said, for integration tests, that'd be fine and expected.

PR #21, opened 2018-08-12
replace compile with implementation

lulciuca:
`compile` and `testCompile` are deprecated as of gradle 3

https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#new_configurations

Since all of the dependencies used in the project are for internal use only (this is an app not a lib), I've updated the configuration to use `implementation` and `testImplementation`.


wslulciuc:
Thanks!

PR #27, opened 2018-08-13
implement get, delete for owner and improve tests

sshah-wework:
Update and delete operations for /owner and /ownership need to be implemented in the DAOs and Resources


src/main/java/marquez/db/dao/OwnerRow.java
```diff
@@ -0,0 +1,14 @@
...
+import marquez.api.Owner;
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.statement.StatementContext;
+
+public class OwnerRow implements RowMapper<Owner> {
+  @Override
+  public Owner map(ResultSet rs, StatementContext ctx) throws SQLException {
+    return new Owner(rs.getString("name"));
+  }
+}
```
wslulciuc: 💯

src/main/java/marquez/resources/OwnerResource.java
```diff
@@ -25,4 +27,11 @@ public Response create(final Owner owner) {
     dao.insert(owner);
     return Response.created(buildURI(Owner.class, owner.getName())).build();
   }
+
+  @Path("{owner}")
+  @DELETE
+  @Consumes(APPLICATION_JSON)
```
wslulciuc: We can remove `@Consumes()` annotation given that no Json payload is provided along with the request.

src/main/java/marquez/db/dao/OwnerDAO.java
```diff
@@ -1,10 +1,20 @@
...
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.customizer.BindBean;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 
+@RegisterRowMapper(OwnerRow.class)
 public interface OwnerDAO {
+  @SqlQuery("SELECT * FROM owners WHERE name = :name")
+  Owner findByName(@Bind("name") String name);
```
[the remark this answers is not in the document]
sshah-wework: totally! I missed that, will add.

src/main/java/marquez/resources/OwnerResource.java
sshah-wework: nice

src/test/java/marquez/db/dao/TestOwnerDAO.java
```diff
@@ -10,10 +15,51 @@
...
+    List<String> deletedAts =
+        daoSetup
+            .getJDBI()
+            .withHandle(
+                handle ->
+                    handle
+                        .createQuery("SELECT deleted_at FROM owners WHERE name = :name")
+                        .bind("name", name)
+                        .mapTo(String.class)
+                        .list());
```
wslulciuc: Use `findOnly()` instead of `list()` to return a single user name given that only one result is possible. Also, maybe we want to bind the result to an `Owner` object instead of a string?

sshah-wework:
Feedback addressed. Ready to merge.

PR #30, opened 2018-08-22
Add circleci config and coveralls badge

wslulciuc:


.circleci/config.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/marquez
+    docker:
+      - image: circleci/openjdk:8-jdk
```
sshah-wework: should we standardize on java 8 for development?

wslulciuc: Actually, let's bump this to `java 9`. And are you thinking enforcing `java 9`?

wslulciuc:
```java
if (JavaVersion.current() != JavaVersion.VERSION_1_9) {
  throw new GradleException("This build must be run with Java 9")
}
```

sshah-wework: I think some warning/enforcement is good just so that there are no surprises when the code builds locally but fails on the build server.

PR #42, opened 2018-08-26
Add db migration on app startup

wslulciuc:
The PR adds an integration test for [MarquezApp](https://github.com/MarquezProject/marquez/blob/migrate-db/src/main/java/marquez/MarquezApp.java) and enables DB migrations on app startup.


src/main/java/marquez/MarquezApp.java
```diff
@@ -15,16 +15,22 @@
...
 import marquez.resources.PingResource;
+import org.flywaydb.core.Flyway;
+import org.flywaydb.core.api.FlywayException;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.postgres.PostgresPlugin;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MarquezApp extends Application<MarquezConfig> {
```
sshah-wework: why the rename to `MarquezApp`? Just for convenience?

src/main/java/marquez/MarquezApp.java
```diff
@@ -33,23 +39,48 @@ public String getName() {
...
+    // Attempt to perform a database migration. An exception is thrown on failed migration attempts
+    // requiring we handle the throwable and apply a repair on the database to fix any
+    // issues before terminating.
+    try {
+      flyway.migrate();
+    } catch (FlywayException e) {
+      LOG.error("Failed to apply migration to database.", e.getMessage());
+      flyway.repair();
+
+      LOG.info("Successfully repaired database, stopping appication...");
```
sshah-wework: s/appication/application/

src/main/java/marquez/MarquezApp.java
```diff
@@ -33,23 +39,48 @@ public String getName() {
...
+      // Next, we'll signal appication termination and not propagating the throwable
```
sshah-wework: s/appication/application

src/test/java/marquez/MarquezAppIntegrationTest.java
```diff
@@ -0,0 +1,29 @@
...
+    final Response res =
+        APP.client()
+            .target(URI.create("http://localhost:" + APP.getLocalPort()))
+            .path("/ping")
+            .request()
+            .get();
+    assertEquals(res.getStatus(), 200);
+    assertEquals(res.readEntity(String.class), "pong");
+  }
+}
```
sshah-wework: this is nice, although not sure it's relevant to DB migrations on startup?

src/main/java/marquez/MarquezApp.java
wslulciuc: Yeah, partially for convenience and also just a habit of mine to abbreviate when possible. (I initially referenced the dropwizard docs to get started).

src/test/java/marquez/MarquezAppIntegrationTest.java
wslulciuc: Since I was already updating `MarquezApp` to add the db migration step, I thought I might as well add some test coverage. But, right, not relevant.

PR #44, opened 2018-08-27
Add Dockerfile

wslulciuc:


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM openjdk:9-jdk-slim AS builder
+WORKDIR /usr/src/app
+COPY . /usr/src/app
+RUN ./gradlew clean build
+
+FROM openjdk:9-jre-slim
```
harels: base docker image seems to be written twice

wslulciuc: I'm leveraging [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build) to keep things compact and cache gradle specific scripts, libs, etc.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM base AS build
+WORKDIR /usr/src/app
+COPY . .
+RUN ./gradlew --no-daemon shadowJar
+
+FROM openjdk:9-jdk-slim
+WORKDIR /usr/src/app
+COPY --from=build /usr/src/app/build/libs/marquez-all.jar marquez.jar
+EXPOSE 5000
+ENTRYPOINT ["java", "-jar", "marquez.jar", "server", "$MARQUEZ_CONFG"]
```
ashulmanWeWork: nit: `${MARQUEZ_CONFIG}`, for stylistic consistency.

wslulciuc: Good catch! Also, still figuring out what the best approach would be to set `config.yml`.

PR #49, opened 2018-08-28
indicate GET on /owners produces JSON

### LINKED ISSUE #48: Fix HTTP 500 for GET on /owners

sshah-wework:
Fixes #48

### DIFF
--- a/src/main/java/marquez/resources/OwnerResource.java
+++ b/src/main/java/marquez/resources/OwnerResource.java
@@ -4,12 +4,12 @@ import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
 import com.codahale.metrics.annotation.Timed;
 import javax.ws.rs.Consumes;
-import javax.ws.rs.Produces;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
 import marquez.api.Owner;
 import marquez.db.dao.OwnerDAO;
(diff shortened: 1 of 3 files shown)


src/main/java/marquez/resources/OwnerResource.java
```diff
@@ -31,6 +32,7 @@ public Response create(final Owner owner) {
 
   @Path("{owner}")
   @GET
+  @Produces(APPLICATION_JSON)
```
wslulciuc: Do we have a test to verify the response?

wslulciuc: We don't actually. Mind writing a test for the Owner resource?

sshah-wework: Ah so I changed it so that we will. We don't want to make an Owner fetchable?

```
HTTP/1.1 200 OK
Connection: close
Date: Tue, 28 Aug 2018 20:45:24 GMT
Content-Type: application/json
Vary: Accept-Encoding
Content-Length: 20

{"name":"analytics"}

sshah-wework: just mocks -- adding one for the response

sshah-wework: Looks like verifying a GET endpoint produces JSON is done via two separate tests:

1) a GET endpoint returns an expected object
2) the object can be serialized to JSON

wslulciuc: Agreed. Based on our discussions, performing `GET` on owner is not an op we want to support. When we introduce `/admin/*` that may be an op we'd like to support.

Also, the `OwnerTest` was in the `api` pkg. Assumed it be on `resources`. Might want to move things around, which is being addressed in #50

PR #54, opened 2018-08-30
Update ID -> UUID in the DB and DAOs

ashulmanWeWork:


ashulmanWeWork:
We looked at coverage and decided to move ahead.

sshah-wework:
OK to merge this one since we paired -- so my approval was implicit :)

PR #63, opened 2018-09-04
Job guid hiding

ashulmanWeWork:


src/test/java/marquez/api/JobEqualityTest.java
```diff
@@ -0,0 +1,69 @@
...
+  private static final String OWNER_NAME = "myOwner";
+  private static final Timestamp NOMINAL_TIME = Timestamp.from(Instant.now());
+  private static final String CATEGORY = "myCategory";
+  private static final String DESCRIPTION = "the first job";
+
+  private static Job JOB;
+
+  @BeforeClass
+  public static void setupTestJob() {
+    JOB = new Job(JOB_UUID, JOB_NAME, OWNER_NAME, NOMINAL_TIME, CATEGORY, DESCRIPTION);
```
sshah-wework: I think you can move this out of `setupTestJob` and just initialize `JOB` to this? If you set `JOB` to be `final` then there's no chance of it getting modified.

ashulmanWeWork:
Good call. Updated.

sshah-wework:
Can you add a serialization test which shows that GUID is not being serialized?

sshah-wework:
Also seems like the changes are not confined to just GUID hiding. In larger PRs the purpose should be more focused 😎

src/test/java/marquez/api/JobSerializationTest.java
```diff
@@ -0,0 +1,36 @@
...
+  @Test
+  public void testGuidNotSerialized() throws Exception {
+    final String serializedOutput = MAPPER.writeValueAsString(TEST_JOB);
+    assertThat(serializedOutput).doesNotContain("guid");
+    assertThat(serializedOutput).doesNotContain(TEST_JOB_UUID.toString());
+  }
+
+  @Test
+  public void testGuidSet() {
+    assertThat(TEST_JOB.getGuid().equals(TEST_JOB_UUID));
```
sshah-wework: We should probably rename this test class to `JobTest`. The test above is not related to serialization, it's testing a getter on the `Job` Representation. I do think serialization tests are intended to live alongside other tests of the Representation class (per the Dropwizard testing page), so this test does belong here.

PR #78, opened 2018-09-21
Replace h2 with embedded postgres for testing

ashulmanWeWork:
* Adjust infrastructure to use a single embedded postgres instance instead of h2
* Remove unnecessary db migration invocation 
* Adjust the SQL format to use syntax not available in h2, both in migration scripts and inline queries
* Minor test adjustments, cleanup, and refactoring
* Add clarity to test infra naming --> DB is the database, APP is the application


ashulmanWeWork:
Looking forward to working on the follow-on patch, as per Willy's suggestion above.

PR #89, opened 2018-10-08
Version 3.15.0 of spotless

wslulciuc:


build.gradle
```diff
@@ -2,7 +2,7 @@ plugins {
   id 'java'
   id 'application'
   id 'com.github.johnrengelman.shadow' version '2.0.4'
-  id "com.diffplug.gradle.spotless" version '3.14.0'
+  id "com.diffplug.gradle.spotless" version '3.15.0'
```
sshah-wework: Seems super weird that the old version went away -- should it be in maven? wondering if we can ask gradle to retry pulling dependencies a couple of times before failing entirely

sshah-wework: So here's 3.14 in the gradle repo: https://plugins.gradle.org/plugin/com.diffplug.gradle.spotless/3.14.0

Technically we should not have had to do a version bump here.

sshah-wework: The [old version](https://plugins.gradle.org/plugin/com.diffplug.gradle.spotless/3.14.0) is still in gradle so this _should_ have succeeded.

sshah-wework: Opened up #90 to increase gradle's HTTP connection timeouts

wslulciuc: We don't really benefit from the spotless `3.15.0` release, closing this PR. Thanks!

sshah-wework: Gotcha, my fault, I thought this was a workaround to 3.14 failing to download.

wslulciuc: In a way, it was sort of a workaround and my attempt to secretly bump the version at the same time :)

PR #90, opened 2018-10-08
Increase gradle HTTP timeouts on build

sshah-wework:
Trying to reduce the number of spurious build failures when a dependency / plugin download times out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-      - run: ./gradlew clean build --stacktrace
+      - run: ./gradlew clean build --stacktrace -Dhttp.socketTimeout=20000 -Dhttp.connectionTimeout=20000
   test:
     <<: *defaults
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-      - run: ./gradlew clean build --stacktrace
+      - run: ./gradlew clean build --stacktrace -Dhttp.socketTimeout=20000 -Dhttp.connectionTimeout=20000
```
wslulciuc: You can specify `-D` command-line options in [`gradle.properties`](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_system_properties). I imagine we'll have other overrides for our builds in the future. Thoughts?

sshah-wework: Ah nice, yeah that's better

gradle.properties
```diff
@@ -0,0 +1,2 @@
+http.socketTimeout=20000
+http.connectionTimeout=20000
```
wslulciuc: I think the property names have been changed? see https://github.com/gradle/gradle/blob/master/subprojects/resources-http/src/main/java/org/gradle/internal/resource/transport/http/JavaSystemPropertiesHttpTimeoutSettings.java#L26

PR #93, opened 2018-10-10
Remove need to run create a test file for Marquez tests

ashulmanWeWork:


src/main/java/marquez/MarquezConfig.java
```diff
@@ -5,9 +5,9 @@
 import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.flyway.FlywayFactory;
 
-public final class MarquezConfig extends Configuration {
+public class MarquezConfig extends Configuration {
   @JsonProperty("database")
-  private final DataSourceFactory database = new DataSourceFactory();
+  protected DataSourceFactory database = new DataSourceFactory();
```
wslulciuc: We wouldn't want to extend the Marquez base configuration. That is, the scope of `database` field should remain private and on unmodifiable once set. Also, you're not actually extending any behavior here. Sure, this does help writing tests but we want to avoid coupling test fixtures this way.

PR #96, opened 2018-10-11
Use openjdk 8 as base image for ci

wslulciuc:
This PR updates the CI base image to `circleci/openjdk:8-jdk` (see #86)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/marquez
   docker:
-    - image: circleci/openjdk:9-jdk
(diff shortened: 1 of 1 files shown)


ashulmanWeWork:
Change looks good but what's the rationale for it?

wslulciuc:
@ashulmanWeWork : It's related to PR #86, but the tl;dr is: libs in jdk8 were removed in jdk9 that Jersey depends on. Here's also a blog about the issue: [Why wont Jersey work on JDK 9](https://www.jeffryhouser.com/index.cfm/2017/12/21/Why-wont-Jersey-work-on-JDK-9)).

PR #104, opened 2018-10-13
Add openapi.yml [skip ci]

### LINKED ISSUE #53: Define OpenAPI spec
Let's define an API specification using [OpenAPI v3.0](https://swagger.io/specification/), the most popular API specification tool around

wslulciuc:
This PR defines the open API spec for Marquez (closes #53).

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -118,8 +118,8 @@ paths:
                 - outputDatasetUrns
                 - location
               example:
-                inputDatasetsUrns: ["urn:wework:room_bookings", "urn:meetup:scheduled_events"]
-                outputDatasetsUrns: ["urn:wework:room_bookings_7_days"]
+                inputDatasetsUrns: ["urn:wework:wedata.room_bookings", "urn:meetup:events.scheduled_events"]
+                outputDatasetsUrns: ["urn:wework:wedata.room_bookings_7_days"]
                 location: https://github.com/wework/jobs/commit/124f6089ad4c5fcbb1d7b33cbb5d3a9521c5d32c
                 description: Determine weekly room booking occupancy patterns.
       responses:
@@ -377,8 +377,8 @@ components:
         - location
         - description
       example:
-        inputDatasetsUrns: ["urn:wework:room_bookings", "urn:meetup:scheduled_events"]
-        outputDatasetsUrns: ["urn:wework:room_bookings_7_days"]
+        inputDatasetsUrns: ["urn:wework:wedata.room_bookings", "urn:meetup:events.scheduled_events"]
+        outputDatasetsUrns: ["urn:wework:wedata.room_bookings_7_days"]
         location: https://github.com/wework/jobs/commit/124f6089ad4c5fcbb1d7b33cbb5d3a9521c5d32c
         description: Determine weekly room booking occupancy patterns.
     jobs:
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -48,6 +48,18 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/namespace'
+    get:
+      summary: Retrieve a namespace
+      description: Returns a namespaces.
+      tags:
+        - Namespaces
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/namespace'
   /namespaces:
     get:
       summary: List all namespaces
@@ -336,8 +348,8 @@ components:
         - location
         - description
       example:
-        inputDatasetsUrns: ["urn:db:wework:wedata.room_bookings", "urn:db:meetup:meetup.scheduled_events"]
-        outputDatasetsUrns: ["urn:db:wework:wedata.room_bookings_7_days"]
+        inputDatasetsUrns: ["urn:wework:room_bookings", "urn:meetup:scheduled_events"]
+        outputDatasetsUrns: ["urn:wework:room_bookings_7_days"]
         location: https://github.com/wework/jobs/commit/124f6089ad4c5fcbb1d7b33cbb5d3a9521c5d32c
         description: Determine weekly room booking occupancy patterns.
     jobs:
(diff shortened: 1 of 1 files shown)


omervk:
Heya! I looked through the spec and have a few notes:

1. Awesome work, thank you :)
2. We've talked about jobs and datasets being globally unique by name. Has that changed?
3. How will job versions be handled? I only see one reference to them in the _List all job versions_ endpoint.
4. We'll need to name the connection between an dataset and a job when the dataset is the job's input (Job X uses Dataset A as an input names _ds_).
5. I'm not sure what nominal start/end times are.
6. Do we need to provide a job id when we have a run id? I'm referring to all of the endpoints for an existing run id.

Sorry if I'm reviewing something halfway through that wasn't meant to be reviewed.

sshah-wework:
Hi Omer, thanks for the questions. This is still in progress (not merged into master yet), but here are some early answers:

> We've talked about jobs and datasets being globally unique by name. Has that changed?
They will be unique to the namespace, but not globally.

> How will job versions be handled? I only see one reference to them in the List all job versions endpoint.
Can you clarify "handled"? Not much has changed about the Job Version concept. It is implicitly bumped when certain properties change (like the the source/artifact path) 

> We'll need to name the connection between an dataset and a job when the dataset is the job's input (Job X uses Dataset A as an input names ds).
If I understand correctly, "ds" is the name that Job X refers to? Is it possible for this to be stored outside of Marquez, like in a configuration file?

> I'm not sure what nominal start/end times are.
This is the start/end date partitions of data which is being processed. For example, if a job is processing only January 1st, 2018 data from a dataset, the start time would be "2018-01-01 00:00:00" and the end time would be "2018-01-01 23:59:59". These will remain optional fields at least until Q1.

> Do we need to provide a job id when we have a run id? I'm referring to all of the endpoints for an existing run id.
Which endpoint is this referring to?

> I'm not sure what namespaces are :)
Each team can have their own namespace where all of their jobs / datasets will be associated. Uniqueness of job / dataset names will be enforced at the namespace level.

HTH,
Shawn

omervk:
Thanks for the answers! I'll follow up on some:

> > How will job versions be handled? I only see one reference to them in the List all job versions endpoint.
> 
> Can you clarify "handled"? Not much has changed about the Job Version concept. It is implicitly bumped when certain properties change (like the the source/artifact path)

Which change in properties will trigger an update to the version? How will the new version be named?
Alternatively, users could share explicit job versions which would carry semantic meaning (JAR versions, Git SHAs, etc.)

> > We'll need to name the connection between an dataset and a job when the dataset is the job's input (Job X uses Dataset A as an input names ds).
> 
> If I understand correctly, "ds" is the name that Job X refers to? Is it possible for this to be stored outside of Marquez, like in a configuration file?

Yes. The reason to this is a job wanting the same dataset twice, only differently pre-filtered. For instance, the delta since the last run and the entirety of the dataset.
This could be specified using a configuration file, but would be better served in Marquez, where you could track what went into the job's run.

> > I'm not sure what nominal start/end times are.
> > This is the start/end date partitions of data which is being processed. For example, if a job is processing only January 1st, 2018 data from a dataset, the start time would be "2018-01-01 00:00:00" and the end time would be "2018-01-01 23:59:59". These will remain optional fields at least until Q1.

Got it. Are those required? If so, what happens if partitioning is done by some other field?

> > Do we need to provide a job id when we have a run id? I'm referring to all of the endpoints for an existing run id.
> 
> Which endpoint is this referring to?

Any endpoint with a run id in its path, I guess. They all are built using the job's id as well. I would assume you could infer the job id from the run id. This would mean, for instance, that the reporting process wouldn't need to carry the job's id around as well, just for the sake of building up the urls to report status to.

> > I'm not sure what namespaces are :)
> 
> Each team can have their own namespace where all of their jobs / datasets will be associated. Uniqueness of job / dataset names will be enforced at the namespace level.

That's a very useful concept :)

sshah-wework:
> > > How will job versions be handled? I only see one reference to them in the List all job versions endpoint.
> > 
> > Can you clarify "handled"? Not much has changed about the Job Version concept. It is implicitly bumped when certain properties change (like the the source/artifact path)
> 
> Which change in properties will trigger an update to the version? How will the new version be named?
> Alternatively, users could share explicit job versions which would carry semantic meaning (JAR versions, Git SHAs, etc.)

As mentioned, we're thinking of the URL to the source/artifact path as the key indicator of a version change. We will see how well this works out. Versions do not have different names from the job itself, only version IDs which are randomly generated and not meaningful outside of Marquez. JAR Versions or Git SHAs should be reflected in the source/artifact URLs provided.

> > > We'll need to name the connection between an dataset and a job when the dataset is the job's input (Job X uses Dataset A as an input names ds).
> > 
> > If I understand correctly, "ds" is the name that Job X refers to? Is it possible for this to be stored outside of Marquez, like in a configuration file?
> 
> Yes. The reason to this is a job wanting the same dataset twice, only differently pre-filtered. For instance, the delta since the last run and the entirety of the dataset.
> This could be specified using a configuration file, but would be better served in Marquez, where you could track what went into the job's run.

This is not metadata which Marquez needs to know at the moment. I am not sure I understand how a job can take the same dataset _twice_ as an input during a single execution. Whether it is processing the entire dataset, or a delta, it is treated the same in Marquez. The only potential difference would be the nominal times (mentioned below).

> > > I'm not sure what nominal start/end times are.
> > > This is the start/end date partitions of data which is being processed. For example, if a job is processing only January 1st, 2018 data from a dataset, the start time would be "2018-01-01 00:00:00" and the end time would be "2018-01-01 23:59:59". These will remain optional fields at least until Q1.
> 
> Got it. Are those required? If so, what happens if partitioning is done by some other field?

Optional. Partitioning by a non-date field should be rare (this only makes sense for _very_ unusual or static datasets). In those cases, it is effectively re-processing the entire potential date range. Let's discuss a concrete example of this offline.

> > > Do we need to provide a job id when we have a run id? I'm referring to all of the endpoints for an existing run id.
> > 
> > Which endpoint is this referring to?
> 
> Any endpoint with a run id in its path, I guess. They all are built using the job's id as well. I would assume you could infer the job id from the run id. This would mean, for instance, that the reporting process wouldn't need to carry the job's id around as well, just for the sake of building up the urls to report status to.

The reporting process has either: 1) been passed the Run ID, or 2) created the Run ID. In 1), job name could be passed as well, and in 2) the job name is already available.

> > > I'm not sure what namespaces are :)
> > 
> > Each team can have their own namespace where all of their jobs / datasets will be associated. Uniqueness of job / dataset names will be enforced at the namespace level.
> 
> That's a very useful concept :)

😺

sshah-wework:
We should move this conversation into a different medium, as it is becoming harder for me to follow the nested comments. Thanks for the responses, will ping you offline. We'll be sure to circle back with the decisions made in a GH issue afterward.

docs/openapi.yml
```diff
@@ -0,0 +1,547 @@
...
+        - Namespaces
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/namespaces'
+ 
+  /namespaces/{namespace}/jobs/{job}:
```
ashulmanWeWork: Add-on work for me to add the endpoint for getting info about a specific namespace based on its name.

PR #116, opened 2018-10-25
Add basic namespace resources and data object

ashulmanWeWork:


src/main/java/marquez/api/Namespace.java
```diff
@@ -0,0 +1,75 @@
+package marquez.api;
```
sshah-wework: I think we are moving non-request/response entities to `core/models`?

src/main/java/marquez/resources/NamespaceResource.java
```diff
@@ -0,0 +1,82 @@
+package marquez.resources;
```
sshah-wework: You can bring up the test coverage with tests for this resource.

src/test/java/marquez/api/JobRunBaseTest.java
```diff
@@ -40,7 +40,7 @@ public static void setUp() {
             handle -> {
               handle.execute(
                   format(
-                      "insert into jobs (guid, name, category, description) values "
+                      "insert into jobs (guid, name, category, DESCRIPTION) values "
```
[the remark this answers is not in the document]
ashulmanWeWork: good catch, yep!

src/main/java/marquez/api/Namespace.java
ashulmanWeWork: I think so, but I'd rather make that refactoring change when we move all the entities over (e.g. Owner, JobRun, etc.).

src/test/java/marquez/api/NamespaceIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+        APP.client()
+            .target(URI.create("http://localhost:" + APP.getLocalPort()))
+            .path("/namespace/" + NAMESPACE_NAME)
+            .request(MediaType.APPLICATION_JSON)
+            .put(createNamespaceRequestEntity);
+    assertEquals(Response.Status.NOT_FOUND.getStatusCode(), res.getStatus());
+  }
+
+  @Test
+  public void testValidNamespace() throws NamespaceException {
```
sshah-wework: this should be in a `NamespaceResource` test suite

src/test/java/marquez/api/NamespaceIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+        APP.client()
+            .target(URI.create("http://localhost:" + APP.getLocalPort()))
+            .path("/namespaces/" + "nosuchnamespace")
+            .request(MediaType.APPLICATION_JSON)
+            .get();
+    assertEquals(Response.Status.NOT_FOUND.getStatusCode(), res.getStatus());
+  }
+
+  @Test
+  public void testGetNamespaceRequestErrorHandling() throws Exception {
```
sshah-wework: `NamespaceResource` test

src/test/java/marquez/api/NamespaceIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+    assertEquals(Response.Status.OK.getStatusCode(), res.getStatus());
+
+    GetAllNamespacesResponse responseBody = res.readEntity(GetAllNamespacesResponse.class);
+    List<Namespace> returnedList = responseBody.getNamespaces();
+
+    assertThat(returnedList.isEmpty());
+  }
+
+  @Test
+  public void testGetAllNamespaceWithSingleResultSet() throws NamespaceException {
```
sshah-wework: `NamespaceResource` test

src/test/java/marquez/api/NamespaceIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+    Response res = namespaceResource.getAllNamespaces();
+
+    GetAllNamespacesResponse responseBody = (GetAllNamespacesResponse) res.getEntity();
+    List<Namespace> returnedList = responseBody.getNamespaces();
+
+    assertThat(returnedList).contains(TEST_NAMESPACE);
+  }
+
+  @Test
+  public void testGetAllNamespaceWithMultipleResultSet() throws NamespaceException {
```
sshah-wework: `NamespaceResource` test

src/test/java/marquez/api/NamespaceIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+    Response res = namespaceResource.getAllNamespaces();
+
+    GetAllNamespacesResponse responseBody = (GetAllNamespacesResponse) res.getEntity();
+    List<Namespace> returnedList = responseBody.getNamespaces();
+
+    assertThat(returnedList).containsExactly(TEST_NAMESPACE, secondNamespace);
+  }
+
+  @Test
+  public void testGetAllNamespacesErrorHandling() throws NamespaceException {
```
sshah-wework: `NamespaceResource` test

src/test/java/marquez/api/NamespaceIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+    NamespaceService namespaceService = Mockito.mock(NamespaceService.class);
+    NamespaceResource namespaceResource = new NamespaceResource(namespaceService);
+    when(namespaceService.getAllNamespaces()).thenThrow(NamespaceException.class);
+
+    Response res = namespaceResource.getAllNamespaces();
+    assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), res.getStatus());
+  }
+
+  @Test
+  public void testGetAllNamespacesWithoutTrailingSlash() {
```
sshah-wework: is this test needed? we won't ever change how the trailing slash is treated, it's really just testing how dropwizard works

src/main/java/marquez/resources/NamespaceResource.java
```diff
@@ -0,0 +1,108 @@
...
+                    returnedNamespace.getOwnerName(),
+                    returnedNamespace.getDescription()))
+            .type(APPLICATION_JSON)
+            .build();
+      } else {
+        return Response.status(Response.Status.NOT_FOUND).type(APPLICATION_JSON).build();
+      }
+    } catch (Exception e) {
+      return Response.status(Response.Status.BAD_REQUEST)
+          .entity(new ErrorResponse("Not able to create namespace: " + e.getLocalizedMessage()))
```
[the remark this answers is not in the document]
ashulmanWeWork: Agree that the stacktrace might be too much, but wouldn't it be helpful to have more detail as opposed to a generic message?

src/main/java/marquez/resources/NamespaceResource.java
```diff
@@ -0,0 +1,108 @@
...
+    this.namespaceService = namespaceService;
+  }
+
+  @PUT
+  @Consumes(APPLICATION_JSON)
+  @Timed
+  @Path("/{namespace}")
+  public Response create(
+      @PathParam("namespace") String namespace, @Valid CreateNamespaceRequest request) {
+    LOG.info("Trying to create about namespace " + namespace);
```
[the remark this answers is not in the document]
ashulmanWeWork: Removed these

src/main/java/marquez/resources/NamespaceResource.java
```diff
@@ -0,0 +1,108 @@
...
+    try {
+      Namespace n =
+          namespaceService.create(namespace, request.getOwner(), request.getDescription());
+      return Response.status(Response.Status.OK)
+          .entity(
+              new CreateNamespaceResponse(
+                  n.getOwnerName(), request.getDescription(), n.getCreatedAt()))
```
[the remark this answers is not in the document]
ashulmanWeWork: Noted - will keep that in mind.

src/test/java/marquez/api/NamespaceIntegrationTest.java
ashulmanWeWork: Good point. Will remove.

src/main/java/marquez/resources/NamespaceResource.java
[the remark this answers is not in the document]
sshah-wework: We can and should log the exception stacktrace. But for the caller, this is way too much info. Something unexpected went wrong on our side, there's nothing they should have to know / do on their side.

ashulmanWeWork:
I noticed that the response from the getAllNamespaces was a list, with each item being the toString representation of a Namespace. Not only is that not elegant and not particularly usable, it leaks GUID info that we wish to keep internal unless absolutely necessary.

In my latest patch, I added a mapper from a Namespace to a GetNamespaceResponse. I also adjusted my tests. This required also adding an equality operator to GetNamespaceResponse.

sshah-wework:
happy to review again whenever the changes are in -- thanks for the quick turn around! This is coming along nicely 🎉

ashulmanWeWork:
@sshah-wework, I updated the logging on errors. Should be ready for review :).

src/main/java/marquez/api/GetAllNamespacesResponse.java
```diff
@@ -0,0 +1,25 @@
+package marquez.api;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.ArrayList;
+import java.util.List;
+import marquez.core.models.Namespace;
+
+public class GetAllNamespacesResponse {
+
+  List<GetNamespaceResponse> namespaces = new ArrayList<>();
```
wslulciuc: Wouldn't `GetAllNamespacesResponse` contain a list of namespaces, not a list of `GetNamespaceResponse`. You can view `XResponse` or `XRequest` as strictly the HTTP body.

src/main/java/marquez/api/GetAllNamespacesResponse.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public GetAllNamespacesResponse(@JsonProperty("namespaces") List<Namespace> namespaceList) {
+    for (Namespace n : namespaceList) {
+      // TODO: Adjust this once we have a mapper in place
```
wslulciuc: You wouldn't need a mapper and can avoid transformations by:
```java
this.namespaces = namespaces
```

src/main/java/marquez/api/GetAllNamespacesResponse.java
```diff
@@ -0,0 +1,25 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.ArrayList;
+import java.util.List;
+import marquez.core.models.Namespace;
+
+public class GetAllNamespacesResponse {
+
+  List<GetNamespaceResponse> namespaces = new ArrayList<>();
+
+  public GetAllNamespacesResponse(@JsonProperty("namespaces") List<Namespace> namespaceList) {
```
wslulciuc: Very minor, but I think `ListNamespacesResponse` instead of `GetAllNamespacesResponse` translate more to the action of the caller.

src/main/java/marquez/api/GetAllNamespacesResponse.java
wslulciuc: Oh, I see why you used `GetNamespaceResponse`, there are fields we don't want to return. We still wouldn't use `GetNamespaceResponse` given it's a bit awkward, wondering if the naming convention `XResponse` / `XRequest` adds much benefits.

src/main/java/marquez/api/GetNamespaceResponse.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class GetNamespaceResponse {
+
+  final String name;
+  final String ownerName;
+  final String description;
+
+  @JsonCreator
+  public GetNamespaceResponse(
+      final String name,
```
wslulciuc: This works without `@JsonProperty("name")`?

src/test/java/marquez/NamespaceBaseTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.sql.Timestamp;
+import java.util.UUID;
+import marquez.api.CreateNamespaceRequest;
+import marquez.core.models.Namespace;
+import marquez.dao.fixtures.AppWithPostgresRule;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+
+public class NamespaceBaseTest {
```
wslulciuc: Seems like just boilerplate stuff you're abstracting away? Unless it's commonly shared across tests, I'd just have this defined in `NamespaceIntegrationTest`.

wslulciuc: So there is some reuse in `NamespaceResourceTest`, but that also means `AppWithPostgresRule` is defined without ever being used, which does add to the overall run of the tests for `NamespaceResource` given we are setting up postgres.

src/main/java/marquez/api/GetAllNamespacesResponse.java
ashulmanWeWork: They're different types right? Are you suggesting a more elegant way transform from one to another, or are you suggesting changing the input type?

src/main/java/marquez/api/GetAllNamespacesResponse.java
ashulmanWeWork: Sounds good. Less awkward-sounding this way.

src/main/java/marquez/api/GetNamespaceResponse.java
ashulmanWeWork: Good point. I'll add it.

src/test/java/marquez/NamespaceBaseTest.java
ashulmanWeWork: Good catch. Moved the app rule down to the class that needs it. Most everything else is used in both.

src/main/java/marquez/api/GetAllNamespacesResponse.java
ashulmanWeWork: Ok so removing the HTTP verb from the name? That might work in the case that the response for a PUT and a GET would be the same. However, looking at the api spec, that's not true in some cases (e.g. jobRun PUT vs. jobRun GET).

ashulmanWeWork: I'm totally on-board that the naming/semantics are a little funky. Open to suggestions.

src/main/java/marquez/api/GetNamespaceResponse.java
ashulmanWeWork: We'll get more clarity on this once we have integration tests that work with data.

src/main/java/marquez/api/GetAllNamespacesResponse.java
ashulmanWeWork: Fixed in latest version.

src/main/java/marquez/resources/NamespaceResource.java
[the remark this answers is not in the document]
ashulmanWeWork: Resolved.

PR #132, opened 2018-11-16
First pass at error handling code

ashulmanWeWork:


src/main/java/marquez/core/exceptions/DaoException.java
```diff
@@ -0,0 +1,3 @@
+package marquez.core.exceptions;
+
+public class DaoException extends Exception {}
```
wslulciuc: Is this exception being used?

src/main/java/marquez/core/exceptions/UnexpectedErrorException.java
```diff
@@ -0,0 +1,3 @@
+package marquez.core.exceptions;
+
+public class UnexpectedErrorException extends Throwable {}
```
wslulciuc: Is `UnexpectedErrorException` viewed more as a placeholder exception until more of the service layer is defined?

src/main/java/marquez/core/exceptions/DaoException.java
ashulmanWeWork: Not yet, but I put it in to represent errors that occur at the DAO layer. We can also remove it for this review if we think that's better.

src/main/java/marquez/core/exceptions/UnexpectedErrorException.java
ashulmanWeWork: I think this is the approach we agreed upon.

src/main/java/marquez/core/exceptions/DaoException.java
wslulciuc: Yeah, I would remove `DaoException` as it wouldn't represent the exceptions that might/would be thrown. We'd mainly be handling `SqlException` and therefore wouldn't have to define domain exceptions.

src/test/java/marquez/api/DatasetSerializationTest.java
```diff
@@ -11,8 +11,10 @@
 import java.util.TimeZone;
 import java.util.UUID;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
+@Ignore("Need to figure out why this started to fail")
```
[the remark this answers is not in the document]
wslulciuc: Sounds good. Mind opening an issue and adding a comment that references the issue #?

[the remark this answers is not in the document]
ashulmanWeWork: Sure!
https://github.com/MarquezProject/marquez/issues/140

ashulmanWeWork:
There are test failures unrelated to the patch that this build marks as ignored. More information in this issue: https://github.com/MarquezProject/marquez/issues/140

src/main/java/marquez/core/exceptions/UnexpectedErrorException.java
sshah-wework: @wslulciuc we will have the stacktrace to pin point where the error was in the stack. `UnexpectedErrorException` can be a really basic exception that all services will throw (in the absence of more specific exceptions) for now so that Aleks can move forward.

sshah-wework: Also, can we rename this to `UnexpectedException`? `UnexpectedErrorException` is a little confusing since it errors and exceptions are not exactly the same thing.

src/main/java/marquez/core/mappers/UnexpectedErrorExceptionMapper.java
```diff
@@ -0,0 +1,21 @@
...
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+import marquez.api.entities.ErrorResponse;
+import marquez.core.exceptions.UnexpectedErrorException;
+
+@Provider
+public class UnexpectedErrorExceptionMapper implements ExceptionMapper<UnexpectedErrorException> {
```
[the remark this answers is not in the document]
sshah-wework: In the future, other service exceptions will potentially map to the resource layer's `ServerError`. The nice thing there is that the only change will be to ensure that the new exception gets converted to a `ServerError`. No change to the `ExceptionMapper` will be needed. The current pattern requires every service exception to have it's own `ExceptionMapper` defined, which is tighter coupling than is needed.

src/test/java/marquez/api/DatasetSerializationTest.java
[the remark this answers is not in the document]
sshah-wework: In the future, we should make sure all tests are passing locally before opening the PR and not rely on CircleCI as the only gate for test errors. If the tests do not reliably pass locally, it will cause issues for other developers

src/main/java/marquez/core/mappers/UnexpectedErrorExceptionMapper.java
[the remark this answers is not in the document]
wslulciuc: +1

src/test/java/marquez/api/DatasetSerializationTest.java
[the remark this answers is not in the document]
ashulmanWeWork: I think we're already following that protocol. Tests marked ignore will not run either locally or remotely.

I do like that we're handling flaky/failing tests as we see them and not punting them for later. I think finding the root cause of this could be interesting, especially since it's passing in CircleCI and not locally.

PR #136, opened 2018-11-19
Always check db host:port on container startup

wslulciuc:


docker/entrypoint.sh
```diff
@@ -4,6 +4,6 @@
 
 set -eu
 
-if [ -f ./wait-for-it.sh ]; then ./wait-for-it.sh "${POSTGRES_HOST}:5432"; fi
+./wait-for-it.sh "${POSTGRES_HOST}:5432" -t 300 -s -- echo "postgres is up!"
```
sshah-wework: can we move the timeout (300) and postgres port (5432) into the docker-compose.yml file too?

wslulciuc: Great idea. Sure can!

PR #154, opened 2018-11-23
Implement job service

### LINKED ISSUE #165: Update jobs table to remove 'category' field
The current ER diagram plan has removed 'category' as a field on jobs. It should be removed and the code/tests should reflect that.

sshah-wework:
Still needs work to merge with issue-100 branch, some cleanup needed
Closes #165


src/main/java/marquez/dao/JobDAO.java
```diff
@@ -0,0 +1,55 @@
...
+            h.createUpdate(
+                    "INSERT INTO jobs (guid, name, current_owner_name, namespace_guid)"
+                        + " VALUES (:guid, :name, :current_owner_name, :namespace_guid)")
+                .bind("guid", job.getGuid())
+                .bind("name", job.getName())
+                .bind("current_owner_name", job.getOwnerName())
+                .bind("namespace_guid", job.getNamespaceGuid())
+                .execute();
+          });
+    } catch (Exception e) {
```
ashulmanWeWork: Is there a more specific exception we can catch here?

src/main/java/marquez/dao/JobRunDAO.java
```diff
@@ -0,0 +1,81 @@
...
+            h.createUpdate(
+                    "UPDATE job_runs SET current_state = :currentState, started_at = :startedAt, ended_at = :endedAt where guid = :guid")
+                .bindBean(jobRun)
+                .execute();
+            createJobRunStateDAO()
+                .insert(UUID.randomUUID(), jobRun.getGuid(), jobRun.getCurrentState());
+          });
+    } catch (Exception e) {
+      // TODO: Add better error handling
+      LOG.error(e.getMessage());
```
ashulmanWeWork: Is there a reason we're swallowing the exception here without re-throwing or otherwise handling it?

src/main/java/marquez/dao/JobRunDAO.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  String updateStateSQL =
```
ashulmanWeWork: nit: Please consider using StringBuilder

src/main/java/marquez/dao/JobRunDAO.java
sshah-wework: good catch --athough this PR was not marked ready for review

src/main/java/marquez/resources/NamespaceResource.java
```diff
@@ -0,0 +1,116 @@
...
+    this.namespaceService = namespaceService;
+  }
+
+  // TODO: Something like this is used in almost every resource so it can be factored out in
+  // the error-handling review
+  private static Response UNEXPECTED_ERROR_RESPONSE =
+      Response.status(Response.Status.INTERNAL_SERVER_ERROR)
+          .entity(new ErrorResponse("Unexpected error occurred"))
+          .type(APPLICATION_JSON)
+          .build();
```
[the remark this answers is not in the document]
sshah-wework: thank you. this PR is not yet tagged ready for review -- I'm finishing my cleanup before rebasing

sshah-wework:
Closes #165

src/main/java/marquez/resources/NamespaceResource.java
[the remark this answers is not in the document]
ashulmanWeWork: haha sorry - got a little ahead of myself :)

src/main/java/marquez/dao/JobRunDAO.java
sshah-wework: btw, just realized I didn't write this this particular DAO -- but I will be cleaning it up as part of this PR

sshah-wework: the queries have been modified and some of the older methods were removed

src/main/java/marquez/api/Job.java
```diff
@@ -12,20 +12,23 @@
...
   private final String category;
   private final String description;
+  private final String location;
 
   public Job(
       final UUID guid,
       final String name,
       final String ownerName,
       final Timestamp nominalTime,
       final String category,
```
ashulmanWeWork: Should we remove category?

src/main/java/marquez/core/exceptions/NamespaceException.java
```diff
@@ -0,0 +1,3 @@
+package marquez.core.exceptions;
+
+public class NamespaceException extends Exception {}
```
ashulmanWeWork: This Exception isn't used - should it be in the review?

src/main/java/marquez/core/models/JobRunState.java
```diff
@@ -0,0 +1,60 @@
...
+    this.jobRunGuid = jobRunGuid;
+    this.state = state;
+  }
+
+  public enum State {
+    NEW,
+    RUNNING,
+    COMPLETED,
+    FAILED,
+    ABORTED;
```
ashulmanWeWork: I think this exists in the other JobRunState.java file - maybe we should factor it out and reference instead of repeating it?

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,224 @@
...
+      if (runArgsJson == null || runArgsDAO.digestExists(runArgsDigest)) {
+        jobRunDAO.insert(jobRun);
+      } else {
+        jobRunDAO.insertJobRunAndArgs(jobRun, runArgs);
+      }
+      return jobRun;
+    } catch (UnableToExecuteStatementException | NoSuchAlgorithmException e) {
+      String err = "error creating job run";
+      log.error(err, e);
+      throw new UnexpectedException();
```
ashulmanWeWork: Can we factor this out since it's used in a few places?

src/test/java/marquez/api/JobRunIntegrationTest.java
```diff
@@ -11,14 +11,13 @@
...
 import javax.ws.rs.core.Response;
-import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class JobRunIntegrationTest extends JobRunBaseTest {
   private static Logger LOG = LoggerFactory.getLogger(JobRunIntegrationTest.class);
 
-  @Test
+  // @Test
```
ashulmanWeWork: Why are the tests being disabled?
Tactically if you want to disable a set of tests, please use `@Ignore("disable reason here")` at the class level. If you want to disable only some tests, you can also use `@Ignore` at the method level.

src/test/java/marquez/api/JobTest.java
```diff
@@ -33,42 +33,46 @@ public void testGuidSet() {
...
-    Job j2 = new Job(JOB_UUID, JOB_NAME, OWNER_NAME, NOMINAL_TIME, CATEGORY, DESCRIPTION);
+    Job j2 = new Job(JOB_UUID, JOB_NAME, OWNER_NAME, NOMINAL_TIME, CATEGORY, DESCRIPTION, null);
     assertEquals(JOB.hashCode(), j2.hashCode());
   }
 
   @Test
   public void testJobInequalityOnUUID() {
-    Job j2 = new Job(UUID.randomUUID(), JOB_NAME, OWNER_NAME, NOMINAL_TIME, CATEGORY, DESCRIPTION);
+    Job j2 =
+        new Job(UUID.randomUUID(), JOB_NAME, OWNER_NAME, NOMINAL_TIME, CATEGORY, DESCRIPTION, null);
```
ashulmanWeWork: Let's adjust the tests to remove references to category, since that's going to be removed.

src/test/java/marquez/core/models/ModelEqualsAndHashCode.java
```diff
@@ -0,0 +1,61 @@
+package marquez.core.models;
```
ashulmanWeWork: If it's a test class, please use the naming convention `XYZTest.java`.

src/test/java/marquez/dao/RunArgsDAOTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @ClassRule public static final AppWithPostgresRule APP = new AppWithPostgresRule();
+
+  final RunArgsDAO runArgsDAO = APP.onDemand(RunArgsDAO.class);
+  final String hexDigest = "07d4ee12aac795ec60a549dce809c8105c541f0c4f3e7715686953f1702940e0";
+  final String argsJson = "{'foo': 1}";
+  final RunArgs runArgs = new RunArgs(hexDigest, argsJson, null);
+
+  @Before
+  public void setUp() {}
```
ashulmanWeWork: Do we need this?

src/main/java/marquez/core/exceptions/NamespaceException.java
sshah-wework: I think this was introduced in one of your commits? 

```
$ find . -name NamespaceException.java | xargs git log
commit c91a6fcb5502804e5c84a954e16badced5131ad2
Author: ashulmanWeWork <41963174+ashulmanWeWork@users.noreply.github.com>
Date:   Wed Oct 31 15:23:26 2018 -0700

    Add basic namespace resources and data object (#116)
```

sshah-wework: I must have gotten picked up in the rebase

src/main/java/marquez/core/models/JobRunState.java
sshah-wework: I didn't refactor across core models and API models in this PR, as it already had quite a bit of scope. I just removed the dead code from this class. 👍 to refactoring further in a future PR

src/test/java/marquez/api/JobTest.java
sshah-wework: I didn't touch these API model tests too much, mostly tried to keep them as-is and work around them

src/main/java/marquez/core/services/JobService.java
sshah-wework: while the pattern is repeated, each one has a potentially different set of exceptions to catch and throw. at the moment they're the all the same, but they will vary independently. Refactoring it is not likely to cut down any code, only make it less idiomatic

src/test/java/marquez/api/JobRunIntegrationTest.java
sshah-wework: missed this, thanks

src/test/java/marquez/core/models/ModelEqualsAndHashCode.java
sshah-wework: great catch, thanks

src/test/java/marquez/dao/RunArgsDAOTest.java
sshah-wework: looks like not :)

src/main/java/marquez/core/services/JobService.java
ashulmanWeWork: Ok cool :)

src/main/java/marquez/core/exceptions/NamespaceException.java
sshah-wework: ah, this test is no longer valid, which is probably why I disabled. will mark ignored -- feel free to re-enable if you use these tests for the endpoints

src/test/java/marquez/api/JobRunIntegrationTest.java
sshah-wework: ah, this test is no longer valid, which is probably why I disabled. will mark ignored -- feel free to re-enable if you use these tests for the endpoints

src/test/java/marquez/dao/deprecated/JobDAOTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+
+public class JobDAOTest {
+
+  @ClassRule public static final AppWithPostgresRule APP = new AppWithPostgresRule();
+
+  @Before
+  public void setUp() {}
```
ashulmanWeWork: You can also remove this instance (and elsewhere).

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,224 @@
...
+    } catch (UnableToExecuteStatementException e) {
+      String err = "failed to get a job";
+      log.error(err, e);
+      throw new UnexpectedException();
+    }
+  }
+
+  public Job createJob(String namespace, Job job) throws UnexpectedException {
+    try {
+      Job existingJob = this.jobDAO.findByName(namespace, job.getName());
```
ashulmanWeWork: We should make sure there is testing to verify the behavior when there is no match (i.e. return null instead of throwing an exception).

sshah-wework: I'm not following -- it does not throw an exception if there's no match. It returns an optional with no value present.

In fact, there is such a test!
https://github.com/MarquezProject/marquez/pull/154/files#diff-c67c728ae0d22570bebb6e68e1266d7aR96

src/main/java/marquez/api/Job.java
sshah-wework: this is an API model, which I tried not to modify much. I think this was initially something I had to do to get a test to pass, but is no longer needed. It might make more sense to just undo my changes to the API models since it's causing some confusion.

src/main/java/marquez/core/services/JobService.java
ashulmanWeWork: Awesome! Reason I ask is b/c I've had the mappers throw an exception when there was no match, so I just want to make sure it's handled gracefully.

src/main/java/marquez/api/Job.java
sshah-wework: I reverted this unnecessary change to the Job API model, so it should now be in it's previous state.

sshah-wework: This should be good now -- there should not have been any API model changes in this PR since you'll be touching those for the resource layer.

src/test/java/marquez/api/JobTest.java
sshah-wework: Reverted these changes back to their original state (along with the API models themselves). Since you're working with the job resource layer, you'll be modifying these API models anyway, so don't want to collide.

src/main/java/marquez/core/models/JobRunState.java
sshah-wework: the JobRun API model probably only needs a string for the state (e.g. "NEW", "RUNNING"), not the enum anymore.

src/test/java/marquez/api/JobTest.java
ashulmanWeWork: Ok no worries - I can handle those.

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+
+  private static JobVersion createJobVersion(Job job) {
+    return new JobVersion(
+        UUID.randomUUID(),
+        job.getGuid(),
+        job.getLocation(),
+        JobService.computeVersion(job),
+        null,
```
sshah-wework: use lombok constructor, remove these nulls

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,224 @@
...
+        null,
+        null);
+  }
+
+  protected static UUID computeVersion(Job job) {
```
sshah-wework: add something in documentation about job versioning

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,224 @@
...
+      RunArgs runArgs = null;
+      if (null == jobDAO.findByName(namespaceName, jobName)) {
+        String err =
+            String.format(
+                "unable to find job <ns='%s', job name='%s'> to create job run",
+                namespaceName, jobName);
+        log.error(err);
+        throw new UnexpectedException();
+      }
+      Optional<JobVersion> latestJobVersion = getLatestVersionOfJob(namespaceName, jobName);
```
sshah-wework: look into `orThrow` to just raise exception automatically

src/main/java/marquez/dao/JobVersionDAO.java
```diff
@@ -1,22 +1,53 @@
...
   JobVersion findByVersion(@Bind("version") UUID version);
 
+  String findVersionsSQL =
+      "SELECT jv.* \n"
+          + "FROM job_versions jv\n"
+          + "INNER JOIN jobs j ON (j.guid = jv.job_guid AND j.name=:job_name)"
+          + "INNER JOIN namespaces n ON (n.guid = j.namespace_guid AND n.name=:namespace_name)\n"
+          + "ORDER BY created_at";
+
+  @SqlQuery(findVersionsSQL)
```
sshah-wework: do multi-line SQL in the `@SqlQuery`

src/main/java/marquez/dao/NamespaceDAO.java
```diff
@@ -0,0 +1,27 @@
...
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.customizer.BindBean;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RegisterRowMapper(NamespaceRow.class)
+public interface NamespaceDAO extends SqlObject {
```
sshah-wework: remove `SqlObject`

src/main/java/marquez/dao/NamespaceDAO.java
```diff
@@ -0,0 +1,27 @@
...
+  static final Logger LOG = LoggerFactory.getLogger(NamespaceDAO.class);
```
sshah-wework: use `@sl4fj` here

sshah-wework: remove logger

src/main/java/marquez/dao/NamespaceDAO.java
sshah-wework: all instances

src/main/resources/db/migration/V5__schema_updates_for_0-1-0.sql
```diff
@@ -65,9 +65,11 @@ ALTER TABLE job_runs ADD nominal_start_time TIMESTAMP;
...
 ALTER TABLE job_runs ADD FOREIGN KEY(job_run_args_guid) REFERENCES job_run_args(guid);
+ALTER TABLE job_runs ADD job_version_guid UUID NOT NULL;
+ALTER TABLE job_runs ADD FOREIGN KEY(job_version_guid) REFERENCES job_versions(guid);
 
 ALTER TABLE datasets ADD COLUMN urn VARCHAR(128) NOT NULL;
-ALTER TABLE datasets ADD namespace_guid UUID;
+ALTER TABLE datasets ADD namespace_guid UUID NOT NULL;
 ALTER TABLE datasets ADD FOREIGN KEY(namespace_guid) REFERENCES namespaces(guid);
-ALTER TABLE datasets ADD datasource_uuid UUID;
+ALTER TABLE datasets ADD datasource_uuid UUID NOT NULL;
```
sshah-wework: move these changes into the V6 migration file

src/test/java/marquez/api/JobRunDaoTest.java
```diff
@@ -28,28 +27,29 @@ public static void setUpRowMapper() {
                     JobRunState.State.fromInt(rs.getInt("state"))));
   }
 
-  @Test
+  // @Test
```
sshah-wework: mark this class with `@Ignore` instead of commenting

src/test/java/marquez/api/JobRunTest.java
```diff
@@ -23,7 +23,12 @@
 
   private static final JobRun JOB_RUN =
       new JobRun(
-          JOB_RUN_UUID, STARTED_AT_TIME, ENDED_AT_TIME, JOB_RUN_DEFINITION_UUID, CURRENT_STATE);
+          JOB_RUN_UUID,
+          STARTED_AT_TIME,
+          ENDED_AT_TIME,
+          JOB_RUN_DEFINITION_UUID,
+          CURRENT_STATE,
+          null);
```
sshah-wework: verify that last null was removed

build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
   compile 'org.jdbi:jdbi3-postgres:3.5.1'
   compile 'org.jdbi:jdbi3-sqlobject:3.5.1'
   compile 'org.postgresql:postgresql:42.2.5'
+  compile 'org.projectlombok:lombok:1.18.4'
```
wslulciuc: `lombok` is only a compile time dependency :) (see [gradle setup](https://projectlombok.org/setup/gradle))

```
compileOnly 'org.projectlombok:lombok:1.18.4'
```

src/main/java/marquez/api/JobRun.java
```diff
@@ -73,6 +76,10 @@ public Integer getCurrentState() {
     return currentState;
   }
 
+  public UUID getJobVersionGuid() {
+    return jobVersionGuid;
+  }
+
   @Override
   public boolean equals(final Object o) {
     if (this == o) return true;
```
wslulciuc: Did we want to clean up this class to us lombok? I think we can remove some of the logic around state transitions in a separate PR?

src/main/java/marquez/api/ListNamespacesResponse.java
```diff
@@ -8,13 +8,12 @@
...
   private final List<Namespace> namespaces;
 
-  // TODO: Constructor should accept marquez.api.models.Namespace instead
-  public ListNamespacesResponse(@JsonProperty("namespaces") List<Namespace> namespaces) {
-    this.namespaces = namespaces;
+  public ListNamespacesResponse(@JsonProperty("namespaces") List<Namespace> namespaceList) {
+    this.namespaces = namespaceList;
   }
 
   @JsonProperty("namespaces")
```
wslulciuc: Think this pulling changes from your branch that are out dated? we'll want to add back the comments + rename (see https://github.com/MarquezProject/marquez/blob/issue-100/src/main/java/marquez/api/ListNamespacesResponse.java)

sshah-wework: Good catch -- I remember picking this one in the rebase over the newer one. I'll fix this.

sshah-wework: Woops, yeah looks like I picked the wrong patch during the rebase, will fix.

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,224 @@
...
+import marquez.core.models.RunArgs;
+import marquez.dao.JobDAO;
+import marquez.dao.JobRunDAO;
+import marquez.dao.JobVersionDAO;
+import marquez.dao.RunArgsDAO;
+import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
+
+@Slf4j
+class JobService {
+  private JobDAO jobDAO;
```
wslulciuc: You can make these `final`

src/main/java/marquez/api/ListNamespacesResponse.java
sshah-wework: reverted this file to the latest from `issue-100`.

src/main/java/marquez/api/JobRun.java
sshah-wework: I left the API models the way they were since Aleks will be addressing in his resource layer PR

src/main/java/marquez/core/services/JobService.java
sshah-wework: good idea, done

build.gradle
sshah-wework: nice, done :)

src/main/java/marquez/api/JobRun.java
sshah-wework: I'll revert this change, probably shouldn't have touched this file

wslulciuc: 👍

src/main/java/marquez/api/ListNamespacesResponse.java
wslulciuc: 👍

src/test/java/marquez/api/JobRunTest.java
sshah-wework: removed.

src/test/java/marquez/api/JobRunDaoTest.java
sshah-wework: done.

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,222 @@
...
+      return jobRun;
+    } catch (UnableToExecuteStatementException | NoSuchAlgorithmException e) {
+      String err = "error creating job run";
+      log.error(err, e);
+      throw new UnexpectedException();
+    }
+  }
+
+  private static JobVersion createJobVersion(Job job) {
+    return new JobVersion(
```
wslulciuc: We may want to define `JobVersion.get(Job)` or similar for simplification.

src/main/java/marquez/dao/JobRunDAO.java
```diff
@@ -18,41 +21,40 @@
...
-                .bindBean(jobRun)
-                .execute();
-            createJobRunStateDAO()
-                .insert(UUID.randomUUID(), jobRun.getGuid(), jobRun.getCurrentState());
-          });
-    } catch (Exception e) {
-      // TODO: Add better error handling
-      LOG.error(e.getMessage());
-    }
+  @Transaction
```
wslulciuc: 💯

src/main/java/marquez/dao/RunArgsDAO.java
```diff
@@ -0,0 +1,21 @@
...
+import marquez.core.models.RunArgs;
+import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.customizer.BindBean;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+@RegisterRowMapper(RunArgsRow.class)
+public interface RunArgsDAO extends SqlObject {
```
wslulciuc: A reminder that we don't need to extend `SqlObject` if we don't need direct access to the handler.

src/main/java/marquez/core/services/JobService.java
sshah-wework: yeah a factory function could be nice in the future

src/main/java/marquez/core/services/JobService.java
```diff
@@ -0,0 +1,222 @@
...
+      log.error(err, e);
+      throw new UnexpectedException();
+    }
+  }
+
+  public JobRun createJobRun(
+      String namespaceName,
+      String jobName,
+      String runArgsJson,
+      Timestamp nominalStartTime,
```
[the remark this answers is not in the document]
sshah-wework: Ok I created #183 to do that change. Will fix in a separate PR and merge into `issue-100`

src/main/java/marquez/dao/NamespaceDAO.java
sshah-wework: done.

src/main/java/marquez/core/services/JobService.java
sshah-wework: #184

src/main/java/marquez/dao/RunArgsDAO.java
sshah-wework: Ah yeah, this was copy pasta. Looking at the JDBI doc, this is just an optional mixin, so 👍

sshah-wework: Looks like we used to need it when we were using `getHandle` to do things in transactions, but now that we're standardizing on the `@Transaction` annotation, we can remove this, which is good!

src/main/java/marquez/core/services/JobService.java
sshah-wework: Looked into `orElseThrow`, but there's an additional heap allocation of an `Optional<Job>` just to get the syntactic sugar. It does look nicer (`Optional.ofNullable(jobDAO.findByName(namespaceName, jobName)).orElseThrow(...)`), but I'm not sure if it's worthwhile unless the DAOs start returning `Optional`s. Maybe we can look into that for the future:

https://stackoverflow.com/questions/31395271/register-optional-for-jdbi

sshah-wework: #185

src/main/java/marquez/dao/JobVersionDAO.java
sshah-wework: this is done.

src/main/java/marquez/dao/JobRunDAO.java
sshah-wework: will probably use `+` for these SQL string concats for readability. but for other cases, agreed that `StringBuilder` can be a better option

src/test/java/marquez/core/models/ModelEqualsAndHashCode.java
sshah-wework: removed this file

sshah-wework:
Barely fails the test coverage threshold because of removed tests, but tests / build are passing, so will merge.

PR #155, opened 2018-11-24
Add optional method to mapper

wslulciuc:
This PR makes the `Mapper` abstract and adds `mapAsOptional()` that maps type `A` to `Optional<B>`.

```java
final Bar bar = new Bar();  // A
final FooMapper fooMapper = new FooMapper();
Optional<Foo> fooOpt = fooMapper.mapAsOptional(bar);  // A => Optional<B>
Optional<Foo> fooOpt = fooMapper.mapIfPresent(Optional.of(bar));  // Optional<A> => Optional<B>
Foo foo = fooMapper.map(bar);  // A => B
List<Foo> foo = fooMapper.map(List.of(bar));  // List(A) => List(B)
```

### DIFF
--- a/src/main/java/marquez/api/ListNamespacesResponse.java
+++ b/src/main/java/marquez/api/ListNamespacesResponse.java
@@ -22,7 +22,7 @@ public class ListNamespacesResponse {
     // TODO: simplify to just: return namespaces;
     return namespaces
         .stream()
-        .map(namespaceMapper::mapIfPresent)
+        .map(namespaceMapper::mapAsOptional)
         .filter(Optional::isPresent)
         .map(Optional::get)
         .collect(Collectors.toList());
--- a/src/main/java/marquez/api/ListNamespacesResponse.java
+++ b/src/main/java/marquez/api/ListNamespacesResponse.java
@@ -20,7 +20,7 @@ public class ListNamespacesResponse {
   public List<marquez.api.Namespace> getNamespaces() {
     return namespaces
         .stream()
-        .map(namespaceMapper::map)
+        .map(namespaceMapper::mapIfPresent)
         .filter(Optional::isPresent)
         .map(Optional::get)
         .collect(Collectors.toList());
--- a/src/main/java/marquez/api/ListNamespacesResponse.java
+++ b/src/main/java/marquez/api/ListNamespacesResponse.java
@@ -8,16 +8,18 @@ import marquez.core.mappers.CoreNamespaceToApiNamespaceMapper;
 import marquez.core.models.Namespace;
 
 public class ListNamespacesResponse {
-
+  // TODO: Remove use of CoreNamespaceToApiNamespaceMapper
   CoreNamespaceToApiNamespaceMapper namespaceMapper = new CoreNamespaceToApiNamespaceMapper();
   private final List<Namespace> namespaces;
 
-  public ListNamespacesResponse(@JsonProperty("namespaces") List<Namespace> namespaceList) {
-    this.namespaces = namespaceList;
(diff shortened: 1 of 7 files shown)


src/main/java/marquez/api/ListNamespacesResponse.java
```diff
@@ -20,7 +20,7 @@ public ListNamespacesResponse(@JsonProperty("namespaces") List<Namespace> namesp
   public List<marquez.api.Namespace> getNamespaces() {
     return namespaces
         .stream()
-        .map(namespaceMapper::map)
+        .map(namespaceMapper::mapIfPresent)
         .filter(Optional::isPresent)
```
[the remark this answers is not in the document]
wslulciuc: I only updated references to `NamespaceMapper.map()` and didn't look too much into its usage. But, yeah, this could be cleaned up. I added comments to address them later (see [abaf94e](https://github.com/MarquezProject/marquez/pull/155/commits/abaf94e3cddd9963e517465ef7ec69ce6dc4e83e)).

src/main/java/marquez/core/mappers/Mapper.java
```diff
@@ -1,7 +1,17 @@
...
-interface Mapper<A, B> {
-  Optional<B> map(A value);
+public abstract class Mapper<A, B> {
+  public Optional<B> mapAsOptional(A value) {
+    return Optional.ofNullable(map(value));
+  }
+
+  public Optional<B> mapIfPresent(Optional<A> value) {
+    return requireNonNull(value).map(this::map);
+  }
```
sshah-wework: Could these both just be called `map`  with overloading?

```
  public Optional<B> map(A value) {
...
   public Optional<B> map(Optional<A> value) {
```

wslulciuc: I tend to favor descriptive method names over overloading methods with the same set of parameter. But, it would simplify the mapper contract, so sure!

sshah-wework: OK, no reason to change it just because of my comment -- it's your call. Just wanted to make sure I understood the alternative.

src/main/java/marquez/core/mappers/GetNamespaceResponseMapper.java
```diff
@@ -1,21 +1,19 @@
...
-public class GetNamespaceResponseMapper implements Mapper<Namespace, GetNamespaceResponse> {
-
-  CoreNamespaceToApiNamespaceMapper namespaceMapper = new CoreNamespaceToApiNamespaceMapper();
-
-  @Override
-  public Optional<GetNamespaceResponse> map(Namespace namespace) {
-    if (namespace == null) {
-      return Optional.empty();
-    }
+// TODO: Move to marquez.api.mappers pgk
```
sshah-wework: are these `TODO`s still valid?

wslulciuc: Yeah, I was thinking to add all model mappers that are API specific to the `marquez.api.mappers` pkg and keep only core mappers in `marquez.core.mappers`. Thoughts?

src/main/java/marquez/core/mappers/Mapper.java
wslulciuc: Well, for sure `Mapper. mapIfPresent()` if I'm given the choice :)

src/main/java/marquez/core/mappers/GetNamespaceResponseMapper.java
ashulmanWeWork: I like that idea.

PR #167, opened 2018-11-28
Use wait-for-db.sh instead of wait-for-it.sh

wslulciuc:
This PR simplifies ensuring postgres is up and removes the need for [wait-for-it.sh](https://github.com/vishnubob/wait-for-it)


docker/build-and-push.sh
```diff
@@ -13,18 +15,17 @@ project_root=$(git rev-parse --show-toplevel)
 cd "${project_root}"
 
 # Version X.Y.Z of Marquez image to build
-version=$1
+version="$1"
```
ashulmanWeWork: "${1}"
Not a big deal here, but in other places will keep you out of trouble if you have more than 10 arguments :)

docker/wait-for-db.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#
+# Usage: $ wait-for-db.sh
```
ashulmanWeWork: The usage comment doesn't seem to include the arguments for host and port.

docker-compose.yml
```diff
@@ -10,10 +10,10 @@ services:
       - MARQUEZ_PORT=5000
       - MARQUEZ_CONFIG=/usr/src/app/config.yml
       - POSTGRES_HOST=db
+      - POSTGRES_PORT=5432
       - POSTGRES_DB=marquez
       - POSTGRES_USER=buendia
```
ashulmanWeWork: 👍

Issue #170, opened 2018-11-29
docker compose functionality is currently broken


arron-green:
attempt to run docker-compose and received the following error

```bash
marquez (master) $ docker-compose up app
Creating network "marquez_net" with the default driver
Pulling db (postgres:9.6)...
9.6: Pulling from library/postgres
a5a6f2f73cd8: Already exists
e50fbea8af5a: Pull complete
73b4855ad326: Pull complete
39616673f22b: Pull complete
94e1b79f69ee: Pull complete
c91e4af2ff8e: Pull complete
16ba314c612f: Pull complete
89011a174cee: Pull complete
19f0371d4631: Pull complete
e7e01858ad6f: Pull complete
afbc9a790799: Pull complete
9cc4f3abece2: Pull complete
1fdd6c84dc56: Pull complete
b6fa74bdb3e4: Pull complete
Creating marquez_db_1_7c04dd4f3958 ... done
Creating marquez_app_1_de8e10ae9d71 ... done
Attaching to marquez_app_1_5849606d14e2
app_1_5849606d14e2 | INFO Waiting for postgres to become available...
app_1_5849606d14e2 | INFO Waiting for postgres to become available...
app_1_5849606d14e2 | INFO Waiting for postgres to become available...
app_1_5849606d14e2 | INFO Waiting for postgres to become available...
app_1_5849606d14e2 | INFO Great news! Postgres is up.
app_1_5849606d14e2 | java.io.FileNotFoundException: /usr/src/app/config.yml (Is a directory)
app_1_5849606d14e2 |     at java.io.FileInputStream.open0(Native Method)
app_1_5849606d14e2 |     at java.io.FileInputStream.open(FileInputStream.java:195)
app_1_5849606d14e2 |     at java.io.FileInputStream.<init>(FileInputStream.java:138)
app_1_5849606d14e2 |     at io.dropwizard.configuration.FileConfigurationSourceProvider.open(FileConfigurationSourceProvider.java:21)
app_1_5849606d14e2 |     at io.dropwizard.configuration.SubstitutingSourceProvider.open(SubstitutingSourceProvider.java:37)
app_1_5849606d14e2 |     at io.dropwizard.configuration.BaseConfigurationFactory.build(BaseConfigurationFactory.java:80)
app_1_5849606d14e2 |     at io.dropwizard.cli.ConfiguredCommand.parseConfiguration(ConfiguredCommand.java:126)
[...]

PR #171, opened 2018-11-29
Copy example config on docker-compose up [skip ci]

wslulciuc:
This PR patches `docker-compose` to use `config.example.yml`, eliminating the need to copy it manually as a pre-configuration step #170


wslulciuc:
Great! Thanks again @arron-green for opening the issue!

PR #175, opened 2018-11-30
update runArgs to be string

### LINKED ISSUE #172: Change runArgs type from map to string in OpenAPI spec

sshah-wework:
The JSON encoded run arguments are simpler to handle as a string
than as a map. Handling a map (with unknown fields) would involve
repeated ser/deserialization and require reflection to preserve
types correctly.

Fixes #172

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -201,14 +201,10 @@ paths:
                   type: string
                   format: date-time
                 runArgs:
-                  description: The runtime arguments of the job.
-                  type: map
+                  description: The runtime arguments of the job. JSON encoded.
+                  type: string
               example:
-                runArgs:
-                  email: data@wework.com
-                  emailOnFailure: false
-                  emailOnRetry: true
-                  retries: 1
+                runArgs: "{\\\"email\\\": \\\"data@wework.com\\\", \\\"emailOnFailure\\\": false, \\\"emailOnRetry\\\": true, \\\"retries\\\": 1}"
       responses:
         '201':
           description: CREATED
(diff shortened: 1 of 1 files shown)


docs/openapi.yml
```diff
@@ -201,14 +201,10 @@ paths:
...
-                  type: map
+                  description: The runtime arguments of the job. JSON encoded.
+                  type: string
               example:
-                runArgs:
-                  email: data@wework.com
-                  emailOnFailure: false
-                  emailOnRetry: true
-                  retries: 1
+                runArgs: "{\\\"email\\\": \\\"data@wework.com\\\", \\\"emailOnFailure\\\": false, \\\"emailOnRetry\\\": true, \\\"retries\\\": 1}"
```
wslulciuc: Do we need the extra `\`? I think we might be ok with:
```json
"{\"email\": ...
```

docs/openapi.yml
```diff
@@ -513,11 +509,7 @@ components:
       example:
         runId: 870492da-ecfb-4be0-91b9-9a89ddd3db90
         runState: COMPLETED
-        runArgs:
-          email: data@wework.com
-          emailOnFailure: false
-          emailOnRetry: true
-          retries: 1
+        runArgs: "{\\\"email\\\": \\\"data@wework.com\\\", \\\"emailOnFailure\\\": false, \\\"emailOnRetry\\\": true, \\\"retries\\\": 1}"
```
ashulmanWeWork: I'm not a fan of escaped json - but for Q4 in the interest of expediency, I'm okay with it.

PR #192, opened 2018-12-13
Implement namespace service

sshah-wework:
Namespace Service implementation

### DIFF
--- a/src/main/java/marquez/core/services/NamespaceService.java
+++ b/src/main/java/marquez/core/services/NamespaceService.java
@@ -22,7 +22,7 @@ public class NamespaceService {
       Namespace newNamespace =
           new Namespace(
               UUID.randomUUID(),
-              namespace.getName().toLowerCase(),
+              namespace.getName(),
               namespace.getOwnerName(),
               namespace.getDescription());
       namespaceDAO.insert(newNamespace);
--- a/src/main/java/marquez/core/services/NamespaceService.java
+++ b/src/main/java/marquez/core/services/NamespaceService.java
@@ -34,9 +34,9 @@ public class NamespaceService {
     }
   }
 
-  public boolean exists(Namespace namespace) throws UnexpectedException {
+  public boolean exists(String namespaceName) throws UnexpectedException {
     try {
-      return namespaceDAO.exists(namespace.getName());
+      return namespaceDAO.exists(namespaceName.toLowerCase());
     } catch (UnableToExecuteStatementException e) {
       String err = "error checking namespace existence";
       log.error(err);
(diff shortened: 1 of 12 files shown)


src/main/java/marquez/core/services/NamespaceService.java
```diff
@@ -1,30 +1,66 @@
...
+  public NamespaceService(NamespaceDAO namespaceDAO) {
+    this.namespaceDAO = namespaceDAO;
+  }
+
+  public Namespace create(Namespace namespace) throws UnexpectedException {
+    try {
+      Namespace newNamespace =
+          new Namespace(
+              UUID.randomUUID(),
+              namespace.getName().toLowerCase(),
```
wslulciuc: I'd call `toLowerCase()` on the namespace name in the namespace constructor. That way we ensure consistent casing when instantiating a namespace.

src/test/java/marquez/api/NamespaceIntegrationTest.java
```diff
@@ -33,6 +33,26 @@ public void testCreateNamespace() {
...
+        .put(Entity.json(createNamespaceRequest));
+    Response res =
+        APP.client()
+            .target(URI.create("http://localhost:" + APP.getLocalPort()))
+            .path("/namespaces/" + NAMESPACE_NAME)
+            .request(MediaType.APPLICATION_JSON)
+            .put(Entity.json(createNamespaceRequest));
+    CreateNamespaceResponse responseBody = res.readEntity(CreateNamespaceResponse.class);
+    assertEquals(Response.Status.OK.getStatusCode(), res.getStatus());
+    assertThat(responseBody.getNamespace().getCreatedAt()).isAfter(START_TIME);
```
wslulciuc: We'll want to discuss which testing framework we'd like to use for consistency across tests. I think there's a lot we can do with the simplicity of `org.junit.Assert` :)

src/test/java/marquez/core/services/NamespaceServiceTest.java
```diff
@@ -0,0 +1,123 @@
...
+    List<Namespace> namespacesFound = namespaceService.listNamespaces();
+    assertEquals(0, namespacesFound.size());
+  }
+
+  @Test(expected = UnexpectedException.class)
+  public void testListNamespaces_findAllException() throws UnexpectedException {
+    doThrow(UnableToExecuteStatementException.class).when(namespaceDAO).findAll();
+    namespaceService.listNamespaces();
+  }
+}
```
wslulciuc: 💯

src/main/java/marquez/core/services/NamespaceService.java
wslulciuc: Sure! In the mapper also works.

sshah-wework: moved into a mapper

Issue #100, opened 2018-10-12
App Design / Structure


wslulciuc:
To date, we have encouraged quick iteration on features to allow for early API feedback. This ensured we understood how Marquez would collect metadata on running jobs as well as handle versioning of datasets (conversations that are still ongoing). Recently, our [data model](https://www.lucidchart.com/documents/view/f918ce01-9eb4-4900-b266-49935da271b8/0) has seen many additions (_namespaces!_). As a result, we need to address our current app structure. Mainly, this means decoupling the _DAO Layer_ from the _Resource Layer_ and introduce a _Service Layer_ to encapsulate all DAO interactions and allow each layer to evolve independently #66.

**Note:** App restructuring is required before opening up the project for contributions.
 
### App Design

The multi-layer app design will consist of: 

- **Resource Layer**
- **Service Layer**
- **DAO Layer**

See [Organizing Marquez Code](https://docs.google.com/document/d/17qrJNFx5YfFWiSNqd53lMed-YOfjTeZUhkis3rAGbN0/edit) doc for more details.

### App Structure

We'll also have the following project structure:

```
marquez/
├── api/
│   ├── exceptions
│   ├── mappers
│   ├── models
│   └── validation
├── core/
│   ├── exceptions
│   ├── mappers
│   ├── models
│   └── services
├── db
├── health
└── resources
```
* `api`:  Representations (i.e. request and response bodies).
** `api/exceptions`:  All registered custom [ExceptionMapper](https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/ext/ExceptionMapper.html).
* `api/models`:  API models *not* including req/res bodies.
* `api/mappers`: Defines model [Mappers](https://github.com/MarquezProject/marquez/blob/update-app-structure/src/main/java/marquez/core/mappers/Mapper.java).
* `api/validation`: Validation of request and response bodies.
* `core`: Domain implementation; objects not used in the API.
* `core/exceptions`: Defines all runtime executions
* `core/mappers`: Provides data transformations across layers  
* `core/models`: Objects not used in the API 
* `core/services`: Service classes encapsulating all DAO interactions / logic #118 
* `db`: Database access classes; defines all [RowMappers](http://jdbi.org/#_row_mappers)
* `health`: Runtime health checks
* `resources`: Resources exposed in our RESTful API

sshah-wework:
Nice 👍

PR #194, opened 2018-12-14
Add DatasetAPI

wslulciuc:
This PR adds the following #100:

- An endpoint to list datasets
- [`DatasetService`](https://github.com/MarquezProject/marquez/blob/add-dataset-api/src/main/java/marquez/service/DatasetService.java) **class**
- [`DatasetDao`](https://github.com/MarquezProject/marquez/blob/add-dataset-api/src/main/java/marquez/db/DatasetDao.java) **class**
- More lombok usage!
- A set of common types / models:
  - [`DatasetUrn`](https://github.com/MarquezProject/marquez/blob/add-dataset-api/src/main/java/marquez/common/models/DatasetUrn.java) **class**
  - [`Namespace`](https://github.com/MarquezProject/marquez/blob/add-dataset-api/src/main/java/marquez/common/models/Namespace.java) **class**
  - [`ConnectionUrl`](https://github.com/MarquezProject/marquez/blob/add-dataset-api/src/main/java/marquez/common/models/ConnectionUrl.java) **class**
  - and more!


src/main/java/marquez/common/models/Namespace.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.regex.Pattern;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+
+@EqualsAndHashCode
+@ToString
+public final class Namespace {
+  private static final Pattern NAMESPACE_PATTERN = Pattern.compile("^[a-zA-Z0-9_]{1,1024}$");
```
sshah-wework: same here re: the `1024`, eventually want to pull it out

src/main/java/marquez/db/DbTableInfoDao.java
```diff
@@ -0,0 +1,27 @@
...
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.customizer.BindBean;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+@RegisterRowMapper(DbTableInfoRowMapper.class)
+public interface DbTableInfoDao {
+  @SqlUpdate(
+      "INSERT INTO db_table_info (uuid, db, table) " + "VALUES (:uuid, :db.value, :schema.value)")
```
sshah-wework: supppperrrrr nit: this SQL fits on one line now, so need for the `+`

sshah-wework: also the `schema.value` is going into the `table` column?

src/main/java/marquez/db/DbTableVersionDao.java
```diff
@@ -0,0 +1,37 @@
...
+          + "VALUES (:uuid, :datasetUuid, :dbTableInfoUuid, :table.value)")
+  void insert(
+      @Bind("uuid") UUID uuid,
+      @Bind("datasetUuid") UUID datasetUuid,
+      @Bind("dbTableInfoUuid") UUID dbTableInfoUuid,
+      @BindBean("table") Table table);
+
+  @Transaction
+  default void insert(UUID datasetUuid, Db db, Schema schema, Table table) {
+    final UUID dbTableInfoUuid = UUID.randomUUID();
```
sshah-wework: eventually these UUIDs should be generated in the service layer instead of the DAO layer

src/main/java/marquez/db/DbTableInfoDao.java
sshah-wework: need to add `ON CONFLICT DO NOTHING`

sshah-wework: this table will also need a unique index on `(db, schema)`, or how do we avoid this table getting a new row every time we insert a new `db_table_version` even when the `db_table_info` does not change?

src/main/java/marquez/api/mappers/DatasetResponseMapper.java
```diff
@@ -0,0 +1,16 @@
...
+import marquez.common.Mapper;
+import marquez.service.models.Dataset;
+
+public final class DatasetResponseMapper implements Mapper<Dataset, DatasetResponse> {
+  @Override
+  public DatasetResponse map(@NonNull Dataset dataset) {
+    return new DatasetResponse(
+        dataset.getUrn().getValue(),
+        dataset.getCreatedAt().toString(),
+        dataset.getDescription().map(String::valueOf).orElse(null));
```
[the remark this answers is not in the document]
wslulciuc: We could. I was trying not to overthink the mappers implementation. A possible alternative would be:

```java
public final class DatasetResponseMapper implements Mapper<Dataset, Optional<DatasetResponse>> {
  ...
  public Optional<DatasetResponse> map(@NonNull Dataset dataset) { ... }
}
```

I guess the question becomes can the result of `Mapper.map()` ever result in a _null_? We also currently support `Mapper.map(List)` which currently returns `List<B>`. We'd need to update the return type accordingly to allow for `List<Optional<B>>` which is an anti-pattern (that I agree with!) (https://dzone.com/articles/optional-anti-patterns)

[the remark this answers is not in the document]
wslulciuc: We could filter out `Optional<B>` that aren't present and return `List<B>` on `Mapper.map(List)`. So I take back what I said :)

src/test/java/marquez/api/resources/DatasetResourceTest.java
```diff
@@ -0,0 +1,64 @@
...
+    final Response response = datasetResource.list(NAMESPACE.getValue(), LIMIT, OFFSET);
+    assertEquals(OK, response.getStatusInfo());
+
+    final DatasetsResponse datasetsResponse = (DatasetsResponse) response.getEntity();
+    final List<DatasetResponse> datasetsResponses = datasetsResponse.getDatasetResponses();
+    assertEquals(1, datasetsResponses.size());
+    assertEquals(URN.getValue(), datasetsResponses.get(0).getUrn());
+    assertEquals(CREATED_AT.toString(), datasetsResponses.get(0).getCreatedAt());
+
+    verify(mockDatasetService, times(1)).getAll(NAMESPACE, LIMIT, OFFSET);
```
sshah-wework: i don't think you need the `times(1)` if it's only getting called once

src/test/java/marquez/service/DatasetServiceTest.java
```diff
@@ -0,0 +1,3 @@
+package marquez.service;
+
+public class DatasetServiceTest {}
```
sshah-wework: 😢

src/main/java/marquez/api/resources/DatasetResource.java
```diff
@@ -0,0 +1,60 @@
...
+  @Path("/datasets")
+  @Produces(APPLICATION_JSON)
+  public Response list(
+      @PathParam("namespace") String namespace,
+      @QueryParam("limit") @DefaultValue("100") Integer limit,
+      @QueryParam("offset") @DefaultValue("0") Integer offset)
+      throws UnexpectedException, WebApplicationException {
+    if (!namespaceService.exists(namespace)) {
+      throw new WebApplicationException(
+          String.format("The namespace %s does not exist.", namespace), NOT_FOUND);
```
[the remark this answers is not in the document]
wslulciuc: /cc @ashulmanWeWork :)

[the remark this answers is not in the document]
sshah-wework: Cool, I like this! Perhaps we can combine w/ sub-resources rather than inheritance to handle the namespace existence check for all resources: http://docs.jboss.org/resteasy/docs/1.0.0.GA/userguide/html/JAX-RS_Resource_Locators_and_Sub_Resources.html

src/main/java/marquez/common/models/Namespace.java
wslulciuc: 👍

src/main/java/marquez/common/models/Urn.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.NonNull;
+import lombok.ToString;
+
+@EqualsAndHashCode
+@ToString
+public final class Urn {
+  private static final String URN_DELIM = ":";
+  private static final String URN_PREFIX = "urn";
+  private static final String URN_REGEX =
+      String.format("^%s(%s[a-zA-Z0-9.]{1,64}){2}$", URN_PREFIX, URN_DELIM);
```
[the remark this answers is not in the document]
wslulciuc: 👍

src/test/java/marquez/service/DatasetServiceTest.java
wslulciuc: // TODO: write all the tests :)

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - checkout
       - restore_cache:
           <<: *cache
-      - run: ./gradlew --no-daemon --stacktrace build -x test checkstyleTest
+      - run: ./gradlew --no-daemon --stacktrace build -x test
```
ashulmanWeWork: This should be part of a different PR, in case it causes any kind of unexpected behavior.

build.gradle
```diff
@@ -1,11 +1,12 @@
 plugins {
   id 'application'
+  id 'java'
   id 'checkstyle'
+  id 'jacoco'
+  id 'com.adarshr.test-logger' version '1.6.0'
   id 'com.diffplug.gradle.spotless' version '3.16.0'
   id 'com.github.jk1.dependency-license-report' version '1.2'
   id 'com.github.johnrengelman.shadow' version '4.0.3'
```
ashulmanWeWork: Why was this re-ordered?

src/main/java/marquez/api/mappers/DatasetResponseMapper.java
```diff
@@ -0,0 +1,27 @@
...
+        dataset.getUrn().getValue(),
+        dataset.getCreatedAt().toString(),
+        dataset.getDescription().map((desc) -> desc.getValue()).orElse(null));
+  }
+
+  public static List<DatasetResponse> map(@NonNull List<Dataset> datasets) {
+    return datasets.isEmpty()
+        ? Collections.emptyList()
+        : Collections.unmodifiableList(
+            datasets.stream().map((dataset) -> map(dataset)).collect(toList()));
```
ashulmanWeWork: does `dataset` need to be in parenthesis?

src/main/java/marquez/api/models/DatasetResponse.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+@RequiredArgsConstructor
+@EqualsAndHashCode
+@ToString
+public class DatasetResponse {
```
ashulmanWeWork: I see that we are able to get away with not having a `NoArgsConstructor`. This makes me hopeful we can remove it elsewhere.

src/main/java/marquez/api/resources/DatasetResource.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @GET
+  @ResponseMetered
+  @ExceptionMetered
+  @Timed
+  @Path("/namespaces/{namespace}/datasets")
+  @Produces(APPLICATION_JSON)
+  public Response list(
+      @PathParam("namespace") String namespaceString,
```
ashulmanWeWork: Could potentially use `@NotBlank`

src/main/java/marquez/api/resources/DatasetResource.java
```diff
@@ -0,0 +1,60 @@
...
+      @QueryParam("offset") @DefaultValue("0") Integer offset)
+      throws UnexpectedException, WebApplicationException {
+    if (!namespaceService.exists(namespaceString)) {
+      throw new WebApplicationException(
+          String.format("The namespace %s does not exist.", namespaceString), NOT_FOUND);
+    }
+    final List<Dataset> datasets =
+        datasetService.getAll(Namespace.of(namespaceString), limit, offset);
+    final List<DatasetResponse> datasetResponses = DatasetResponseMapper.map(datasets);
+    return Response.ok(new DatasetsResponse(datasetResponses)).build();
```
ashulmanWeWork: Nice - I like that we're moving to one <Resource>Response object that is returned for gets, creates, (and potentially updates)!

src/main/java/marquez/common/models/ConnectionUrl.java
```diff
@@ -0,0 +1,92 @@
...
+
+@EqualsAndHashCode
+@ToString
+public final class ConnectionUrl {
+  @Getter private final DataSource dataSource;
+  @Getter private final Db db;
+  @Getter private final String rawValue;
+
+  private ConnectionUrl(
+      @NonNull final DataSource dataSource, @NonNull final Db db, @NonNull final String rawValue) {
```
ashulmanWeWork: `@NotBlank` might be better here for `rawValue`.

src/main/java/marquez/common/models/ConnectionUrl.java
```diff
@@ -0,0 +1,92 @@
...
+      private static final int DATA_SOURCE_PART = 1;
+      private static final int PORT_AND_DB_PART = 3;
+      private static final String PORT_AND_DB_PART_DELIM = "/";
+      private static final int DB_PART = 1;
+      private static final String DB_PART_DELIM = ";";
+      private static final int DB_PART_NO_PARAMS = 0;
+
+      @Override
+      public ConnectionUrl parse(@NonNull String value) {
+        final String[] urlParts = value.split(URL_DELIM);
```
ashulmanWeWork: Would a regex check help, instead of writing the validation yourself?

src/main/java/marquez/db/DataSourceDao.java
```diff
@@ -0,0 +1,26 @@
...
+  @SqlUpdate(
+      "INSERT INTO data_sources (uuid, data_source, connection_url) "
+          + "VALUES (:uuid, :data_source, :connection_url)")
+  void insert(@BindDataSourceRow DataSourceRow dataSourceRow);
+
+  @SqlQuery("SELECT * FROM data_sources WHERE uuid = :uuid")
+  Optional<DataSourceRow> findBy(@Bind("uuid") UUID uuid);
+
+  @SqlQuery("SELECT * FROM data_sources LIMIT :limit OFFSET :offset")
+  List<DataSourceRow> findAll(@Bind("limit") Integer limit, @Bind("offset") Integer offset);
```
ashulmanWeWork: Should we add a constraint that the limit be >0 and potentially provide a default value?

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -0,0 +1,66 @@
...
+      @Bind("updatedAt") Instant updatedAt,
+      @Bind("currentVersion") UUID currentVersion);
+
+  @SqlQuery("SELECT * FROM datasets WHERE uuid = :uuid")
+  Optional<DatasetRow> findBy(@Bind("uuid") UUID uuid);
+
+  @SqlQuery("SELECT * FROM datasets WHERE urn = :datasetUrn.value")
+  Optional<DatasetRow> findBy(@BindBean("datasetUrn") DatasetUrn datasetUrn);
+
+  @SqlQuery("SELECT * FROM datasets LIMIT :limit OFFSET :offset")
```
ashulmanWeWork: usability thought - for user-friendliness, could we order dataset results by `createdAt` or `updatedAt` DESC?

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -0,0 +1,66 @@
...
+  List<DatasetRow> findAll(
+      @BindBean("namespace") Namespace namespace,
+      @Bind("limit") Integer limit,
+      @Bind("offset") Integer offset);
```
ashulmanWeWork: for user-friendliness, add a default of 0?

src/main/java/marquez/db/sql/BindDataSourceRow.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.jdbi.v3.sqlobject.customizer.SqlStatementCustomizingAnnotation;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER})
+@SqlStatementCustomizingAnnotation(BindDataSourceRowFactory.class)
```
ashulmanWeWork: Very cool!

src/main/java/marquez/db/sql/BindDatasetRowFactory.java
```diff
@@ -0,0 +1,31 @@
...
+      Parameter param,
+      int index,
+      Type type) {
+    return (stmt, obj) -> {
+      final DatasetRow datasetRow = (DatasetRow) obj;
+      stmt.bind("uuid", datasetRow.getUuid().toString())
+          .bind("namespace_uuid", datasetRow.getNamespaceUuid().toString())
+          .bind("datasource_uuid", datasetRow.getDataSourceUuid().toString())
+          .bind("urn", datasetRow.getUrn().getValue())
+          .bind("description", datasetRow.getDescription().orElse(NO_DESCRIPTION).getValue());
```
ashulmanWeWork: You have this pattern in a few places. Why not just implement `getDescription` to return either its description, or `NO_DESCRIPTION`?

src/main/java/marquez/service/DatasetService.java
```diff
@@ -0,0 +1,71 @@
...
+    final DatasetRow datasetRow = DatasetRowMapper.map(namespace, dataSourceRow, dbTableVersion);
+    final DbTableInfoRow dbTableInfoRow = DbTableInfoRowMapper.map(dbTableVersion);
+    final DbTableVersionRow dbTableVersionRow =
+        DbTableVersionRowMapper.map(datasetRow, dbTableInfoRow, dbTableVersion);
+    try {
+      datasetDao.insertAll(dataSourceRow, datasetRow, dbTableInfoRow, dbTableVersionRow);
+      final Optional<DatasetRow> datasetRowIfFound = datasetDao.findBy(datasetRow.getUuid());
+      return datasetRowIfFound.map(DatasetMapper::map).orElseThrow(UnexpectedException::new);
+    } catch (UnableToExecuteStatementException e) {
+      log.error(e.getMessage());
```
ashulmanWeWork: Are we logging things now as `log.error(e.getMessage(), e)`?

src/main/java/marquez/service/models/Dataset.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+import marquez.common.models.DatasetUrn;
+import marquez.common.models.Description;
+
+@RequiredArgsConstructor
+@EqualsAndHashCode
+@ToString
+public final class Dataset {
+  @Getter @NonNull private final DatasetUrn urn;
```
ashulmanWeWork: Maybe I'm missing something, but shouldn't this have a `UUID`?

src/test/java/marquez/MarquezAppIntegrationTest.java
```diff
@@ -31,13 +32,13 @@ public void teardown() {
...
+    final Response response =
         APP.client()
             .target(URI.create("http://localhost:" + APP.getLocalPort()))
             .path("/ping")
             .request()
             .get();
-    assertEquals(200, res.getStatus());
-    assertEquals("pong", res.readEntity(String.class));
+    assertEquals(OK.getStatusCode(), response.getStatus());
+    assertEquals("pong", response.readEntity(String.class));
```
ashulmanWeWork: :+1:

src/test/java/marquez/api/mappers/DatasetResponseMapperTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Test
+  public void testMapDatasetResponseNoDescription() {
+    final Optional<String> noDescriptionString = Optional.of(NO_DESCRIPTION.getValue());
+    final Dataset dataset = new Dataset(DATASET_URN, CREATED_AT, NO_DESCRIPTION);
+    final DatasetResponse datasetResponse = DatasetResponseMapper.map(dataset);
+    assertNotNull(datasetResponse);
+    assertEquals(DATASET_URN.getValue(), datasetResponse.getUrn());
+    assertEquals(CREATED_AT.toString(), datasetResponse.getCreatedAt());
+    assertEquals(noDescriptionString, datasetResponse.getDescription());
```
ashulmanWeWork: Maybe also worth asserting that `Optional.of(NO_DESCRIPTION.getValue());` is equal to the `NO_VALUE` string and that `datasetResponse.getDescription()` is also equal to that.

src/test/java/marquez/api/resources/DatasetResourceTest.java
```diff
@@ -0,0 +1,76 @@
...
+import marquez.service.DatasetService;
+import marquez.service.models.Dataset;
+import org.junit.Test;
+
+public class DatasetResourceTest {
+  private static final DatasetUrn DATASET_URN = DatasetUrn.of("urn:a:b.c");
+  private static final Instant CREATED_AT = Instant.now();
+  private static final Namespace NAMESPACE = Namespace.of("test");
+  private static final Integer LIMIT = 100;
+  private static final Integer OFFSET = 0;
```
ashulmanWeWork: Although these parameters don't come from the user, it still might be worth adding a couple negative tests for invalid values of limit and offset, at least to guarantee there'll be test noise if we inadvertently start to ignore/misuse those values at the DAO layer.

src/main/java/marquez/common/models/ConnectionUrl.java
```diff
@@ -0,0 +1,92 @@
...
+    final ValueParser valueParser = ValueParser.get(trimmed);
+    if (valueParser == ValueParser.UNKNOWN) {
+      throw new IllegalArgumentException(
+          String.format("Failed to parse connection url value '%s', unknown data source.", value));
+    }
+    return valueParser.parse(trimmed);
+  }
+
+  private enum ValueParser {
+    JDBC("jdbc") {
```
sshah-wework: does this mean we only support `jdbc://` URIs?

src/main/java/marquez/common/models/ConnectionUrl.java
```diff
@@ -0,0 +1,92 @@
...
+                dbString.contains(DB_PART_DELIM)
+                    ? dbString.split(DB_PART_DELIM)[DB_PART_NO_PARAMS]
+                    : dbString);
+        return new ConnectionUrl(dataSource, db, value);
+      }
+    },
+    UNKNOWN("") {
+      @Override
+      public ConnectionUrl parse(@NonNull String value) {
+        throw new UnsupportedOperationException();
```
sshah-wework: is this the default case if the parser doesn't match the above cases?

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -0,0 +1,66 @@
...
+  @CreateSqlObject
+  DataSourceDao createDataSourceDao();
+
+  @CreateSqlObject
+  DbTableVersionDao createDbTableVersionDao();
+
+  @SqlUpdate(
+      "INSERT INTO datasets (uuid, namespace_uuid, datasource_uuid, urn, description) "
+          + "VALUES (:uuid, :namespace_uuid, :datasource_uuid, :urn, :description)")
+  void insert(@BindDatasetRow DatasetRow datasetRow);
```
sshah-wework: what does `@BindDatasetRow` offer that `@BindBean` wouldnt? If we carry this pattern forward, will we need to write custom `Bind*Row` and `Bind*RowFactory` classes for every DAO model? Worth the calories?

Issue #200, opened 2018-12-18
Add a test to verify that job uniqueness within a namespace, but not across namespaces


ashulmanWeWork:
Stack trace from recent run:
```
ERROR [2018-12-18 00:28:12,063] marquez.core.services.JobService: failed to create new job
! org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "jobs_name_key"
!   Detail: Key (name)=(job1) already exists.
! at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2440)
! at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2183)
! at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:308)
! at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:441)
! at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:365)
! at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:143)
! at org.postgresql.jdbc.PgPreparedStatement.execute(PgPreparedStatement.java:132)
! at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
! at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
! at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
! at java.lang.reflect.Method.invoke(Method.java:498)
! at org.apache.tomcat.jdbc.pool.StatementFacade$StatementProxy.invoke(StatementFacade.java:114)
! at com.sun.proxy.$Proxy79.execute(Unknown Source)
! at org.jdbi.v3.core.statement.SqlLoggerUtil.wrap(SqlLoggerUtil.java:29)
! at org.jdbi.v3.core.statement.SqlStatement.internalExecute(SqlStatement.java:1434)
! ... 98 common frames omitted
! Causing: org.jdbi.v3.core.statement.UnableToExecuteStatementException: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "jobs_name_key"
[...]

PR #205, opened 2018-12-18
Redefine unique jobs idx

sshah-wework:
- redefine unique index on `jobs(namespace_guid, name)`
- verify with DAO tests

### DIFF
--- a/src/test/java/marquez/dao/JobDAOTest.java
+++ b/src/test/java/marquez/dao/JobDAOTest.java
@@ -3,7 +3,6 @@ package marquez.dao;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
 
 import java.util.Arrays;
 import java.util.Collections;
@@ -55,8 +54,8 @@ public class JobDAOTest {
             handle -> {
               handle.execute(
                   "INSERT INTO namespaces(guid, name, current_ownership)" + "VALUES (?, ?, ?);",
-                  nsID,
-                  nsName,
+                  namespaceId,
+                  name,
                   ownerName);
             });
   }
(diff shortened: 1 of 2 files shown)


src/test/java/marquez/dao/JobDAOTest.java
```diff
@@ -89,6 +97,39 @@ public void testInsert() {
...
+              UUID.randomUUID(),
+              job.getName(),
+              "location",
+              newNamespaceId,
+              "desc",
+              Collections.<String>emptyList(),
+              Collections.<String>emptyList());
+      jobDAO.insert(jobWithDiffNsSameName);
+    } catch (UnableToExecuteStatementException e) {
+      fail("failed to insert a job with the same name into a different namespace");
```
ashulmanWeWork: This is a little confusing. You are attempting to insert a job with the same name but a different NS than an existing job. This should work and your test should just pass. 

It also makes sense to me that you're catching the exception it can throw, and then marking the test as failed. It's not a necessary thing to do, since the exception being thrown is enough to fail the test, but some people like to do it that way and that's fine by me.

What's not clear to me is why you are expecting the UnableToExecuteStatementException. First off, is the test intending to demonstrate that this will always throw an exception? If so, then I don't think that's the agreed-upon behavior.
Secondly, since it's being expected, are you expecting it from the namespace creation request or first job creation request, since the second job creation attempt will catch it? If so, what would that demonstrate?

src/test/java/marquez/dao/JobDAOTest.java
```diff
@@ -89,6 +97,39 @@ public void testInsert() {
...
+            UUID.randomUUID(),
+            job.getName(),
+            "location",
+            job.getNamespaceGuid(),
+            "desc",
+            Collections.<String>emptyList(),
+            Collections.<String>emptyList());
+    jobDAO.insert(jobWithSameNsSameName);
+  }
+
```
ashulmanWeWork: This should be two test cases, since you're trying to prove at least two things here.

1. Same job name, diff NS is ok.
2. Same job name, same ns is NOT ok.

The test is trying to do too much and is considerably less clear than two single-intent testes.

sshah-wework: sure, can break it up

ashulmanWeWork:
This resolves Issue #200.

PR #211, opened 2018-12-20
Api model cleanup

ashulmanWeWork:


src/main/java/marquez/api/resources/JobResource.java
```diff
@@ -214,17 +211,15 @@ public Response abortJobRun(@PathParam("runId") final String runId) throws Resou
...
+  @Path("/jobs/runs/{runId}/abort")
+  public Response abortJobRun(@PathParam("runId") final String runId) throws ResourceException {
     try {
-      Optional<marquez.core.models.JobRun> jobRun = jobService.getJobRun(runId);
+      final Optional<JobRun> jobRun = jobService.getJobRun(UUID.fromString(runId));
       if (jobRun.isPresent()) {
-        return Response.status(Response.Status.OK)
-            .entity(new CoreJobRunToApiJobRunMapper().map(jobRun.get()))
-            .build();
+        jobService.updateJobRunState(UUID.fromString(runId), JobRunState.State.ABORTED);
```
sshah-wework: interesting...this is the resource layer reaching into the `JobRunState` core model. `JobService` might need to make this easier for the resource layer

src/main/java/marquez/core/mappers/CoreJobToApiJobMapper.java
```diff
@@ -6,12 +6,15 @@
 public class CoreJobToApiJobMapper extends Mapper<marquez.core.models.Job, Job> {
   @Override
   public Job map(marquez.core.models.Job value) {
-    // TODO: Update to add real values for input and output datasets
     return new Job(
         value.getName(),
         value.getCreatedAt(),
-        Collections.emptyList(),
-        Collections.emptyList(),
+        value.getInputDatasetUrns().isEmpty()
```
sshah-wework: is this ternary necessary? won't it accomplish the same thing if you just pass `value.getInputDatasetUrns()`? It will either be an empty, or a non-empty list

src/main/java/marquez/core/mappers/CoreJobToApiJobMapper.java
```diff
@@ -6,12 +6,15 @@
...
+            ? Collections.emptyList()
+            : value.getInputDatasetUrns(),
+        value.getOutputDatasetUrns().isEmpty()
```
sshah-wework: same here

src/main/java/marquez/dao/JobRunStateRow.java
```diff
@@ -3,17 +3,17 @@
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.UUID;
+import marquez.core.models.JobRunState;
```
sshah-wework: this is a good change, but seems totally unrelated to this PR

src/test/java/marquez/api/JobIntegrationTest.java
```diff
@@ -151,18 +154,52 @@ public void testJobRunAfterUpdateEndToEnd() {
             .put(Entity.json(""));
 
     assertEquals(Response.Status.OK.getStatusCode(), res.getStatus());
+
+    final JobRunResponse getJobRunResponse = getJobRunApiResponse(CREATED_JOB_RUN_UUID);
+    assertThat(getJobRunResponse.getRunState()).isEqualTo(JobRunState.State.COMPLETED.name());
+  }
+
+  @Test
+  public void testJobRunAfterMarkedFailedEndToEnd() {
```
sshah-wework: is this just to increase test coverage which got missed in the previous PR?

src/main/java/marquez/core/mappers/CoreJobToApiJobMapper.java
ashulmanWeWork: Good point. Removed :)

src/main/java/marquez/dao/JobRunStateRow.java
ashulmanWeWork: Yup, you're right. Will keep PRs discrete.

src/test/java/marquez/api/JobIntegrationTest.java
ashulmanWeWork: Yup!

src/main/java/marquez/api/resources/JobResource.java
ashulmanWeWork: Yea, agreed, it almost feels like breaking encapsulation. What we have is good, but it would feel more natural to ask the service to mark the job as {failed, aborted, completed}, but not prescribing that it should cause this field to change value.

PR #212, opened 2018-12-20
Return Job from database after database update

### LINKED ISSUE #188: `createdAt` field not set in the service layer for newly-created Jobs

sshah-wework:
Initially, createJob would short-circuit a database read after inserting a new job.
With this change, the Job is always re-read from the database to capture any implicitly
updated fields.

Closes #188.

### DIFF
--- a/src/test/java/marquez/core/services/JobServiceIntegrationTest.java
+++ b/src/test/java/marquez/core/services/JobServiceIntegrationTest.java
@@ -66,7 +66,8 @@ public class JobServiceIntegrationTest {
   @Test
   public void testCreate() throws UnexpectedException {
     Job job = Generator.genJob(namespaceID);
-    jobService.createJob(namespaceName, job);
+    Job jobReturned = jobService.createJob(namespaceName, job);
+    assertNotNull(jobReturned.getCreatedAt());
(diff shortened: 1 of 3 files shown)


src/test/java/marquez/core/services/JobServiceIntegrationTest.java
```diff
@@ -66,7 +66,8 @@ public void tearDown() {
   @Test
   public void testCreate() throws UnexpectedException {
     Job job = Generator.genJob(namespaceID);
-    jobService.createJob(namespaceName, job);
+    Job jobReturned = jobService.createJob(namespaceName, job);
+    assertNotNull(jobReturned.getCreatedAt());
     Optional<Job> jobFound = jobService.getJob(namespaceName, job.getName());
     assertTrue(jobFound.isPresent());
     assertEquals(job.getName(), jobFound.get().getName());
```
ashulmanWeWork: Would also be cool to test if the result of a create operation is equal to the result of a get operation. I can't think of a reason why they wouldn't be equal.

sshah-wework: Makes sense, we can add this

PR #213, opened 2018-12-20
Compare createjob result with getjob

sshah-wework:


src/test/java/marquez/core/services/JobServiceIntegrationTest.java
```diff
@@ -66,23 +66,50 @@ public void tearDown() {
...
+        jobService.createJob(namespaceName, jobWithNewLoc); // should create new version implicitly
+    Optional<Job> jobGetRet = jobService.getJob(namespaceName, job.getName());
+    assertTrue(jobGetRet.isPresent());
+    assertEquals(jobGetRet.get(), jobCreateRet);
+    assertEquals(job.getName(), jobGetRet.get().getName());
+    assertEquals(jobCreateRet, jobGetRet.get());
+    List<JobVersion> versions = jobService.getAllVersionsOfJob(namespaceName, job.getName());
+    assertEquals(jobGetRet.get().getGuid(), versions.get(0).getJobGuid());
+    assertEquals(jobGetRet.get().getGuid(), versions.get(1).getJobGuid());
+  }
```
ashulmanWeWork: Can also assert that the latest version of the job points to `versions.get(1).getGuid()`.
You can also consider making it a separate test if this one is getting too big.

ashulmanWeWork: On that thought, might be interesting to make explicit the ordering of the list that comes back in versions.

sshah-wework: i think that's out of scope for this test

sshah-wework: but a good thing to test eventually, right now `0.1.0` makes no claims about the order of versions

PR #223, opened 2018-12-29
Use 'id' as path variable in job runs endpoint

wslulciuc:
This PR shortens the path variable name `jubId` to just `id` for updating the state of a job run (the `run` in the name is redundant):

```
GET   /api/v1/jobs/runs/{id} (marquez.api.resources.JobResource)
PUT   /api/v1/jobs/runs/{id}/abort (marquez.api.resources.JobResource)
PUT   /api/v1/jobs/runs/{id}/complete (marquez.api.resources.JobResource)
PUT   /api/v1/jobs/runs/{id}/fail (marquez.api.resources.JobResource)
```


src/main/java/marquez/api/resources/JobResource.java
```diff
@@ -213,8 +213,8 @@ public Response failJobRun(@PathParam("runId") final String runId) throws Resour
 
   @PUT
   @Timed
-  @Path("/jobs/runs/{runId}/abort")
-  public Response abortJobRun(@PathParam("runId") final String runId) throws ResourceException {
+  @Path("/jobs/runs/{id}/abort")
+  public Response abortJobRun(@PathParam("id") final String runId) throws ResourceException {
```
ashulmanWeWork: This will break integrations, but it's not a hard change to adjust to,.

wslulciuc: Which integration test would break?

ashulmanWeWork: Sorry, on second thought, it wouldn't break any tests b/c it's a pathparam, defined right above. This change is fine.

Issue #224, opened 2018-12-29
Move all daos and services to correct pkg


wslulciuc:
Let's move the following classes to the new pkg structure introduced in #100 :

### `core` -> [`service`](https://github.com/MarquezProject/marquez/tree/master/src/main/java/marquez/service) pkg

```
.core/models/JobRun.java
.core/models/Namespace.java
.core/models/Job.java
.core/models/RunArgs.java
.core/models/JobVersion.java
.core/models/JobRunState.java
.core/services/JobService.java
.core/services/NamespaceService.java
```

### `dao` -> [`db`](https://github.com/MarquezProject/marquez/tree/master/src/main/java/marquez/db) pkg

```
JobDAO.java -> JobDao.java
JobRunDAO.java -> JobRunDao.java
JobRunStateDAO.java -> JobRunStateDao.java
JobVersionDAO.java -> JobVersionDao.java
NamespaceDAO.java -> NamespaceDao.java
RunArgsDAO.java -> JobRunArgsDao.java
```

This will allow us to remove the `core` and `dao` pkgs.

wslulciuc:
@sshah-wework : Assigning this to you as you have been focusing on implementing the `JobService`.

PR #226, opened 2018-12-29
Add job row classes

wslulciuc:
This PR adds all db models for accessing a row in the following tables:

* `jobs`
* `job_versions`
* `job_runs`
* `job_run_args`
* `job_run_states`

**NOTE:** This PR is depended on issue #224 in order to allow for proper test coverage

### DIFF
--- a/src/test/java/marquez/db/models/JobRunRowTest.java
+++ b/src/test/java/marquez/db/models/JobRunRowTest.java
@@ -1,0 +1,124 @@
+package marquez.db.models;
+
+import static org.junit.Assert.assertEquals;
+
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+import org.junit.Test;
+
+public class JobRunRowTest {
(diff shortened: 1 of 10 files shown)


src/test/java/marquez/db/models/JobRunRowTest.java
```diff
@@ -94,6 +94,20 @@ public void testNewJobRow_nullJobVersionUuid() {
         .build();
   }
 
+  @Test(expected = NullPointerException.class)
```
ashulmanWeWork: Maybe it's better to throw an IllegalStateException instead of an NPE?

wslulciuc: This is during object instantiation, meaning there's no _initial_ state. With a null check, we are enforcing a constraint (or precondition) on what a parameter value can or cannot be i.e. _it cannot be null_

If you look at the JavaDoc for **IllegalStateException** it's not what we are checking for:
> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.

PR #231, opened 2019-01-03
[MARQUEZ-195]: Remove job_run_definitions table

ashulmanWeWork:


wslulciuc:
@ashulmanWeWork : Based on our offline discussion, can we close this PR? as we'll want to consolidating all Flyway migration  in a single `.sql` script?

PR #234, opened 2019-01-04
Closes #216 - Add 'running' endpoint

### LINKED ISSUE #216: Add job run endpoint to mark a job run as running
I want Marquez to have an endpoint to signify that a job run is running, as opposed to merely created (i.e. new). Often there isn't much of a delay between a job run being created and it actually transitioning to running, but that may be valuable information to capture. The delta between new and running could also be a valuable metric.

Spec would be:
/api/v1/jobs/runs/{job_run}/run

Open to other suggestions.

ashulmanWeWork:


docs/openapi.yml
```diff
@@ -241,6 +241,18 @@ paths:
               schema:
                 $ref: '#/components/schemas/jobRun'
 
+  /jobs/runs/{id}/start:
```
wslulciuc: This will also have to be uodated

PR #246, opened 2019-01-15
Drop 'v' prefix from docker img tag

wslulciuc:
This PR is a continuation of #240 removing the `v` prefix from our tagged releases (see https://semver.org)


ashulmanWeWork:
https://skymeteor.com/85259/drop-the-the-social-network-drop-the

PR #260, opened 2019-01-26
Apply YAML fix

ashulmanWeWork:


docs/openapi.yml
```diff
@@ -297,7 +294,6 @@ paths:
       description: Creates a multiple output dataset objects.
       tags:
         - Jobs
-      required: true
```
wslulciuc: Where these `required` fields not used?

ashulmanWeWork: The location here doesn't support this field per the spec.

wslulciuc: 👍

PR #271, opened 2019-02-04
Add preconditions

wslulciuc:
This PR adds the following assertions in order to avoid external dependencies:

- `checkNotNull()`
- `checkNotBlank()`
- `checkArgument()`

### Usage

```java
private RunId(@NonNull final String value) {
  checkNotBlank(value, "value must not be blank or empty");
  checkArgument(value.length() != ID_LENGTH, String.format("value length must = %d", ID_LENGTH))

  this.value = UUID.fromString(value);
}
```

Inspired by https://github.com/google/guava/wiki/PreconditionsExplained


src/test/java/marquez/common/PreconditionsTest.java
```diff
@@ -0,0 +1,121 @@
...
+  private static final Object NULL_REFERENCE = null;
+  private static final Object NON_NULL_REFERENCE = new Object();
+  private static final String NULL_ERROR_MESSAGE = null;
+  private static final String NON_NULL_ERROR_MESSAGE = "test error message";
+  private static final String BLANK_STRING = " ";
+  private static final String NON_BLANK_STRING = "test string";
+  private static final String NULL_STRING = null;
+  private static final String NON_NULL_STRING = NON_BLANK_STRING;
+  private static final boolean TRUE = 0 < 1;
+  private static final boolean FALSE = 0 > 1;
```
ashulmanWeWork: why are you redefining True and False?

PR #323, opened 2019-02-16
Update data source row mapper test

wslulciuc:
This PR makes the following updates to [`DataSourceRowMapperTest`](https://github.com/MarquezProject/marquez/blob/master/src/test/java/marquez/service/mappers/DataSourceRowMapperTest.java):

- Add `UnitTests` category
- Add clearer test descriptions

### DIFF
--- a/src/test/java/marquez/service/mappers/DataSourceRowMapperTest.java
+++ b/src/test/java/marquez/service/mappers/DataSourceRowMapperTest.java
@@ -13,7 +13,10 @@ import marquez.common.models.Description;
 import marquez.db.models.DataSourceRow;
 import marquez.service.models.DbTableVersion;
 import org.junit.Test;
+import marquez.UnitTests;
+import org.junit.experimental.categories.Category;
 
+@Category(UnitTests.class)
 public class DataSourceRowMapperTest {
   private static final DataSource DATA_SOURCE = DataSource.fromString("postgresql");
   private static final ConnectionUrl CONNECTION_URL =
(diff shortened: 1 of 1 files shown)


wslulciuc:
@sshah-wework : We can link this test as a good example on how to write tests for our service mappers (_service model_ => _db model_) i.e. open issues addressing test coverage.

src/test/java/marquez/service/mappers/DataSourceRowMapperTest.java
```diff
@@ -1,57 +1,46 @@
...
   private static final DataSource DATA_SOURCE = DataSource.fromString("postgresql");
+  private static final DbName DB_NAME = DbName.fromString("test_db");
   private static final ConnectionUrl CONNECTION_URL =
       ConnectionUrl.fromString(
-          String.format("jdbc:%s://localhost:5432/novelists", DATA_SOURCE.getValue()));
-  private static final DbSchemaName DB_SCHEMA_NAME = DbSchemaName.fromString("marquez");
-  private static final DbTableName DB_TABLE_NAME = DbTableName.fromString("quotes");
-  private static final Description DESCRIPTION =
-      Description.fromString("It's enough for me to be sure that you and I exist as this moment.");
+          String.format("jdbc:%s://localhost:5432/%s", DATA_SOURCE.getValue(), DB_NAME.getValue()));
```
ashulmanWeWork: Are you sure we want to change the description here? Seems like the current value more closely resembles what it would be than the new value?

wslulciuc: The diff is hard to read, but the description isn't changed, but rather removed. We are setting this value to `CONNECTION_URL`. I'm going to review test to make sure I didn't remove `DESCRIPTION` unintentionally (which might be the case, actually).

wslulciuc: The description was removed because a `DataSourceRow` does not have a description.

ashulmanWeWork: Ah, right. Sorry about that!

PR #324, opened 2019-02-16
Update data source row class

wslulciuc:
This PR explicitly marks `DataSourceRow.createdAt` as  nullable with `@Nullable`. The getter has  also been updated to return an optional. This change also required adding unit tests (yeah!).


wslulciuc:
**NOTE:** This PR is dependent on #333

ashulmanWeWork:
Added @roaraya8 as a reviewer in case he had any feedback.

PR #333, opened 2019-02-18
Add column constants

wslulciuc:
This PR defines all column names as constants in [`Columns`](https://github.com/MarquezProject/marquez/blob/db-columns/src/main/java/marquez/db/Columns.java). The following set of helper methods have also been added:
- **`toInstantOrNull(String)`**
- **`toUuidOrNull(String)`**
- **`toList(Array)`**


src/main/java/marquez/db/Columns.java
```diff
@@ -0,0 +1,60 @@
...
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+
+public final class Columns {
+  public static final String ROW_UUID = "guid";
```
ashulmanWeWork: nit: might be good to alphabetize this list.

wslulciuc: Good point. I'll break down the list by table: `namespace`, `owner`, etc

PR #334, opened 2019-02-18
Db utils

wslulciuc:


src/test/java/marquez/db/UtilsTest.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testToUuidOrNull_throwsException_onBlankUuidString() {
+    final String blankUuidString = " ";
+    Utils.toUuidOrNull(blankUuidString);
+  }
+
+  @Test
+  public void testToListOrEmpty() throws SQLException {
```
ashulmanWeWork: I don't think this function name matches the test (at least the empty list part). It's important to add a test for an empty list, and I don't think I see that case exercised. Can we add it?

PR #335, opened 2019-02-19
Adjust to use class-level setup instead of method-level setup

ashulmanWeWork:
Brought this up during the review of this file, but since this is a 1-method test class (for now), it won't make a meaningful difference. Still, this is a good practice to follow and wanted to bring attention to it.

### DIFF
--- a/src/test/java/marquez/db/DatasetDaoTest.java
+++ b/src/test/java/marquez/db/DatasetDaoTest.java
@@ -11,6 +11,7 @@ import marquez.service.models.Generator;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 import org.jdbi.v3.testing.JdbiRule;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
@@ -20,7 +21,7 @@ import org.slf4j.LoggerFactory;
 public class DatasetDaoTest {
   static final Logger logger = LoggerFactory.getLogger(DatasetDaoTest.class);
   private static DatasetDao datasetDAO;
-  private static marquez.service.models.Namespace namespace = Generator.genNamespace();
+  private static marquez.service.models.Namespace namespace;
   private static NamespaceDao namespaceDAO;
 
   @ClassRule
@@ -30,8 +31,13 @@ public class DatasetDaoTest {
   @BeforeClass
   public static void setup() {
     Jdbi jdbi = dbRule.getJdbi();
-    datasetDAO = jdbi.onDemand(DatasetDao.class);
     namespaceDAO = jdbi.onDemand(NamespaceDao.class);
+    datasetDAO = jdbi.onDemand(DatasetDao.class);
+  }
+
+  @Before
+  public void setupTest() {
+    namespace = Generator.genNamespace();
     namespaceDAO.insert(namespace);
   }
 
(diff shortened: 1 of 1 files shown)


sshah-wework:
If the database is being moved to class-level, you have to remember to clean up state. there's only one test method here, so it passes but if there were more it would cause failures

sshah-wework:
We are likely doing this elsewhere in the code, so might be worth fixing up the same thing across all tests in a single PR

sshah-wework:
But thanks for following up and being proactive about the cleanup! 👍 😎

ashulmanWeWork:
Good call on the cleanup. We should add that to our tech debt backlog wherever it slots in in terms of priority.

sshah-wework:
Hm, this PR is not in a place to merge without the DB cleanup, it is a very partial implementation

ashulmanWeWork:
Yep, I'll add some cleanup code in this PR and that'll be a reference for future tech-debt-reduction work.

ashulmanWeWork:
@sshah-wework, I thought about what cleanup code might be necessary. However, since all the created test objects have random UUIDs (namespaces and datasets), it might not be necessary at all to delete the namespaces or datasets that are created. Instead, for each test case, we'd create a fresh namespace, which solves the problem of re-using a polluted namespace. Creating a single namespace is still considerably more economical that the existing per-test setup code. This change is also transparent to the test-writer. All they know is that they have a fresh namespace available.

Further, the DAOs don't have utilities to delete namespaces or datasets and adding DB connections to delete the DB is also messy and prone to regression.

To demonstrate the effectiveness of the technique I've chosen, I added an additional test. Given this, what do you think of this approach I've provided in the patch?

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -12,26 +12,32 @@
...
-  private marquez.service.models.Namespace namespace = Generator.genNamespace();
-  private NamespaceDao namespaceDAO;
+  private static DatasetDao datasetDAO;
+  private static marquez.service.models.Namespace namespace;
+  private static NamespaceDao namespaceDAO;
 
-  @Rule
-  public final JdbiRule dbRule =
+  @ClassRule
+  public static final JdbiRule dbRule =
```
wslulciuc: `static` variables and methods aren't associated with an instance of `DatasetDaoTest`, is that what we want? That is, a single copy of `JdbiRule` we be shared across `DatasetDaoTest`, which for tests isn't techincally a problem, just redunt, so I would define `JdbiRule ` and `setup()` as non-static and revert the change.

ashulmanWeWork: It's required to be `static` b/c the test class setup method, where it's used, is static.

wslulciuc: OOh, right. Sorry, disregard my comment.

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -45,12 +51,23 @@ private void insertRandomDataset() {
...
     assertEquals(
         0, datasetDAO.findAll(NamespaceName.fromString(namespace.getName()), 10, 0).size());
     insertRandomDataset();
     insertRandomDataset();
     assertEquals(
         2, datasetDAO.findAll(NamespaceName.fromString(namespace.getName()), 10, 0).size());
   }
+
+  @Test
+  public void testFindAllWithLimit() {
```
wslulciuc: 🥇

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -12,26 +12,32 @@
...
-import org.junit.Rule;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class DatasetDaoTest {
   static final Logger logger = LoggerFactory.getLogger(DatasetDaoTest.class);
-  private DatasetDao datasetDAO;
```
wslulciuc: minor: the `logger` can be removed

PR #341, opened 2019-02-26
Add basic datasources resources and integration test

ashulmanWeWork:


src/main/java/marquez/api/models/DatasourceResponse.java
```diff
@@ -0,0 +1,21 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class DatasourceResponse {
+
+  @JsonProperty("name")
+  String name;
```
wslulciuc: Is there a reason we are defaulting instances variables to package-private ?

ashulmanWeWork: No specific reason. Happy to make them private. In that case, we'd also want to go back and make sure that's the case in other places, such as `NamespaceResponse`.

wslulciuc: Great. We'll want to have issues first before we start address other classes

src/main/java/marquez/api/models/DatasourceResponse.java
```diff
@@ -0,0 +1,21 @@
+package marquez.api.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
```
wslulciuc: Do we need 3 lombok annotations here? If we have `@Data`, what are the values of adding `@NoArgsConstructor` and `@AllArgsConstructor`?

src/main/java/marquez/api/models/DatasourceResponse.java
```diff
@@ -0,0 +1,21 @@
...
+public class DatasourceResponse {
+
+  @JsonProperty("name")
+  String name;
+
+  @JsonProperty("createdAt")
+  String createdAt;
+
+  @JsonProperty("connectionUrl")
+  String owner;
```
wslulciuc: Should this be: `connectionUrl `?

ashulmanWeWork: Yup - fixed in another PR which will be merged into this.

src/main/java/marquez/api/models/DatasourceResponse.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class DatasourceResponse {
+
+  @JsonProperty("name")
+  String name;
+
+  @JsonProperty("createdAt")
```
wslulciuc: We can remove `@JsonProperty` if the name remains unchanged. We also no longer need to annotate fields: https://github.com/MarquezProject/marquez/pull/342

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: https://jira.we.co/browse/DATAP-979
Also a good first issue.

ashulmanWeWork: For this particular class, it'll be handled in a subsequent PR.

src/main/java/marquez/api/models/DatasourcesResponse.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode
+@ToString
```
wslulciuc: Why not just use `@Data`?

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: I was following the NamespaceResponse example, but agree that a cleanup to only have what's necessary is a good idea.

src/test/java/marquez/api/DatasourceIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.net.URI;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import marquez.api.models.DatasourcesResponse;
+import marquez.db.fixtures.AppWithPostgresRule;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class DatasourceIntegrationTest {
```
wslulciuc: Can you provide a unit test for `DatasourceResource` first?

src/main/java/marquez/api/models/DatasourceResponse.java
wslulciuc: Great. I opened an issue to capture the changes we'll need to make across API models: #354

src/main/java/marquez/api/models/DatasourcesResponse.java
wslulciuc: You can follow the recommended approach for writing response payloads in #355

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: You're right - if we add the `@Getter` annotation to each member, then we can remove the `@Data` annotation.

That said, the approach advocated in #354 doesn't seem to work here. This might be b/c this is a Response object. However, I'll give it a try on `DatasourceRequest` when I add it.

If we remove the AllArgsConstructor, we get compilation issues when building the DatasourceResponse object. 

If we remove the NoArgsConstructor, we get this issue that we saw before:
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `marquez.api.models.DatasourceResponse` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 18] (through reference chain: marquez.api.models.DatasourcesResponse["datasources"]->java.util.ArrayList[0])
```

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: Nice!

src/main/java/marquez/api/models/DatasourceResponse.java
wslulciuc: Oh, sorry. I linked the wrong issue :P #355

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: Please see PRs #394, #387, and #384, where this is fixed.

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: Please see PRs #394, #387, and #384, where this is fixed.

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: Please see PRs #394, #387, and #384, where this is fixed.

src/main/java/marquez/api/models/DatasourceResponse.java
ashulmanWeWork: Please see PRs #394, #387, and #384, where this is fixed.

src/main/java/marquez/api/models/DatasourcesResponse.java
ashulmanWeWork: Please see PRs #394, #387, and #384, where this is fixed.

src/test/java/marquez/api/DatasourceIntegrationTest.java
ashulmanWeWork: Please see PRs #394, #387, and #384, where this is fixed.

Issue #193, opened 2018-12-13
Investigate why `@NoArgsConstructor` is necessary in Lombok-ified objects


ashulmanWeWork:
Sample error we get when we do not have it:
```
INFO  [2018-12-13 15:03:36,265] org.glassfish.jersey.test.inmemory.InMemoryTestContainerFactory$InMemoryTestContainer: Creating InMemoryTestContainer configured at the base URI http://localhost:9998/
ERROR [2018-12-13 15:03:36,348] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: fdf955b6acb0f986
! com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `marquez.api.CreateJobRequest` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
!  at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 2]
! at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:67)
! at com.fasterxml.jackson.databind.DeserializationContext.reportBadDefinition(DeserializationContext.java:1451)
! at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1027)
! at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeFromObjectUsingNonDefault(BeanDeserializerBase.java:1297)
! at com.fasterxml.jackson.module.afterburner.deser.SuperSonicBeanDeserializer.deserializeFromObject(SuperSonicBeanDeserializer.java:239)
! at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:159)
! at com.fasterxml.jackson.module.afterburner.deser.SuperSonicBeanDeserializer.deserialize(SuperSonicBeanDeserializer.java:116)
! at com.fasterxml.jackson.databind.ObjectReader._bind(ObjectReader.java:1574)
! at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:965)
! at com.fasterxml.jackson.jaxrs.base.ProviderBase.readFrom(ProviderBase.java:815)
[...]

PR #349, opened 2019-02-27
Update job request

wslulciuc:
This PR makes the following changes to `CreateJobRequest`:

- Renamed to `JobRequest`
- Explicitly use `@Getter`, `@EqualsAndHashCode` and `@ToString`
- Annotate `JobRequest. description` with `@Nullable`
- Annotate construct with `@JsonCreator` to allow for jackson deserialization
- Add `JobRequest.getDescription()` returning an optional string

**Related issue:** #193

### DIFF
--- a/src/test/java/marquez/api/JobIntegrationTest.java
+++ b/src/test/java/marquez/api/JobIntegrationTest.java
@@ -30,9 +30,9 @@ import java.util.UUID;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-import marquez.api.models.JobRequest;
 import marquez.api.models.CreateJobRunRequest;
 import marquez.api.models.Job;
+import marquez.api.models.JobRequest;
 import marquez.api.models.JobRunResponse;
 import marquez.db.JobDao;
 import marquez.db.JobRunArgsDao;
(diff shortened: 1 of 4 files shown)


src/test/java/marquez/api/JobResourceTest.java
```diff
@@ -333,12 +333,12 @@ private Response getJob(String jobName) {
   }
 
   private Response insertJob(Job job) {
-    CreateJobRequest createJobRequest =
-        new CreateJobRequest(
-            job.getLocation(),
-            job.getDescription(),
+    JobRequest createJobRequest =
```
roaraya8: maybe renaming 'createJobRequest' to 'jobRequest' for consistency

src/test/java/marquez/api/JobIntegrationTest.java
```diff
@@ -255,12 +255,12 @@ private void evaluateResponse(Response res, Job inputJob) {
...
 
   private Response createJobOnNamespace(String namespace, Job job) {
-    CreateJobRequest createJobRequest =
-        new CreateJobRequest(
-            job.getLocation(),
-            job.getDescription(),
+    JobRequest createJobRequest =
+        new JobRequest(
             job.getInputDataSetUrns(),
-            job.getOutputDataSetUrns());
```
roaraya8: I guess the plan would be to later normalize  '..DataSet..' to '..Dataset..' everywhere, also for consistency

wslulciuc: Yep, exactly! Mind opening an issue to capture the change?

src/test/java/marquez/api/JobResourceTest.java
wslulciuc: Good catch!

PR #364, opened 2019-03-01
Fixed Issue-343

hjpatel16:
Removed the @JsonProperty from api models in DatasetResponse.java, DatasetsResponse.java, ErrorResponse.java, Job.java, JobRunRequest.java, JobRunResponse.java, NamespaceResponse.java, NamespacesResponse.java

### DIFF
--- a/src/main/java/marquez/api/models/DatasetsResponse.java
+++ b/src/main/java/marquez/api/models/DatasetsResponse.java
@@ -14,7 +14,6 @@
 
 package marquez.api.models;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.List;
 import lombok.Data;
 import lombok.NonNull;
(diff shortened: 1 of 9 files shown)


wslulciuc:
Thanks, @hjpatel16! Looks like you have some minor formatting issues that's causing the build to fail. They can easily be fixed with:

```
$ ./gradlew spotlessApply
```

src/main/java/marquez/api/models/DatasetsResponse.java
```diff
@@ -14,14 +14,12 @@
...
-import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.List;
 import lombok.Data;
 import lombok.NonNull;
 
 @Data
 public final class DatasetsResponse {
   @NonNull
-  @JsonProperty("datasets")
   private final List<DatasetResponse> datasetResponses;
```
wslulciuc: Unfortunately, `@JsonProperty` is required here because we're telling jackson to override the name `datasetResponses` with `datasets` when serializing the class as Json. In order to remove `@JsonProperty`, let's rename the field as:

```
List<DatasetResponse> datasets;
```

hjpatel16: Fixed it.. Thanks for the help!!

hjpatel16:
Requested changes are made. Thanks

PR #367, opened 2019-03-01
Fixed issue 344

hjpatel16:
Removed the @Bind Property from DataSourceDao, DatasetDao, DbTableInfoDao, JobDao, JobRunArgsDao, JobRunDao, JobRunStateDao, JobVersionDao, NamespaceDao

### DIFF
--- a/src/main/java/marquez/db/JobDao.java
+++ b/src/main/java/marquez/db/JobDao.java
@@ -36,7 +36,7 @@ public interface JobDao {
           + " VALUES (:guid, :name, :namespaceGuid, :description, :inputDatasetUrns, :outputDatasetUrns)")
   public void insert(@BindBean Job job);
 
-  @SqlUpdate("UPDATE jobs SET current_version_guid = :version_guid WHERE guid = :job_guid")
+  @SqlUpdate("UPDATE jobs SET current_version_guid = :currentVersionGuid WHERE guid = :jobGuid")
   public void setCurrentVersionGuid(UUID jobGuid, UUID currentVersionGuid);
 
   @Transaction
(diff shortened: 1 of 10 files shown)


src/main/java/marquez/db/JobDao.java
```diff
@@ -38,8 +37,7 @@
   public void insert(@BindBean Job job);
 
   @SqlUpdate("UPDATE jobs SET current_version_guid = :version_guid WHERE guid = :job_guid")
-  public void setCurrentVersionGuid(
-      @Bind("job_guid") UUID jobGuid, @Bind("version_guid") UUID currentVersionGuid);
+  public void setCurrentVersionGuid(UUID jobGuid, UUID currentVersionGuid);
```
wslulciuc: When binding method arguments to the SQL statement in ` @SqlUpdate` it's important to note that `@Bind` does allow for overriding what argument name it's bound to. That is, `@Bind("job_guid")` will bind `job_guid` in the SQL statement (below) to `:job_guid` instead of `jobGuid`:

```sql
UPDATE jobs SET current_version_guid = :version_guid WHERE guid = :job_guid
```
Knowing that, this is way the tests are failing. We need to update the argument name referenced in the SQL statement with that name of the parameter method.

```sql
UPDATE jobs SET current_version_guid = : currentVersionGuid WHERE guid = :jobGuid
```

see: http://jdbi.org/#_arguments

wslulciuc:
@hjpatel16: I left a comment on how to fix the failing tests so we can get your PR merge, but otherwise this looks great!

hjpatel16:
Sure, Ill fix it and push the changes

PR #377, opened 2019-03-03
[datasource-urn Part 1] Extract base Urn, refactor DatasetUrn, implem…

### LINKED ISSUE #373: Implement DatasourceUrn common model
We have a single URN parser that is designed to work with Datasets. There is a lot that can be re-used for Datasources, for example: validating the URN format and generating a valid URN from components. We should factor out what's common and use that in validating URNs for various types.

sshah-wework:
…ent DatasourceUrn

Closes #373


src/main/java/marquez/common/models/DatasetUrn.java
```diff
@@ -14,45 +14,22 @@
...
-  private static final String URN_DELIM = ":";
-  private static final String URN_PREFIX = "urn";
-  private static final String URN_REGEX =
-      String.format(
-          "^%s(%s[a-zA-Z0-9._]{%d,%d}){2}$", URN_PREFIX, URN_DELIM, URN_MIN_SIZE, URN_MAX_SIZE);
-  private static final Pattern URN_PATTERN = Pattern.compile(URN_REGEX);
+@EqualsAndHashCode(callSuper = false)
+public final class DatasetUrn extends Urn {
+  private static final int NUM_COMPONENTS = 2;
+  private static final String URN_TYPE = "dataset";
```
wslulciuc: As we define URNs, it'd be great to have a list known of types. Thoughts on defining the enum `UrnType`? So, initially, we'll have `UrnType.DATASET`, `UrnType.DATASOURCE`

sshah-wework: Hm, since we are creating an actual `*Urn` type for each urn type, does the enum provide much there? Adding a new Urn subtype would then require a change in the base `Urn`.

sshah-wework:
Yeah, I added the urn type in there (e.g. `dataset` and `datasource`) because it will help us avoid a dataset URN getting mixed up with a datasource URN. I can remove it, but it gives us a way to tell what kind of URN something is and avoid things like accidentally accepting a DatasetUrn for a DatasourceUrn.

sshah-wework:
Had some more discussion about this w/ Aleks. I actually think DatasetURN should be:

```
urn:{datasource_name}:{dataset_name}
```
... and `name` should be globally unique for all `Datasource`s to avoid something silly like having two databases named "Production DW". `{datasource_type}` is not unique enough, and two datasets could have the same name in two different datasources which happen to be the same type.

Thoughts?

src/main/java/marquez/common/models/Urn.java
```diff
@@ -0,0 +1,61 @@
...
+        String.format(
+            "^" + REGEX_PATTERN + "{%d}$",
+            URN_PREFIX,
+            urnType,
+            URN_DELIM,
+            URN_MIN_SIZE,
+            URN_MAX_SIZE,
+            numComponents);
+    return Pattern.compile(regex);
+  }
```
ashulmanWeWork: Nice!

PR #378, opened 2019-03-03
[datasource-urn Part 2] Add DatasetUrn generator, use generated Datas…

sshah-wework:
…etUrns in existing tests

### DIFF
--- a/src/test/java/marquez/service/models/Generator.java
+++ b/src/test/java/marquez/service/models/Generator.java
@@ -19,7 +19,9 @@ import java.util.Arrays;
 import java.util.Date;
 import java.util.Random;
 import java.util.UUID;
+import marquez.common.models.DatasetName;
 import marquez.common.models.DatasetUrn;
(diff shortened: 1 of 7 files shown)


src/test/java/marquez/service/models/Generator.java
```diff
@@ -166,14 +168,20 @@ public static DataSourceRow genDataSourceRow() {
         .build();
   }
 
+  // DatasetUrn
+  public static DatasetUrn genDatasetUrn() {
+    return DatasetUrn.from(
+        NamespaceName.fromString("ns" + randNum()), DatasetName.fromString("dataset" + randNum()));
```
ashulmanWeWork: There is a non-zero chance of a collision. In the past, I've just used a timestamp instead of a random.
Also, maybe it would be a good idea to have the same random value be used for the NS and the DS, b/c it might make correlation between the two easier when debugging.

PR #384, opened 2019-03-06
[Datasource Service]: Full resource-service-dao implementation

ashulmanWeWork:
Putting up this PR for the time being. I've just merged master into it, so it's up-to-date. It also has the feedback from the other two reviews. Right now it's still in progress as I add more tests and tighten it up, so maybe we should review it at EOD or early tmw.


src/test/java/marquez/service/DatasourceServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+    DatasourceRow row = Generator.genDatasourceRow();
+    when(datasourceDao.findBy(any())).thenReturn(Optional.of(row));
+
+    Datasource response =
+        datasourceService.create(
+            ConnectionUrl.fromString(row.getConnectionUrl()),
+            DatasourceName.fromString(row.getName()));
+    assertThat(response.getConnectionUrl())
+        .isEqualTo(ConnectionUrl.fromString(row.getConnectionUrl()));
+  }
```
ashulmanWeWork: Will add more testing in this file.

src/test/java/marquez/api/JobRunResponseTest.java
```diff
@@ -78,7 +78,7 @@ public void testHashCodeEquality() {
   }
 
   @Test
-  public void testJobRunInequalityOnUUID() {
+  public void testJobRunInequalityOnUuid() {
     JobRunResponse jr2 =
```
ashulmanWeWork: This was a recommended change for my code in one of the other PRs, so I decided to also change it in the last couple places in the project where this was the case.

ashulmanWeWork: Fixed in the reviews stacked atop this.

src/main/java/marquez/api/resources/DatasourceResource.java
```diff
@@ -0,0 +1,106 @@
...
+    this.datasourceService = datasourceService;
+  }
+
+  @POST
+  @ResponseMetered
+  @ExceptionMetered
+  @Timed
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  public Response create(@NonNull final DatasourceRequest datasourceRequest)
```
ashulmanWeWork: @wslulciuc should we potentially call this `request` instead of `datasourceRequest`?

src/main/java/marquez/api/mappers/DatasourceResponseMapper.java
```diff
@@ -0,0 +1,47 @@
...
+  public static DatasourceResponse map(@NonNull Datasource datasource) {
+    return new DatasourceResponse(
+        datasource.getName().getValue(),
+        datasource.getUrn().getValue(),
+        datasource.getConnectionUrl().getRawValue(),
+        datasource.getCreatedAt().toString());
+  }
+
+  public static List<DatasourceResponse> map(@NonNull List<Datasource> datasources) {
+    return datasources.isEmpty()
```
wslulciuc: We can simplify the return statement:

```java
  public static List<DatasourceResponse> map(@NonNull List<Datasource> datasources) {
    return unmodifiableList(
        datasource.stream().map(datasource -> map(datasource)).collect(toList()));
  }
```

We'll eventually want to cleanup `map(List)` in our other mappers as well

src/main/java/marquez/api/models/DatasourceRequest.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+
+@AllArgsConstructor(onConstructor = @__(@JsonCreator))
+@EqualsAndHashCode
+@ToString
+public class DatasourceRequest {
+  @Getter @NotNull private String connectionUrl;
+  @Getter @NotNull private String name;
```
wslulciuc: Minor: Though order of fields don't matter when deserializing the request payload, for consistency in how they are / will be defined in `openapi.yml` I'd order them as:

```java
DatasourceRequest {
  name
  connectionUrl
}
```

src/main/java/marquez/api/models/DatasourceResponse.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.ToString;
+
+@AllArgsConstructor(onConstructor = @__(@JsonCreator))
+@EqualsAndHashCode
+@ToString
+public class DatasourceResponse {
+  @Getter @NonNull private final String name;
+  @Getter @NonNull private final String urn;
+  @Getter @NonNull private final String connectionUrl;
+  @Getter @NonNull private final String createdAt;
```
wslulciuc: Minor: (similar comment in `DatasourceRequest `):

```java
DatasourceResponse {
  name
  createdAt
  urn
  connectionUrl
}
```

src/main/java/marquez/api/models/DatasourceRequest.java
wslulciuc: You'll want to use `@NotBlank` instead of `@NotNull`. That is, `@NotBlank`: _a constrained String is valid as long as it’s not null and the trimmed length is greater than zero_. Which is really what we want :)

src/main/java/marquez/api/resources/DatasourceResource.java
wslulciuc: Yep 👍

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -35,23 +35,23 @@
...
-  DataSourceDao createDataSourceDao();
+  DatasourceDao createDatasourceDao();
 
   @CreateSqlObject
   DbTableVersionDao createDbTableVersionDao();
 
   @SqlUpdate(
       "INSERT INTO datasets (guid, namespace_guid, datasource_uuid, urn, description) "
-          + "VALUES (:uuid, :namespaceUuid, :dataSourceUuid, :urn, :description)")
+          + "VALUES (:uuid, :namespaceUuid, :datasourceUuid, :urn, :description)")
```
wslulciuc: Did we also apply the column rename in our DB schema?

src/main/java/marquez/db/DatasourceDao.java
```diff
@@ -17,23 +17,28 @@
...
   @SqlQuery("SELECT * FROM datasources WHERE guid = :uuid")
-  Optional<DataSourceRow> findBy(UUID uuid);
+  Optional<DatasourceRow> findBy(@Bind("uuid") UUID uuid);
+
+  @SqlQuery("SELECT * FROM datasources WHERE urn = :urn")
+  Optional<DatasourceRow> findBy(@Bind("urn") DatasourceUrn urn);
 
   @SqlQuery("SELECT * FROM datasources LIMIT :limit OFFSET :offset")
-  List<DataSourceRow> findAll(Integer limit, Integer offset);
+  List<DatasourceRow> findAll(@Bind("limit") Integer limit, @Bind("offset") Integer offset);
```
wslulciuc: We no longer need to apply `@Bind` to our method params, which is why they have been removed in this class. See #367

src/main/java/marquez/service/DatasetService.java
```diff
@@ -47,7 +47,7 @@ public DatasetService(@NonNull final DatasetDao datasetDao) {
   public Dataset create(
       @NonNull NamespaceName namespaceName, @NonNull DbTableVersion dbTableVersion)
       throws MarquezServiceException {
-    final DataSourceRow dataSourceRow = DataSourceRowMapper.map(dbTableVersion);
+    final DatasourceRow dataSourceRow = DatasourceRowMapper.map(dbTableVersion);
```
wslulciuc: Minor: `dataSourceRow` => `datasourceRow`

src/main/java/marquez/service/DatasourceService.java
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      final Optional<DatasourceRow> datasourceRowIfFound = datasourceDao.findBy(datasourceUrn);
+      return datasourceRowIfFound.map(DatasourceMapper::map);
+    } catch (UnableToExecuteStatementException e) {
+      log.error(e.getMessage());
+      throw new MarquezServiceException();
+    }
+  }
+
+  public Optional<Datasource> get(@NonNull final UUID uuid) throws MarquezServiceException {
```
wslulciuc: I don't think we are supporting datasource lookup by `uuid`?

src/main/java/marquez/service/mappers/DatasourceMapper.java
```diff
@@ -0,0 +1,48 @@
...
+    final DatasourceUrn datasourceUrn = DatasourceUrn.from(connectionUrl, datasourceName);
+    return new Datasource(
+        datasourceName,
+        datasourceUrn,
+        ConnectionUrl.fromString(datasourceRow.getConnectionUrl()),
+        datasourceRow.getCreatedAt().get());
+  }
+
+  public static List<Datasource> map(@NonNull List<DatasourceRow> datasourceRows) {
+    return datasourceRows.isEmpty()
```
wslulciuc: Minor: See comment on `DatasourceResponseMapper.map(List)` on a cleaner approach.

src/main/java/marquez/api/resources/DatasourceResource.java
wslulciuc: You'll also need to annotate the request with [`@Valid`](https://docs.oracle.com/javaee/7/api/javax/validation/Valid.html)

src/main/java/marquez/db/DatasetDao.java
ashulmanWeWork: This doesn't affect the DB, where the field name is `datasource_uuid`:
`ALTER TABLE datasets ALTER COLUMN datasource_uuid SET NOT NULL;`

This is just the name of the field from `DatasetRow`, which is also in that same casing:
`@Getter @NonNull private final UUID datasourceUuid;`

src/main/resources/db/migration/V10__fix_datasources.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE datasources ADD COLUMN urn VARCHAR(128) NOT NULL;
+ALTER TABLE datasources ADD CONSTRAINT unique_name UNIQUE (name, urn)
```
wslulciuc: The `urn` itself is good enough for uniqueness. The datasource URN will have the name as part of the URN:

```
urn:{datasource_type}:{datasource_name}
```

src/main/java/marquez/common/models/DatasourceUrn.java
```diff
@@ -17,8 +17,10 @@
 import static marquez.common.Preconditions.checkNotBlank;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import lombok.EqualsAndHashCode;
 import lombok.NonNull;
 
+@EqualsAndHashCode(callSuper = true)
```
wslulciuc: We'll need to be careful on merges. Looks like `@EqualsAndHashCode` has been re-introduced. Mind comparing you `DatasourceUrn.java` with what's in master?

src/main/java/marquez/common/models/Urn.java
```diff
@@ -27,7 +27,8 @@
 @EqualsAndHashCode
 @ToString
 public abstract class Urn {
-  @Getter private final String value;
+
+  @Getter protected final String value;
```
wslulciuc: We'll need to be careful on merges. Looks like the scope of `value` has changed? Mind comparing you `Urn.java` with what's in master?

src/main/java/marquez/common/models/DatasourceUrn.java
```diff
@@ -33,8 +35,19 @@ public static DatasourceUrn from(@NonNull DatasourceType type, @NonNull Datasour
...
+  public static DatasourceUrn from(
+      @NonNull ConnectionUrl connectionUrl, @NonNull DatasourceName name) {
+    return from(connectionUrl.getDatasourceType(), name);
+  }
+
+  public static DatasourceUrn from(@NonNull String connectionUrl, @NonNull String name) {
+    return from(
+        ConnectionUrl.fromString(connectionUrl).getDatasourceType(),
+        DatasourceName.fromString(name));
+  }
```
wslulciuc: Do we need:

```
DatasourceUrn.from(ConnectionUrl, DatasourceName)
DatasourceUrn.from(String, String)
```

We only want to define `DatasourceUrn.from(DatasourceType, DatasourceName)`. Do the new builder methods add convenience when constructing datasource URNs?

src/test/java/marquez/common/models/DatasourceTypeTest.java
```diff
@@ -0,0 +1,42 @@
...
+    assertThat(myDatasourceType.toString()).isEqualTo(datasourceType);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testValueOf_invalidValue() {
+    final String datasourceType = "mysql_";
+    final DatasourceType myDatasourceType = DatasourceType.valueOf(datasourceType.toUpperCase());
+    assertThat(myDatasourceType.toString()).isEqualTo(datasourceType);
+  }
+}
```
wslulciuc: Is there an added benefit of testing `DatasourceType`. If we added helper methods for parsing the string value to an enum, then I'd introduce a test suite. The Java lang already provides sufficient testing for enums.

src/test/java/marquez/db/mappers/DatasourceRowMapperTest.java
```diff
@@ -47,12 +50,14 @@ public void testMap() throws SQLException {
     when(results.getTimestamp(Columns.CREATED_AT)).thenReturn(Timestamp.from(CREATED_AT));
     when(results.getString(Columns.NAME)).thenReturn(NAME);
     when(results.getString(Columns.CONNECTION_URL)).thenReturn(CONNECTION_URL);
+    when(results.getString(Columns.URN)).thenReturn(URN);
     final StatementContext context = mock(StatementContext.class);
 
-    final DataSourceRowMapper dataSourceRowMapper = new DataSourceRowMapper();
-    final DataSourceRow dataSourceRow = dataSourceRowMapper.map(results, context);
+    final DatasourceRowMapper dataSourceRowMapper = new DatasourceRowMapper();
```
wslulciuc: Minor: Standardizing on `Datasource` also requires we update variable names

src/test/java/marquez/common/models/DatasourceTypeTest.java
ashulmanWeWork: Good point. This doesn't test a whole lot more than just basic enum behavior. Will remove.

src/main/java/marquez/common/models/DatasourceUrn.java
ashulmanWeWork: They are pretty convenient and make the code much less clunky, especially for test code. The code to extract the type from the `connectionUrl` can get verbose and muddy up the code.

I think it's also nice because it takes in the two attributes that a user gives to create a Datasource, namely the connection url and the name.

src/main/java/marquez/common/models/Urn.java
ashulmanWeWork: This happened because we needed to use `value` in the subclass and modified it to be protected. Later, when I removed the methods that used it, I didn't adjust the scope back to being private.

src/main/java/marquez/common/models/DatasourceUrn.java
ashulmanWeWork: That's the only difference. I thought to keep it in during the merge b/c I thought that the equality tests (which had failed without it) might still need it. Turns out this is not the case. No other differences WRT master between this file, nor `DatasetUrn.java`.

src/main/java/marquez/common/models/DatasourceUrn.java
wslulciuc: Makes sense. We'll only want to define `DatasourceUrn.from(ConnectionUrl, DatasourceName)` and an provide an accompanying test for the new method.

src/main/java/marquez/service/DatasourceService.java
```diff
@@ -0,0 +1,74 @@
...
+  public Datasource create(@NonNull ConnectionUrl connectionUrl, @NonNull DatasourceName name)
+      throws MarquezServiceException {
+    final DatasourceRow datasourceRow = DatasourceRowMapper.map(connectionUrl, name);
+
+    try {
+      datasourceDao.insert(datasourceRow);
+      final DatasourceUrn datasourceUrn = DatasourceUrn.fromString(datasourceRow.getUrn());
+      final Optional<DatasourceRow> datasourceRowIfFound =
+          datasourceDao.findBy(datasourceUrn.getValue());
+      return datasourceRowIfFound.map(DatasourceMapper::map).orElseThrow(MarquezException::new);
```
sshah-wework: we throw and catch our own exception here?

sshah-wework: why not just throw the right type of exception initially?

ashulmanWeWork: makes sense - hadn't really considered replacing which exception was caught, but I think the right thing to do is to model it off of what is done in DatasetService. It will also more closely match the rest of the file as well.

src/main/java/marquez/api/mappers/DatasourceResponseMapper.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public static List<DatasourceResponse> map(@NonNull List<Datasource> datasources) {
+    return unmodifiableList(
+        datasources.stream().map(datasource -> map(datasource)).collect(toList()));
+  }
+
+  public static DatasourcesResponse toDatasourcesResponse(@NonNull List<Datasource> datasources) {
+    return new DatasourcesResponse(map(datasources));
+  }
+}
```
wslulciuc: We'll want to write a test for this mapper. See [`DatasetResponseMapperTest `](https://github.com/MarquezProject/marquez/blob/master/src/test/java/marquez/api/mappers/DatasetResponseMapperTest.java)

PR #385, opened 2019-03-06
Bound {job} to JobName

pbrahmbhatt3:
Fix375
Bound {job} in the path variable to JobName in JobResource.java file.

### DIFF
--- a/src/main/java/marquez/api/resources/JobResource.java
+++ b/src/main/java/marquez/api/resources/JobResource.java
@@ -102,13 +102,13 @@ public final class JobResource {
   @Produces(APPLICATION_JSON)
   @Timed
   public Response getJob(
-      @PathParam("namespace") final String namespace, @PathParam("job") JobName job)
+      @PathParam("namespace") final String namespace, @PathParam("job") JobName jobName)
       throws ResourceException {
     try {
       if (!namespaceService.exists(namespace)) {
         return Response.status(Response.Status.NOT_FOUND).entity("Namespace not found").build();
       }
-      final Optional<Job> returnedJob = jobService.getJob(namespace, job.getVal());
+      final Optional<Job> returnedJob = jobService.getJob(namespace,  jobName.getVal());
       if (returnedJob.isPresent()) {
         return Response.ok().entity(coreJobToApiJobMapper.map(returnedJob.get())).build();
       }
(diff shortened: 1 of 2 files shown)


pbrahmbhatt3:
Hi @wslulciuc, I am new to the pull request process. Can you please guide me through the process?

wslulciuc:
Sure thing. Have you looked over our contributing guide? [CONTRIBUTING.md](https://github.com/MarquezProject/marquez/blob/master/CONTRIBUTING.md)

pbrahmbhatt3:
For some reason I cannot run ./gradlew test. Also, once I have made the changes according to the document I have to write a test. Can you please give me an idea about what kind of test? If you can give a for instance it would be great.

wslulciuc:
> For some reason I cannot run ./gradlew test.

I'd make sure you're running `./gradlew test` at the project root. Mind providing the error you are seeing?

> Also, once I have made the changes according to the document I have to write a test. 

Writing a test depends on the change. For a new feature, or in the case of simply adding logic to existing code, there may not be proper code overage. So, it's important tests are included along with the PR.

In the case of your changes, we have existing tests. Meaning, once you update how we bind `{job}`, they'll be tests that fail and it's important to get those passing again as part of your PR.

pbrahmbhatt3:
@wslulciuc Thanks for the help. I have updated the code. Can you have a look again?

pbrahmbhatt3:
@wslulciuc Thanks for the help. I have solved the errors. Can you please review the code?

src/main/java/marquez/api/resources/JobResource.java
```diff
@@ -69,7 +70,7 @@ public JobResource(final NamespaceService namespaceService, final JobService job
   @Timed
   public Response create(
       @PathParam("namespace") final String namespace,
-      @PathParam("job") final String job,
+      @PathParam("job") JobName job,
```
wslulciuc: Minor: Let's also update the param name: `job` **=>** `jobName`

src/main/java/marquez/common/models/JobName.java
```diff
@@ -25,4 +25,8 @@
 @ToString
 public final class JobName {
   @Getter @NonNull private final String value;
+
+  public String getVal() {
```
wslulciuc: A getter method will be generated by lombok with [`@Getter`](https://projectlombok.org/features/GetterSetter).  So, we can remove `getVal()`. Please see [`JobNameTest`](https://github.com/MarquezProject/marquez/blob/master/src/test/java/marquez/common/models/JobNameTest.java) for example usage.

pbrahmbhatt3: Question: How can I access the @Getter in jobResource.java? 
marquezerror
I replaced all the job.getVal() with jobName(I know it doesn't make sense because I am trying to switch a class with a string) I get errors. @wslulciuc

pbrahmbhatt3:
@wslulciuc I have updated my code according to your comments. Please let me know if its good to go?

Issue #300, opened 2019-02-10
Add NamespaceMapper


wslulciuc:
Maps **class** [`NamespaceRequest`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/api/models/NamespaceRequest.java) to a [`Namespace`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/service/models/Namespace.java)

**NOTE:** For a great example mapper, see [`DatasetMapper`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/api/mappers/DatasetMapper.java)

wslulciuc:
Thanks for picking up this issue, @DavidLi119!

DavidLi119:
Moving my questions to here as it is more relevant
I have created a new file called NamespaceMapper.java and made the following changes after copying what was in NamespaceApiMapper.java.
```java
public **final** class NamespaceMapper extends Mapper<NamespaceResponse, Namespace> {
  private NamespaceMapper() {}

  public **static** Namespace map(NamespaceResponse namespace) {
    requireNonNull(namespace, "namespace must not be null");
    return new Namespace(
        null, namespace.getName().toLowerCase(), namespace.getOwner(), namespace.getDescription());
  }
```
 I tried the `./gradlew test` command in VSCode, but it says that it fails because it cannot build due to failing the task `:compileJava` despite me downloading a JDK. Are these changes what you wanted? And how do I test if it works?

wslulciuc:
@DavidLi119, you've made some amazing progress on this issue! Let me know if you need further guidance on writing a test for `NamespaceMapper`.

DavidLi119:
Apologies, I've been working on other stuff at the moment, I will get to it soon.

wslulciuc:
Fixed #488

Issue #363, opened 2019-03-01
Service models must not include a UUID field


ashulmanWeWork:
For consistency, a UUID field should be included on all object models. For example, `service.models.Job` (and most others) contains a UUID field, but `service.models.Dataset` does not.

wslulciuc:
The UUIDs set in our service models reference the primary key of a given row and should be removed. It's something I've been meaning to address. We now map individual rows to a **db** [row](https://github.com/MarquezProject/marquez/tree/master/src/main/java/marquez/db/models) instead of a service model.

So, it's not about consistency but rather code we never got around to cleaning up

wslulciuc:
I'll work on defining a set of clean up tasks for our service models

wslulciuc:
Fixed #627

PR #388, opened 2019-03-06
Add NamespaceMapper

DavidLi119:
issue#300
Created NamespaceMapper.java based on NamespaceApiMapper with changes from epic of issue#301. Reverted NamespaceApiMapper back to previous version due to accidentally changing it.


src/main/java/marquez/api/mappers/NamespaceMapper.java
```diff
@@ -0,0 +1,37 @@
...
+import static java.util.Objects.requireNonNull;
+
+import marquez.api.models.NamespaceRequest;
+import marquez.api.models.NamespaceResponse;
+import marquez.service.models.Namespace;
+
+public final class NamespaceMapper extends Mapper<NamespaceResponse, Namespace> {
+  private NamespaceMapper() {}
+
+  public static Namespace map(NamespaceResponse namespace) {
```
wslulciuc: This class should only convert a `NamespaceRequest` to a `Namespace`. So, we'll want to remove this method.

src/main/java/marquez/api/mappers/NamespaceMapper.java
```diff
@@ -0,0 +1,37 @@
...
+
+package marquez.api.mappers;
+
+import static java.util.Objects.requireNonNull;
+
+import marquez.api.models.NamespaceRequest;
+import marquez.api.models.NamespaceResponse;
+import marquez.service.models.Namespace;
+
+public final class NamespaceMapper extends Mapper<NamespaceResponse, Namespace> {
```
wslulciuc: **Mapper** is a legacy class that will be deprecated soon. Here's the related issue: #389. Meaning, we define the class without extending mapper class

src/main/java/marquez/api/mappers/NamespaceMapper.java
```diff
@@ -0,0 +1,37 @@
...
+  private NamespaceMapper() {}
+
+  public static Namespace map(NamespaceResponse namespace) {
+    requireNonNull(namespace, "namespace must not be null");
+    return new Namespace(
+        null, namespace.getName().toLowerCase(), namespace.getOwner(), namespace.getDescription());
+  }
+
+  public Namespace of(String namespaceName, NamespaceRequest request) {
```
wslulciuc: Nice! So, since you copied over the code from `NamespaceApiMapper`. Let's make sure we update and name all converter methods as `map()`.

Here's what I'd recommend for your `map()` method:
-  Use the type [`NamespaceName`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/common/models/NamespaceName.java) instead of a string
- Annotate all method args with [`@NonNull`](https://projectlombok.org/features/NonNull)

DavidLi119: Ok, so gathering what you said and referring to `DbTableVersionMapper.java`, I made the following changes:
1. Removed the `extends Mapper<>`
2. Added `@NonNull` before all method args
3. changed `public Namespace of()` to `public static Namespace map()`
4. Commented out the first method

```java
public final class NamespaceMapper {
  private NamespaceMapper() {}

/*  public static Namespace map(NamespaceResponse namespace) {
    requireNonNull(namespace, "namespace must not be null");
    return new Namespace(
        null, namespace.getName().toLowerCase(), namespace.getOwner(), namespace.getDescription());
  }*/

  public static Namespace map(@NonNull NamespaceName namespaceName, @NonNull NamespaceRequest request) {
    return map(
        new NamespaceResponse(
            namespaceName, null, request.getOwner(), request.getDescription().orElse(null)));
  }
}
```

wslulciuc: You're making great progress, @DavidLi119!  I noticed the CI build is failing. Are you able to run the following cmd successfully locally on your machine?

```
$ ./gradlew compileJava
```

src/main/java/marquez/api/mappers/NamespaceMapper.java
```diff
@@ -0,0 +1,37 @@
...
+import static java.util.Objects.requireNonNull;
+
+import marquez.api.models.NamespaceRequest;
+import marquez.api.models.NamespaceResponse;
+import marquez.service.models.Namespace;
+
+public final class NamespaceMapper {
+  private NamespaceMapper() {}
+
+/*  public static Namespace map(NamespaceResponse namespace) {
```
wslulciuc: Let's remove this method.

src/main/java/marquez/api/mappers/NamespaceMapper.java
```diff
@@ -0,0 +1,37 @@
...
+    requireNonNull(namespace, "namespace must not be null");
+    return new Namespace(
+        null, namespace.getName().toLowerCase(), namespace.getOwner(), namespace.getDescription());
+  }*/
+
+  public static Namespace map(@NonNull NamespaceName namespaceName, @NonNull NamespaceRequest request) {
+    return map(
+        new NamespaceResponse(
```
wslulciuc: The return type for this method is `Namespace`, but you're instantiating a `NamespaceResponse`. Mind double checking the logic of the method to ensure we are returning the correct class type?

DavidLi119:
So I removed the constructor like you said, though the reason I put it there was because `DbTableVersionMapper.java` had it. I also made it more similar to `DbTableVersionMapper.java` in the sense that it returns a `new Namespace`, however comparing to the `Namespace` Constructor I have a few questions below:
```java
public final class NamespaceMapper {
   public static Namespace map(@NonNull NamespaceName namespaceName, @NonNull 
      NamespaceRequest request) {
         return new Namespace(
            //UUID?
            namespaceName,
            request.getOwner(),
            request.getDescription().orElse(null)
            );
  }
}
```
Im looking at this `Namespace` constructor:
```java
public Namespace(UUID guid, String name, String ownerName, String description) {
    this.guid = guid;
    this.name = name;
    this.ownerName = ownerName;
    this.description = description;
    this.createdAt = null;
  }
```

DavidLi119:
So I installed gradle 5.2.1, however the command ./gradlew test still fails at TASK :compileJava

wslulciuc:
Hey, @DavidLi119. Thanks for making the changes and sorry for the delayed review.

> So I removed the constructor like you said, though the reason I put it there was because DbTableVersionMapper.java had it.

You were right to add a **private** constructor. Mappers are more like utility classes, so we avoid allowing class instantiation (similar to **class** [`Math`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Math.html) in Java). In my earlier comment to "remove _this_ method", I was referring to `Namespace map(NamespaceResponse)` that was commented out. Mind adding the **private** constructor back?

> Im looking at this Namespace constructor:
>
> public Namespace(UUID guid, String name, String ownerName, String description) {
> this.guid = guid;
> this.name = name;
> this.ownerName = ownerName;
> this.description = description;
> this.createdAt = null;
>}

We are working on updating our service models to _not_ contain `UUID`s (see #363). We use `UUID`s as the primary key for a row in our DB and don't want to expose them to the callers of our API.

In  **class** [`Namespace`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/service/models/Namespace.java), I recommend defining a new constructor:

```java
public Namespace(String name, String ownerName, String description) {
...
}
```

That way you can avoid generating a UUID in the mapper.

wslulciuc:
> So I installed gradle 5.2.1, however the command ./gradlew test still fails at TASK :compileJava

I'm sure you do, but wanted to ask if you have the prerequisites installed to run gradle? https://gradle.org/install/#prerequisites

DavidLi119:
Sorry for the late reply. So it turns out I have not updated my Java to the latest version for a very long time, thus .\gradlew test not working is obvious (solved).
Now I have a few questions I hope you can help me resolve:
1)  I added the new constructor into the `Namespace` class, however when I test it, it fails the test saying that the variable `guid` has not been initialized, however this is one of the variables you did not want to include. How should I proceed?
Error:
`> Task :compileJava
C:\Users\darky\Documents\GitHub\marquez\src\main\java\marquez\service\models\Namespace.java:53: error: variable guid might not have been initialized
  }
  ^`
2) With this constructor, you wanted the name to be type `String`, however in the `NamespaceMapper.java`, you wanted the name to be type `NamespaceName`. On another note, I had to add an import line so the test could actually find type `NamespaceName`. How do I proceed?
3) A few minor questions about the code itself:
```java
package marquez.api.mappers;

import static java.util.Objects.requireNonNull; \\necessary?
import lombok.NonNull;                       \\added otherwise `@NonNull` couldn't be found
import marquez.api.models.NamespaceRequest;
import marquez.api.models.NamespaceResponse;
import marquez.service.models.Namespace;
import marquez.common.models.NamespaceName; 
\\^added so type NamespaceName could be referenced

public final class NamespaceMapper {
  private NamespaceMapper() {}

  public static Namespace map(@NonNull NamespaceName namespaceName, @NonNull NamespaceRequest request) {
    return new Namespace(
            namespaceName,
            request.getOwner(),
            request.getDescription().orElse(null)
            );
  }
}
```

.gitignore
```diff
@@ -16,3 +16,4 @@ build/
 
 # Marquez configuration
 config.yml
+/.metadata/
```
wslulciuc: You'll want to define a [`.gitignore_global`](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) to ignore files specific to your dev environment. The `.gitignore` in this repo serves as excluding files specific to the project.

src/main/java/marquez/api/mappers/NamespaceMapper.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package marquez.api.mappers;
+
+import static java.util.Objects.requireNonNull;
```
wslulciuc: The `requireNonNull()` import is no longer required and replaced by  `@NonNull`

src/main/java/marquez/service/models/Namespace.java
```diff
@@ -17,6 +17,7 @@
 import java.sql.Timestamp;
 import java.util.UUID;
 import lombok.Data;
+import marquez.common.models.NamespaceName;
```
wslulciuc: I don't think this import is being used?

src/main/java/marquez/service/models/Namespace.java
```diff
@@ -43,4 +44,11 @@ public Namespace(
     this.ownerName = ownerName;
     this.description = description;
   }
+
+  public Namespace(
+      String name, String ownerName, String description) {
+    this.name = name;
```
wslulciuc: You'll also want to set the `guid` and `createdAt` to `null` in the constructor:

```java
this.guid = null;
this.name = name;
this.createdAt = null;
this.ownerName = ownerName;
this.description = description;
```

This isn't ideal, but we are working towards cleaning up our service models

wslulciuc:
> With this constructor, you wanted the name to be type String, however in the NamespaceMapper.java, you wanted the name to be type NamespaceName. On another note, I had to add an import line so the test could actually find type NamespaceName. How do I proceed?

Have the types be different is fine. The constructor for `Namespace()` takes a namespace as a string, which is what we want.

For your other questions, I left comments in your PR. They should help and set you on the right path to get this PR merged. Let me know if you need more clarity.

Thanks again for helping out, @DavidLi119!

DavidLi119:
@wslulciuc So apparently `./gradlew test` has a problem with type NamespaceName. the error is:
` Task :compileJava
..\marquez\src\main\java\marquez\api\mappers\NamespaceMapper.java:28: error: incompatible types: NamespaceName cannot be converted to String
            namespaceName,
            ^`
So is there a problem with my `NamespaceMapper.java` or the model `NamespaceName.java`?

wslulciuc:
> So is there a problem with my NamespaceMapper.java or the model NamespaceName.java?

You'll want to use `NamespaceName,getValue()` to return the string representation of the namespace name. That should resolve your compilation error.

DavidLi119:
@wslulciuc The compilation error was fixed and all subsequent problems have been fixed. Only requiring codecov checks to be passed. Thanks for working with me through this issue. Been a pleasure.

wslulciuc:
Great work, @DavidLi119! Since this PR is introducing new code, you will also need to write a test to meet our coverage requirements (see [CONTRIBUTING.md](https://github.com/MarquezProject/marquez/blob/master/CONTRIBUTING.md)). As an example, I would reference [`DbTableVersionMapperTest`](https://github.com/MarquezProject/marquez/blob/master/src/test/java/marquez/api/mappers/DbTableVersionMapperTest.java)

DavidLi119:
@wslulciuc I can't seem to find the test coverage requirements on the CONTRIBUTING.md, can you link them to me?

wslulciuc:
We block PRs if codecoverage drops (i.e. `codecov/patch`, `codecov/project`). It's really in place to minimize the introduction of bugs. See step `5` in ["Submitting a pull request"](https://github.com/MarquezProject/marquez/blob/master/CONTRIBUTING.md#submitting-a-pull-request)

DavidLi119:
step 5 has no links attached, all it says is "write a test for your change"

wslulciuc:
>  I do not have any experience writing tests.

For a background on writing Java tests, see: https://junit.org/junit5. And for an example on how to write a test for **NamespaceMapper**, see [`DbTableVersionMapperTest`](https://github.com/MarquezProject/marquez/blob/master/src/test/java/marquez/api/mappers/DbTableVersionMapperTest.java)

wslulciuc:
Closes #300

wslulciuc:
@DavidLi119: Was the links provided above helpful? Let me know If I can assist in anyway.

wslulciuc:
@DavidLi119, We'll have to close this PR as it has been a while since we've seen activity. Please feel free to re-open.

PR #392, opened 2019-03-06
disbaled javadoc check

pbrahmbhatt3:
Disbaled the checks for javadoc from checklist.xml by assigning "ignore" to value property.


pbrahmbhatt3:
@wslulciuc I am having difficulty figuring out why this fails. Any suggestions

checkstyle.xml
```diff
@@ -231,21 +231,21 @@
              value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
         </module>
         <module name="JavadocMethod">
-            <property name="scope" value="public"/>
```
wslulciuc: I believe we only need to change the scope to ignore JavaDoc on methods. Can you try changing the scope from 
`public` to `nothing`?
```
 <property name="scope" value="nothing"/>
```

pbrahmbhatt3: sure. Thanks. :D

wslulciuc: I'd be easier (I think) if you reverted the values you introduced and only modify the scope?

checkstyle.xml
```diff
@@ -6,12 +6,9 @@
 <!--
     Checkstyle configuration that checks the Google coding conventions from Google Java Style
     that can be found at https://google.github.io/styleguide/javaguide.html.
-
```
wslulciuc: Minor: The newlines here help with readability. Mind adding them back?

pbrahmbhatt3:
Thanks for your help! You are the best. You have been really helpful. @wslulciuc

PR #394, opened 2019-03-08
[Datasource Service - Part 3]: Incorporate urns

ashulmanWeWork:


src/main/java/marquez/api/mappers/DatasourceResponseMapper.java
```diff
@@ -41,7 +42,7 @@ public static DatasourceResponse map(@NonNull Datasource datasource) {
   }
 
   public static DatasourcesResponse toDatasourcesResponse(@NonNull List<Datasource> datasources) {
-    List<DatasourceResponse> datasourceResponses = map(datasources);
+    final List<DatasourceResponse> datasourceResponses = map(datasources);
     return new DatasourcesResponse(datasourceResponses);
```
wslulciuc: Minor: It's fine to inline `map()`

```java
return new DatasourcesResponse(map(datasources));
```

src/main/java/marquez/api/resources/DatasourceResource.java
```diff
@@ -91,10 +89,18 @@ public Response get(@PathParam("datasourceId") final DatasourceId datasourceId)
   @Produces(APPLICATION_JSON)
   public Response create(@NonNull final DatasourceRequest datasourceRequest)
       throws MarquezServiceException {
+
+    final ConnectionUrl connectionUrl;
+    try {
```
wslulciuc: The `try/catch` block can be removed when we add an exception mapper to handle thrown `IllegalArgumentException`s #403

src/main/java/marquez/common/models/DatasourceType.java
```diff
@@ -14,15 +14,30 @@
...
+  }
+
+  public static final DatasourceType fromString(@NonNull String value) {
+    for (DatasourceType type : DatasourceType.values()) {
+      if (type.value.equals(value)) {
+        return type;
+      }
+    }
+    return null;
+  }
```
wslulciuc: Let's define this enum as:

```java
public enum DatasourceType {
  REDSHIFT,
  MYSQL,
  POSTGRESQL;

  @Override
  public String toString() {
    return name().toLowerCase();
  }
}
```

src/main/java/marquez/db/DatasourceDao.java
```diff
@@ -28,13 +28,16 @@
...
-  void insert(@BindBean DatasourceRow dataSourceRow);
+      "INSERT INTO datasources (guid, urn, name, connection_url) "
+          + "VALUES (:uuid, :urn, :name, :connectionUrl)")
+  void insert(@BindBean DatasourceRow datasourceRow);
 
   @SqlQuery("SELECT * FROM datasources WHERE guid = :uuid")
   Optional<DatasourceRow> findBy(@Bind("uuid") UUID uuid);
 
+  @SqlQuery("SELECT * FROM datasources WHERE urn = :urn")
+  Optional<DatasourceRow> findBy(@Bind("urn") String urn);
```
wslulciuc: Use the `DatasourceUrn` type instead of a string

src/main/java/marquez/db/DatasourceDao.java
```diff
@@ -28,13 +28,16 @@
 @RegisterRowMapper(DatasourceRowMapper.class)
 public interface DatasourceDao {
   @SqlUpdate(
-      "INSERT INTO datasources (guid, name, connection_url) "
-          + "VALUES (:uuid, :name, :connectionUrl)")
-  void insert(@BindBean DatasourceRow dataSourceRow);
+      "INSERT INTO datasources (guid, urn, name, connection_url) "
+          + "VALUES (:uuid, :urn, :name, :connectionUrl)")
```
wslulciuc: To avoid the _insert-then-get_ pattern in `DatasourceService`, use:

```java
@SqlQuery("INSERT INTO datasources (guid, urn, name, connection_url) "
	          + "VALUES (:uuid, :urn, :name, :connectionUrl"
                  + "RETURNING *")
Optional<DatasourceRow> insertAndGet(@BindBean DatasourceRow datasourceRow)
```

I'd double check my query.

see: https://www.postgresql.org/docs/9.6/dml-returning.html

src/main/java/marquez/common/models/DatasourceUrn.java
```diff
@@ -19,7 +19,22 @@ public static DatasourceUrn from(@NonNull String type, @NonNull String name) {
...
+
+  public DatasourceType getDatasourceType() {
+    final String datasourceType = value.split(":")[2];
+    return DatasourceType.fromString(datasourceType);
+  }
+
+  public DatasourceName getDatasourceName() {
+    final String datasourceName = value.split(":")[3];
+    return DatasourceName.fromString(datasourceName);
+  }
```
wslulciuc: I don't think we want URNs to be bidirectional (build <-> parse components). That is, we should only provide ways to create a URN, using it as an identifier for lookups when interfacing with our DB. @sshah-wework let me know if I'm thinking about this correctly.

src/main/java/marquez/service/mappers/DatasourceRowMapper.java
```diff
@@ -18,24 +18,32 @@
...
   public static DatasourceRow map(@NonNull DbTableVersion dbTableVersion) {
+    final DatasourceName datasourceName =
+        DatasourceName.fromString(dbTableVersion.getConnectionUrl().getDatasourceType().name());
+    final DatasourceUrn datasourceUrn =
+        DatasourceUrn.from(dbTableVersion.getConnectionUrl(), datasourceName);
     return DatasourceRow.builder()
         .uuid(UUID.randomUUID())
-        .name(dbTableVersion.getConnectionUrl().getDatasource().getValue())
+        .urn(datasourceUrn.getValue())
+        .name(dbTableVersion.getConnectionUrl().getDatasourceType().name())
```
wslulciuc: Isn't the `urn` from a datasource different than it's `name`?

src/main/java/marquez/common/models/DatasourceUrn.java
sshah-wework: Correct, it's just an ID. We should never parse it because the details are hidden from the rest of the application intentionally.

src/main/java/marquez/api/resources/DatasourceResource.java
ashulmanWeWork: Although it may introduce changes originally not intended for the review, I'm happy to add it here with this PR.
Another option might be to open a new stacked PR atop #397.
LMK :)

src/main/java/marquez/db/DatasourceDao.java
ashulmanWeWork: This is a really cool idea that would simplify our code, but inserts as part of the `SQLQuery` annotation seem to not accept this type of output.
When I tried adding the `RETURNING *` clause, I received this from my dao test: 
```
org.jdbi.v3.sqlobject.UnableToCreateSqlObjectException: DatasourceDao.insert method is annotated with @SqlUpdate so should return void, boolean, or Number but is returning: java.util.Optional<marquez.db.models.DatasourceRow>
```

More info: https://github.com/jdbi/jdbi/blob/master/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlUpdateHandler.java#L35-L69

src/main/java/marquez/service/mappers/DatasourceRowMapper.java
ashulmanWeWork: I think what happened is that in the past, the `datasource` field was used both as the type as well as naming. I originally did not write the above method and I tried to have it deviate as little as possible, so this was necessary to keep the original intent as close as possible (since tests or other behavior may have been relying on it).

As part of my next checkin, I've modified it to return the name and adjusted a test to reflect that.

src/main/java/marquez/common/models/DatasourceUrn.java
ashulmanWeWork: Makes sense - removing 👍

src/main/java/marquez/db/DatasourceDao.java
ashulmanWeWork: 👍

src/main/java/marquez/api/resources/DatasourceResource.java
ashulmanWeWork: Will handle this at another time.

src/main/java/marquez/service/DatasourceService.java
```diff
@@ -76,7 +76,7 @@ public Datasource create(@NonNull ConnectionUrl connectionUrl, @NonNull Datasour
     }
   }
 
-  public List<Datasource> getAll(@NonNull Integer limit, @NonNull Integer offset) {
+  public List<Datasource> list(@NonNull Integer limit, @NonNull Integer offset) {
```
wslulciuc: Don't we want this to be `DatasourceService.getAll()`?

ashulmanWeWork: Yes, sorry, misunderstood a comment on a different PR. Addressed that, and fixed this mis-naming.

Issue #314, opened 2019-02-13
Add DatasetDao tests


wslulciuc:
Our test coverage for [`DatasetDao`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/db/DatasetDao.java) is currently `< 25%`, let's add some tests to get it closer to `75%`

codecov link: https://codecov.io/gh/MarquezProject/marquez/src/master/src/main/java/marquez/db/DatasetDao.java

hjpatel16:
Hey, I would like to work on it

wslulciuc:
Hey @hjpatel16. Great to hear you'd like to help out! I'm actually fixing this issue in PR #395 (sorry, should've assigned this to myself sooner). Once merged into `master`, it will serve as an example on how to write tests for our DAO classes.

Is there an issue tagged with ["good first issue"](https://github.com/MarquezProject/marquez/labels/good%20first%20issue) you'd like to pick up?

hjpatel16:
Actually, I was looking more than "good first issue" where I would be able to code more and contribute more. Let me know if you know any specific issue which will be perfect for me

wslulciuc:
@hjpatel16, sure! How does adding a factory method to our response models sound? Full details can be found here #309. I know the issues related to the epic are labeled as "good first issue", but these changes are non-trivial and would require writing tests, with new methods used right away in our [resource](https://github.com/MarquezProject/marquez/tree/master/src/main/java/marquez/api/resources) classes.

Let me know if that's something you're interested in.

NOTE: You would need a separate PR per issue in the epic

hjpatel16:
Yes, I would like to work on that epic. Thanks @wslulciuc, much appreciated

wslulciuc:
Fixed #395

Issue #306, opened 2019-02-10
Add DatasetResponse.toDatasetsResponse()


wslulciuc:
Returns an instance of [`DatasetsResponse`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/api/models/DatasetsResponse.java)

wslulciuc:
Fixed #395

PR #395, opened 2019-03-08
Add dataset creation and retrieval endpoints

wslulciuc:
This PR introduces the creation and retrieval of datasets #314 #306

##### `POST` `/namespaces/{namespace}/datasets`

Creates a dataset object, associating the newly created dataset with the provided namespace.

##### `GET` `/namespaces/{namespace}/datasets/{urn}`

Retrieves the dataset object with the given URN under the provided namespace. 

**NOTE:** There isn't an epic associated with these changes (sadly), but the plan is to create one and will make sure it's open before merging in this feature.


src/main/java/marquez/api/resources/DatasetResource.java
```diff
@@ -86,7 +86,9 @@ public Response create(@Valid DatasetRequest request) throws MarquezServiceExcep
   public Response create(
       @PathParam("namespace") NamespaceName namespaceName, @Valid DatasetRequest request)
       throws MarquezServiceException {
-    checkNamespaceExists(namespaceName);
+    if (!namespaceService.exists(namespaceName.getValue())) {
+      throw new NamespaceNotFoundException(namespaceName);
+    }
```
ashulmanWeWork: Why are we inlining this?

sshah-wework:
I will review today

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -40,40 +40,45 @@
...
+
   @SqlUpdate(
-      "UPDATE datasets SET updated_at = :updatedAt, current_version_uuid = :currentVersion "
-          + "WHERE guid = :uuid")
+      "UPDATE datasets SET updated_at = :updatedAt, current_version_uuid = :currentVersionUuid "
+          + "WHERE uuid = :uuid")
   void updateCurrentVersion(UUID uuid, Instant updatedAt, UUID currentVersion);
 
-  @SqlQuery("SELECT * FROM datasets WHERE uuid = :uuid")
+  @SqlQuery("SELECT * FROM datasets WHERE guid = :uuid")
```
ashulmanWeWork: Is the field on datasets `uuid` or `guid`?

wslulciuc: `guid`! I'll also add a test

src/main/java/marquez/db/models/DatasetRow.java
```diff
@@ -15,30 +15,20 @@
...
-import lombok.Getter;
-import lombok.NonNull;
-import lombok.RequiredArgsConstructor;
-import lombok.ToString;
+import lombok.Data;
 
-@RequiredArgsConstructor
-@EqualsAndHashCode
-@ToString
+@Data
```
ashulmanWeWork: Should we use `@Value` instead. It's like `@Data`, but I think it also gives us immutability built in.

src/main/java/marquez/db/Columns.java
```diff
@@ -86,9 +87,9 @@ public static UUID uuidOrThrow(ResultSet results, String column) throws SQLExcep
   }
 
   public static Instant timestampOrNull(ResultSet results, String column) throws SQLException {
-    if (results.getObject(column) == null) {
-      return null;
-    }
+      if (results.getObject(column) == null) {
+        return null;
+      }
```
ashulmanWeWork: nit: indentation seems off?

src/test/java/marquez/api/mappers/DatasetMapperTest.java
```diff
@@ -0,0 +1,63 @@
...
+    final Dataset dataset = DatasetMapper.map(REQUEST_NO_DESCRIPTION);
+    assertThat(dataset).isNotNull();
+    assertThat(dataset.getName()).isEqualTo(DATASET_NAME);
+    assertThat(dataset.getDescription()).isEqualTo(null);
+  }
+
+  @Test
+  public void testMap_throwsException_onNullRequest() {
+    final DatasetRequest nullRequest = null;
+    assertThatNullPointerException().isThrownBy(() -> DatasetMapper.map(nullRequest));
```
ashulmanWeWork: nice!

src/main/java/marquez/db/Columns.java
wslulciuc: It is, applied formatting.

src/test/java/marquez/api/resources/DatasetResourceTest.java
```diff
@@ -15,77 +15,173 @@
...
+      DatasourceUrn.from(CONNECTION_URL, DATASOURCE_NAME);
+  private static final DatasetName DATASET_NAME = newDatasetName();
+  private static final DatasetUrn DATASET_URN = DatasetUrn.from(DATASOURCE_NAME, DATASET_NAME);
+  private static final Description DESCRIPTION = newDescription();
+  private static final Dataset DATASET = newDatasetWith(DATASET_NAME, DATASET_URN, DESCRIPTION);
+  private static final DatasetRequest DATASET_REQUEST =
+      new DatasetRequest(DATASET_NAME, DATASOURCE_URN, DESCRIPTION);
+  private static final Dataset NEW_DATASET = DatasetMapper.map(DATASET_REQUEST);
+
+  private NamespaceService namespaceService;
```
ashulmanWeWork: can you make these static?

src/test/java/marquez/api/resources/DatasetResourceTest.java
```diff
@@ -15,77 +15,173 @@
...
+  private DatasetService datasetService;
+  private DatasetResource datasetResource;
+
+  @Before
+  public void setUp() {
+    namespaceService = mock(NamespaceService.class);
```
ashulmanWeWork: You only really want to set these up once per class, not once per method. Please use `@BeforeClass`.

src/main/java/marquez/db/models/DatasetRow.java
wslulciuc: We'll also need setters and `@Data` generates mutators for us while `@Value` only provides `@AllArgsConstructor` to initialize fields.

src/test/java/marquez/api/resources/DatasetResourceTest.java
wslulciuc: I normally use `@BeforeClass` for more expensive setup steps we only want to run once per test suite. With `@Before`, we can reinstantiate our mocks per tests. What are the benefits of using `@BeforeClass` instead?

src/test/java/marquez/service/NamespaceServiceTest.java
```diff
@@ -1,132 +1,107 @@
...
+  //   when(namespaceDao.findAll()).thenReturn(new ArrayList<Namespace>());
+  //   List<Namespace> namespacesFound = namespaceService.getAll();
+  //   assertEquals(0, namespacesFound.size());
+  // }
+  //
+  // @Test(expected = MarquezServiceException.class)
+  // public void testListNamespaces_findAllException() throws MarquezServiceException {
+  //   doThrow(UnableToExecuteStatementException.class).when(namespaceDao).findAll();
+  //   namespaceService.getAll();
+  // }
```
wslulciuc: FYI: Re-enabling the `NamespaceService` test suite :)

src/test/java/marquez/service/DatasetServiceTest.java
```diff
@@ -1,3 +1,157 @@
 package marquez.service;
```
ashulmanWeWork: Please add license info.

src/test/java/marquez/service/NamespaceServiceTest.java
```diff
@@ -1,132 +1,107 @@
...
-    List<Namespace> namespacesFound = namespaceService.getAll();
-    assertEquals(0, namespacesFound.size());
-  }
-
-  @Test(expected = MarquezServiceException.class)
-  public void testListNamespaces_findAllException() throws MarquezServiceException {
-    doThrow(UnableToExecuteStatementException.class).when(namespaceDao).findAll();
-    namespaceService.getAll();
-  }
+  // private final NamespaceName NAMESPACE_NAME = NamespaceName.fromString("test");
```
ashulmanWeWork: Not sure we should be checking in commented-out tests. If you want to ignore some tests, please use the JUnit annotations for doing so.

wslulciuc: Tests we're commented out locally as WIP

src/test/java/marquez/service/DatasetServiceTest.java
wslulciuc: We can use `spotless` to help us catch missing license headers. Something worth looking into.

src/test/java/marquez/service/NamespaceServiceTest.java
ashulmanWeWork: nice :)

src/main/java/marquez/service/DatasetService.java
```diff
@@ -63,12 +110,20 @@ public Dataset create(
     }
   }
 
-  public Optional<Dataset> get(@NonNull DatasetUrn urn) throws MarquezServiceException {
+  public boolean exists(@NonNull DatasetUrn datasetUrn) throws MarquezServiceException {
```
ashulmanWeWork: nit: why not just call it `urn`?

src/test/java/marquez/api/resources/DatasetResourceTest.java
ashulmanWeWork: The benefit is you can build something once and re-use (by clearing mocks before or after every test), instead of rebuilding it every time. I agree that it's not computationally intensive to re-instantiate a mock, but I think it's a good practice to do things only when necessary, since that thinking carries over to other test fixtures that could be more expensive.

src/main/java/marquez/service/DatasetService.java
wslulciuc: Originally, as I was defining DatasetService.create()`, I had explicitly listed all params instead of passing in a `Dataset` object. So, `urn` became `datasetUrn`, but makes sense now to shorten it. Good catch.

src/test/java/marquez/api/resources/DatasetResourceTest.java
ashulmanWeWork: The benefit is you can build something once and re-use (by clearing mocks before or after every test), instead of rebuilding it every time. I agree that it's not computationally intensive to re-instantiate a mock, but I think it's a good practice to do things only when necessary, since that thinking carries over to other test fixtures that could be more expensive.

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -14,72 +14,206 @@
...
+    datasetDao = jdbi.onDemand(DatasetDao.class);
   }
 
-  @Before
-  public void setupTest() {
-    namespace = Generator.genNamespace();
-    namespaceDAO.insert(namespace);
+  @Test
+  public void testInsert() {
+    int rowsBefore = datasetDao.count();
```
ashulmanWeWork: nit: this can be final

ashulmanWeWork: nit: this can be final

ashulmanWeWork: nit: this can be final

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -14,72 +14,206 @@
...
+    final DatasourceRow datasourceRow = newDatasourceRow();
+    namespaceDao.insert(namespaceRow);
+    datasourceDao.insert(datasourceRow);
+
+    final DatasetUrn datasetUrn = newDatasetUrn();
+    final DatasetRow newDatasetRow =
+        newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid(), datasetUrn);
+    datasetDao.insertAndGet(newDatasetRow);
+
+    boolean exists = datasetDao.exists(datasetUrn);
```
ashulmanWeWork: nit: this can be final too

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: nit: this can be final

src/test/java/marquez/db/JobRunDaoTest.java
```diff
@@ -82,6 +80,9 @@ public static void setup() throws MarquezServiceException {
     CREATED_JOB_RUN_UUID = createdJob.getNamespaceGuid();
   }
 
+  @BeforeClass
+  public static void setup() throws MarquezServiceException {}
```
ashulmanWeWork: Why introduce an empty class-level setup method?

src/test/java/marquez/db/NamespaceDaoTest.java
```diff
@@ -14,91 +14,97 @@
...
   }
 
-  private void assertFieldsMatchExceptTS(Namespace ns1, Namespace ns2) {
-    assertEquals(ns1.getGuid(), ns2.getGuid());
-    assertEquals(ns1.getName(), ns2.getName());
-    assertEquals(ns1.getDescription(), ns2.getDescription());
-    assertEquals(ns1.getOwnerName(), ns2.getOwnerName());
+  @Test
+  public void testInsert() {
+    int rowsBefore = namespaceDao.count();
```
ashulmanWeWork: nit: these ints can be final

src/test/java/marquez/db/JobRunDaoTest.java
wslulciuc: This method shouldn't have been introduced. Was something I was experimenting with the get a test to pass.

sshah-wework:
Great write up in the description!!

src/main/java/marquez/api/models/DatasetRequest.java
```diff
@@ -0,0 +1,26 @@
...
+import marquez.common.models.DatasourceUrn;
+import marquez.common.models.Description;
+
+@AllArgsConstructor(onConstructor = @__(@JsonCreator))
+@EqualsAndHashCode
+@ToString
+public final class DatasetRequest {
+  @Getter @NotNull private final DatasetName name;
+  @Getter @NotNull private final DatasourceUrn datasourceUrn;
+  @Nullable private final Description description;
```
sshah-wework: `createdAt`?

sshah-wework: woops, this is the request

sshah-wework: nvm :)

src/test/java/marquez/api/resources/DatasetResourceTest.java
sshah-wework: It's safer to just use `@Before` and get a fresh one. `@BeforeClass` is better for suite-level setup / teardown or avoiding something too expensive to do for every test method. In this case, instantiating a mock `Namespace` is super cheap, so no harm doing it in `@Before`, and no need to worry about cleaning things up manually.

src/test/java/marquez/api/resources/DatasetResourceTest.java
```diff
@@ -15,77 +15,192 @@
...
-  public void testListDatasetsNamespaceDoesNotExist() throws MarquezServiceException {
-    when(mockNamespaceService.exists(NAMESPACE_NAME)).thenReturn(false);
+  @Test
+  public void testList_throwsException_onNamespaceDoesNotExist() throws MarquezServiceException {
+    when(namespaceService.exists(NAMESPACE_NAME)).thenReturn(false);
+    assertThatExceptionOfType(NamespaceNotFoundException.class)
+        .isThrownBy(() -> datasetResource.list(NAMESPACE_NAME, LIMIT, OFFSET));
 
-    datasetResource.list(NAMESPACE_NAME, LIMIT, OFFSET);
+    verify(datasetService, never()).getAll(NAMESPACE_NAME, LIMIT, OFFSET);
```
sshah-wework: can we try the empty list case?

src/test/java/marquez/api/resources/DatasetResourceTest.java
```diff
@@ -15,77 +15,192 @@
...
   }
 }
```
sshah-wework: can we try the empty list case?

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -14,72 +14,205 @@
...
+    namespaceDao.insert(namespaceRow);
+    datasourceDao.insert(datasourceRow);
+
+    final DatasetRow newDatasetRow =
+        newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid());
+    datasetDao.insert(newDatasetRow);
+
+    final int rowsAfter = datasetDao.count();
+    assertThat(rowsAfter).isEqualTo(rowsBefore + 1);
+  }
```
sshah-wework: should we test the double insert and verify idempotence?

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -14,72 +14,205 @@
...
+    datasourceDao.insert(datasourceRow);
+
+    final DatasetUrn datasetUrn = newDatasetUrn();
+    final DatasetRow newDatasetRow =
+        newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid(), datasetUrn);
+    datasetDao.insertAndGet(newDatasetRow);
+
+    final DatasetRow datasetRow = datasetDao.findBy(datasetUrn).orElse(null);
+    assertThat(datasetRow).isNotNull();
+    assertThat(datasetRow.getUrn()).isEqualTo(datasetUrn.getValue());
```
sshah-wework: what about the case where `findBy` returns nothing?

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -14,72 +14,205 @@
...
+    namespaceDao.insert(namespaceRow);
+    datasourceDao.insert(datasourceRow);
+
+    final List<DatasetRow> newDatasetRows =
+        newDatasetRowsWith(namespaceRow.getUuid(), datasourceRow.getUuid(), 4);
+    newDatasetRows.forEach(newDatasetRow -> datasetDao.insert(newDatasetRow));
+
+    final List<DatasetRow> datasetRows = datasetDao.findAll(namespaceName, LIMIT, OFFSET);
+    assertThat(datasetRows).isNotNull();
+    assertThat(datasetRows).hasSize(4);
```
sshah-wework: empty list case?

src/test/java/marquez/db/NamespaceDaoTest.java
```diff
@@ -14,91 +14,97 @@
...
-    assertEquals(ns1.getOwnerName(), ns2.getOwnerName());
+  @Test
+  public void testInsert() {
+    final int rowsBefore = namespaceDao.count();
+
+    final NamespaceRow newNamespaceRow = newNamespaceRow();
+    namespaceDao.insert(newNamespaceRow);
+
+    final int rowsAfter = namespaceDao.count();
+    assertThat(rowsAfter).isEqualTo(rowsBefore + 1);
```
sshah-wework: verify idempotence?

src/test/java/marquez/db/NamespaceDaoTest.java
```diff
@@ -14,91 +14,97 @@
...
-  public void testFindAll_Empty() {
-    assertEquals(0, namespaceDao.findAll().size());
+  public void testFindBy_name() {
+    final NamespaceName namespaceName = newNamespaceName();
+    final NamespaceRow newNamespaceRow = newNamespaceRowWith(namespaceName);
+    namespaceDao.insert(newNamespaceRow);
+
+    final NamespaceRow namespaceRow = namespaceDao.findBy(namespaceName).orElse(null);
+    assertThat(namespaceRow).isNotNull();
+    assertThat(namespaceRow.getName()).isEqualTo(namespaceName.getValue());
```
sshah-wework: case where `findBy` returns nothing

src/test/java/marquez/db/NamespaceDaoTest.java
```diff
@@ -14,91 +14,97 @@
...
-    assertTrue(namespaceDao.exists(found));
+  public void testFindAll() {
+    final List<NamespaceRow> newNamespaceRows = newNamespaceRows(4);
+    newNamespaceRows.forEach(newNamespaceRow -> namespaceDao.insert(newNamespaceRow));
 
-    final NamespaceName notFound = NamespaceName.fromString("test");
-    assertFalse(namespaceDao.exists(notFound));
+    final List<NamespaceRow> namespaceRows = namespaceDao.findAll();
+    assertThat(namespaceRows).isNotNull();
+    assertThat(namespaceRows).hasSize(4);
```
sshah-wework: empty list case?

src/test/java/marquez/db/mappers/NamespaceRowMapperTest.java
```diff
@@ -0,0 +1,116 @@
...
+import marquez.common.models.NamespaceName;
+import marquez.common.models.OwnerName;
+import marquez.db.Columns;
+import marquez.db.models.NamespaceRow;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(UnitTests.class)
+public class NamespaceRowMapperTest {
```
sshah-wework: does the mapper handle `updatedAt`?

src/test/java/marquez/service/NamespaceServiceTest.java
```diff
@@ -1,132 +1,163 @@
...
-    namespaceService.create(ns);
+    verify(namespaceDao, times(1)).insert(any(NamespaceRow.class));
   }
 
-  @Test(expected = MarquezServiceException.class)
-  public void testCreate_insertException() throws MarquezServiceException {
-    Namespace ns = Generator.genNamespace();
+  @Test
+  public void testCreate_throwsException_onDbError() throws MarquezServiceException {
+    reset(namespaceDao);
```
sshah-wework: why the reset here?

src/test/java/marquez/service/NamespaceServiceTest.java
```diff
@@ -1,132 +1,163 @@
...
-    doThrow(UnableToExecuteStatementException.class).when(namespaceDao).findBy(NAMESPACE_NAME);
-    namespaceService.get(NAMESPACE_NAME);
+  @Test
+  public void testExists() throws MarquezServiceException {
+    when(namespaceDao.exists(NAMESPACE_NAME)).thenReturn(true);
+
+    final boolean exists = namespaceService.exists(NAMESPACE_NAME);
+    assertThat(exists).isTrue();
+
+    verify(namespaceDao, times(1)).exists(NAMESPACE_NAME);
```
sshah-wework: case where namespace doesn't exist?

src/test/java/marquez/service/NamespaceServiceTest.java
```diff
@@ -1,132 +1,163 @@
...
-    assertEquals(0, namespacesFound.size());
+  public void testList() throws MarquezServiceException {
+    final List<NamespaceRow> namespaceRows = newNamespaceRows(4);
+    when(namespaceDao.findAll()).thenReturn(namespaceRows);
+
+    final List<Namespace> datasets = namespaceService.getAll();
+    assertThat(datasets).isNotNull();
+    assertThat(datasets).hasSize(4);
+
+    verify(namespaceDao, times(1)).findAll();
```
sshah-wework: empty list case?

src/test/java/marquez/service/NamespaceServiceTest.java
wslulciuc: A reset of the `NamespaceDao` mock is needed because `NamespaceService` now has an [`init()`](https://github.com/MarquezProject/marquez/blob/feature/dataset-resource-create-get-ops/src/main/java/marquez/service/NamespaceService.java#L41) methods which calls out to the DB in order to create the _default_ namespace.

Meaning, for the test to pass we would need the following mock verification (note the `2` count):

```java
verify(namespaceDao, times(2)).insert(any(NamespaceRow.class));
```

instead of:

```java
verify(namespaceDao, times(1)).insert(any(NamespaceRow.class));
```

_To sum up_: the reset allows us to verify the insert was the result of the service (**after** instantiation) performing a DAO call instead of **on** service instantiation.

wslulciuc:
@sshah-wework: Thanks! Yeah, let's totally open issues for the test cases you mentioned. I think a far amount of code has been introduced in this PR, so we'll open separate PRs to account for the additional test scenarios.

src/test/java/marquez/service/NamespaceServiceTest.java
wslulciuc: Probably worth commenting on why!

PR #397, opened 2019-03-12
[Datasource Service - Part 4]: Adjust datasource create response code

ashulmanWeWork:


src/main/java/marquez/api/resources/DatasourceResource.java
```diff
@@ -101,6 +101,8 @@ public Response create(@NonNull final DatasourceRequest datasourceRequest)
     final Datasource createdDatasource =
         datasourceService.create(
             connectionUrl, DatasourceName.fromString(datasourceRequest.getName()));
-    return Response.ok(DatasourceResponseMapper.map(createdDatasource)).build();
+    return Response.status(Response.Status.CREATED)
+        .entity(DatasourceResponseMapper.map(createdDatasource))
```
wslulciuc: Minor: For readability, I would convert a `Datasource` to a `DatasourceResponse` separate from the return statement and statically import `Response.Status.CREATED`:

```java
import static javax.ws.rs.core.Response.Status.CREATED;
...
final DatasourceResponse response = DatasourceResponseMapper.map(datasource)
return Response.status(CREATED).entity(response).build();
```

PR #398, opened 2019-03-13
Cleanup class DatasourceUrn and DatasetUrn

wslulciuc:
This PR makes the following changes / additions to our URN classes:

- [`UrnPattern`](https://github.com/MarquezProject/marquez/blob/feature/cleanup-urn-model/src/main/java/marquez/common/models/UrnPattern.java) - Added to encapsulate the URN regex pattern
- The **class** [`Urn`](https://github.com/MarquezProject/marquez/blob/feature/cleanup-urn-model/src/main/java/marquez/common/models/Urn.java) is now **abstract**
- `Urn.valueFrom()` - A helper method to build `Urn.Value` from the parts that make up the URN
- `Urn.namespace()` - A package-private abstract method returning the URN namespace (see _Terminology_ section in [rfc8141](https://tools.ietf.org/html/rfc8141#section-1.1) for namespace defintion)
- `Urn.pattern()` - An abstract method returning the patten of the URN 
- [`DatasourceName`](https://github.com/MarquezProject/marquez/blob/feature/cleanup-urn-model/src/main/java/marquez/common/models/DatasourceName.java) - Add type to encapsulate a datasource name

### Resources:

* URN RFC - https://tools.ietf.org/html/rfc8141


src/main/java/marquez/common/models/UrnType.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package marquez.common.models;
+
+public enum UrnType {
```
sshah-wework: Is it necessary to have `Type` in the class name since it is a type?

src/main/java/marquez/common/models/DatasourceName.java
```diff
@@ -0,0 +1,28 @@
+/*
```
sshah-wework: Is this necessary as part of this PR? Also, we're starting to head down a path of every field in the database which requires some validation gets a concrete Java class. While I like the typing, I also think this can get a bit out of hand.

src/main/java/marquez/common/models/UrnType.java
```diff
@@ -0,0 +1,25 @@
...
+  DATASET,
+  DATASOURCE;
+
+  @Override
+  public String toString() {
+    return name().toLowerCase();
+  }
+}
```
ashulmanWeWork: Discussed in-person, but I think it makes sense to add the number of parts as part of the definition of this enum. For example, we'll never have a DATASOURCE URN with 5 parts. It'll have to be exactly 2. So let's just codify that and disallow people to create invalid scenarios that will error out down the line.

ashulmanWeWork: Additionally, we should have a few unit tests, where we exercise `valueOf` and convert back and forth. It'll remove issues with capitalization, etc.

src/main/java/marquez/common/models/DatasourceUrn.java
```diff
@@ -1,24 +1,56 @@
...
+        new StringJoiner(URN_DELIM)
+            .add(URN_PREFIX)
+            .add(UrnType.DATASOURCE.toString())
+            .add(type.toString())
+            .add(datasourceName.getValue())
+            .toString();
     return fromString(value);
   }
 
+  @JsonProperty
```
ashulmanWeWork: Why is this necessary?

src/main/java/marquez/common/models/DatasourceType.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package marquez.common.models;
+
+public enum DatasourceType {
```
sshah-wework: not 100% sure this is related.. would like to consider `DatasourceType` separately from the changes to `DatasourceURN` and `DatasetURN`.

I'd like to eventually see #401 happen where the allowed types are config driven which will make it easier to extend Marquez Core simply with a config change. Enums can only be statically defined, so it may not support a config-driven list. No need to change now, but we should address when we tackle #401.

src/main/java/marquez/common/models/UrnType.java
```diff
@@ -14,9 +14,21 @@
...
 
+import lombok.NonNull;
+
 public enum UrnType {
-  DATASET,
-  DATASOURCE;
+  DATASET(2),
+  DATASOURCE(2);
+
+  private final Integer value;
```
ashulmanWeWork: nit: Value is too overloaded a term when it comes to enums - maybe we can call this `numParts` or something more descriptive?

src/main/java/marquez/common/models/DatasourceUrn.java
```diff
@@ -1,24 +1,56 @@
...
-  public DatasourceUrn(@NonNull String value) {
-    super(value, REGEX);
+  private DatasourceUrn(@NonNull final String value) {
+    URN_PATTERN.throwIfNoMatch(checkNotBlank(value));
+    this.value = value;
   }
 
-  public static DatasourceUrn from(@NonNull String type, @NonNull String name) {
-    final String value = fromComponents(URN_TYPE, type, name);
+  public static DatasourceUrn from(
```
sshah-wework: why not build the string within `UrnPattern` as well since `UrnPattern` is enforcing a consistent format with `throwIfNotMatch`?

sshah-wework:
Did you remove the char stripping?

src/main/java/marquez/common/models/UrnType.java
sshah-wework: basically, why do we need `UrnType` to enforce what is a valid Urn type (e.g. `dataset`, `datasource`,...)? the concrete classes (e.g. `DatasourceUrn`) already do that

src/main/java/marquez/common/models/DatasourceName.java
wslulciuc: Added `DatasourceName` to allow for **DatasourceUrn.from**(_DatasourceType_, _DatasourceName_), enabling us to use the type when creating a new datasource URN.

Good point. We really aren't doing much validation for the types we have defined, specifically if they are just wrapped strings. So, agree not much value is added (_currently_) in terms of validation. But, types allow for adding both semantics and context to the params passed to our methods. In the future, types will look more like:

```java
@EqualsAndHashCode
@ToString
public final class DatasourceName {
  @Getter @NonNull private final String value;

  private DatasourceName(@NonNull  final String value) {
    checkNotBlank(value)
    checkArgument(value.length() > 128)
    this.value = value
  }
}
```

Personally, if we are accepting input from callers, we should validate and ensure there's a type defined. But, maybe that's something we should discuss more broadly as a team :)

src/main/java/marquez/common/models/DatasourceType.java
wslulciuc: Nice! I haven't given the _config driven_ approach to defining datasources much thought, but I see how it can be a powerful customization option. Thanks for starting the discussion! 🥇

src/main/java/marquez/common/models/DatasourceUrn.java
wslulciuc: Good catch. This actually should be `@JsonCreator`. That is, `fromString()` is the factory method Jackson should use on deserialization.

src/main/java/marquez/common/models/DatasourceType.java
sshah-wework: /cc @roaraya8 @ashulmanWeWork thoughts?

src/main/java/marquez/common/models/UrnType.java
wslulciuc: Yeah, think you're right, and this might be overkill. If we define a URN class, it assumes a type and therefore defining an `enum` only externalizes how a URN is really created. Also, means it would require more than one file to be touched. I'll go ahead and remove the enum.

src/main/java/marquez/common/models/UrnType.java
wslulciuc: I'll be removing the enum (see above comment) and we can just use `numberOfParts` to communicate the number of parts in the URN.

src/main/java/marquez/common/models/UrnType.java
ashulmanWeWork: Agreed

src/main/java/marquez/common/models/DatasourceType.java
ashulmanWeWork: Wanted to understand more - is it the case that we just won't be able to fully anticipate all the types of datasources that an organization using Marquez will want to track? I guess that would block an organization from using Marquez fully until the release where their datasource is an entry in the enum. If that's the case, why not just make this a free-form field instead of an enum? 

OTOH, suppose an admin wants to restrict the types, but only to the types that the admin deems appropriate? If so, seems healthier if that's established as a team decision and not a configuration owned by the Marquez admin. Or am I overthinking this? :)

src/main/java/marquez/common/models/DatasourceUrn.java
wslulciuc: Depends on the definition of `UrnPattern`. I view it more as: _representing and enforcing a pattern_ vs building the URN string. That is, with patterns, we ask does string `S` match pattern `P`. Combining functionality would clean the code quite a bit. Maybe `UrnPattern.buildOrthrowIfNotMatch()` would make sense?

wslulciuc:
> Did you remove the char stripping?

@sshah-wework I didn't account for stripping URN strings given that this PR is more experimental and wanted to see how sensible the changes were.

src/main/java/marquez/common/models/UrnType.java
wslulciuc: We decided to remove this type in favor of defining the namespace (internally) as part of the class

src/main/java/marquez/common/models/DatasourceUrn.java
wslulciuc: Really appreciate the comment, @sshah-wework! It inspired a much cleaner approach that both enforces and builds a URN string in the base Urn class. That is, with the introduction of the method `pattern()`,  we can enforce Urn validation in a single place (on Urn instantiation!)

```java
protected Urn(@NonNull final String value) {
  pattern().throwIfNoMatch(checkNotBlank(value));
  this.value = value;
}
```

src/main/java/marquez/common/models/DatasourceUrn.java
```diff
@@ -1,25 +1,50 @@
...
   }
 
+  @JsonCreator
   public static DatasourceUrn fromString(String value) {
     return new DatasourceUrn(value);
   }
+
+  @Override
+  public String namespace() {
+    return NAMESPACE;
```
ashulmanWeWork: Can this be brought into the parent class, since it's implemented the same way here and also in `DatasetUrn`, both of which extend `Urn`?

src/main/java/marquez/common/models/DatasourceName.java
wslulciuc: Will be addressed in #404

src/main/java/marquez/common/models/UrnPattern.java
```diff
@@ -0,0 +1,54 @@
...
+  static final String DELIM = ":";
+  static final String PREFIX = "urn";
+
+  private final Pattern pattern;
+
+  private UrnPattern(@NonNull final String value) {
+    this.pattern = Pattern.compile(checkNotBlank(value));
+  }
+
+  static UrnPattern from(@NonNull String namespace, @NonNull Integer numOfParts) {
```
ashulmanWeWork: We should safeguard that `numOfParts` is greater than zero.

src/test/java/marquez/common/models/DatasetUrnTest.java
```diff
@@ -14,96 +14,61 @@
...
+
+  @Test(expected = NullPointerException.class)
+  public void testFrom_throwsException_onNullDatasetName() {
+    final DatasetName nullDatasetName = null;
+    DatasetUrn.from(DATASOURCE_NAME, nullDatasetName);
   }
 
   @Test(expected = NullPointerException.class)
-  public void testNewDatasetUrn_throwsException_onNullValue() {
+  public void testFromStringn_throwsException_onNullValue() {
```
ashulmanWeWork: nit: `testFromStringn_` -> `testFromString_`

src/test/java/marquez/common/models/UrnPatternTest.java
```diff
@@ -0,0 +1,116 @@
...
+    PATTERN.throwIfNoMatch(partGreaterThan64Value);
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void testThrowIfNoMatch_throwsException_onNullValue() {
+    final String nullValue = null;
+    PATTERN.throwIfNoMatch(nullValue);
+  }
+
+  private String newUrnWithPartGreaterThan64() {
```
ashulmanWeWork: To promote a one-stop-shop for generating test members, this can be moved to the `Generator` class.

wslulciuc: `newUrnWithPartGreaterThan64()` is strictly for testing the Urn class and probably not something that we'd want in **class** `Generator`.

We'll eventually want to revisit `Generator` usage to clean up a few things. Might be worth having the conversation then on where this method should live.

src/main/java/marquez/common/models/DatasourceUrn.java
wslulciuc: The abstract methods `namespace()` and `pattern()` forces anyone who extends the **class** `Urn` to implement those methods. Giving us a clear contract of how to interface with and define a URN.

So, yeah, there is some redundancy, but the current approach favors a more formal definition of a URN.

roaraya8:
Looks pretty good to me, much cleaner than before.

src/main/java/marquez/common/models/DatasourceType.java
wslulciuc: Let's move the discussion to the issue :)

PR #405, opened 2019-03-18
Adding openapi spec definition for datasources

ashulmanWeWork:


ashulmanWeWork:
Testing done - I used this spec to generate a new API client. It did so without issues and the relevant `datasource`-related classes were created as expected.

docs/openapi.yml
```diff
@@ -278,6 +278,53 @@ paths:
...
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/datasources'
+    post:
+      summary: Create a new datasource
+      description: Creates a new datasource and returns the created datasource
```
wslulciuc: Minor: Let's be consistent with how we describe endpoints. I recommend looking over the descriptions for endpoints that have been already defined.

docs/openapi.yml
```diff
@@ -278,6 +278,53 @@ paths:
...
+              $ref: '#/components/schemas/createDatasource'
+      responses:
+        '201':
+          description: CREATED
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/datasources'
+
+  /datasources/{datasourceUrn}:
```
wslulciuc: Minor: `/datasources/{urn}` **=>** `/datasources/{urn}`

docs/openapi.yml
```diff
@@ -309,6 +356,17 @@ components:
...
+    datasourceUrn:
+      name: datasourceUrn
+      in: path
+      description: The URN of the datasource.
+      required: true
+      schema:
+        type: string
+        format: URN
+        maxLength: 1024
+        example: urn:datasource:jdbcpostgresqllocalhost5431novelists834:Datasource2168
```
wslulciuc: Minor: This most likely isn't the URN returned by the API if you provide:

```
{
  "name": "Datasource2168",
  "connectionUrl": "jdbc:postgresql://localhost:5431/novelists_834"
}
```

Mind verifying the response?

ashulmanWeWork: Ran the inputs/outputs through a locally-running instance of Marquez and updated the results.

Perhaps a nit, but I also think that hostname `localhost` may not make a lot of sense here, just b/c we our storage of datastore to have a universally-identifiable host (so it can be repro'd, audited, etc.). If folks think it's worth changing, I can change it.

docs/openapi.yml
ashulmanWeWork: I updated the create description to be much more informative, similar to what is done for namespace creation and job creation. Also updated for punctuational consistency. Is there anything else I should be looking at to make it consistent with the rest of the file?

docs/openapi.yml
```diff
@@ -278,6 +278,62 @@ paths:
...
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/datasources'
+    post:
+      summary: Create a datasource
+      description: "Creates a new datasource and returns the created datasource.
```
wslulciuc: **description**: Creates a new datasource object. A datasource is the physical location of a dataset such as a table in `PostgreSQL`, or topic in `Kafka`. A datasource enables the grouping of physical datasets to their physical source.

docs/openapi.yml
```diff
@@ -278,6 +278,62 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/datasources'
+
+  /datasources/{urn}:
+    parameters:
+      - $ref: '#/components/parameters/datasourceUrn'
+    get:
+      summary: Get a datasource
+      description: 'Returns a datasource response for the supplied urn.
```
wslulciuc: Let's simplify the wording for now:

**description**:  Retrieve a datasource.

We'll want to eventually introduce a section to the spec that describes the format of each URN and their usage.

PR #415, opened 2019-03-23
Update NamespaceService and NamespaceDao interface

wslulciuc:
This PR introduces the following changes:

- Use `NamespaceName` type as param in `NamespaceService`
- Use `NamespaceName` type as param in `NamespaceDao`
- Rename `NamespaceResource.listNamespaces()` **=>** `NamespaceResource.list()`
- Rename `NamespaceService.listNamespaces()` **=>** `NamespaceService.getAll()`
- Rename `NamespaceDao.find()` **=>** `NamespaceDao.findBy()`
- Add `NamespaceDao.exists()`

### DIFF
--- a/src/test/java/marquez/api/JobResourceTest.java
+++ b/src/test/java/marquez/api/JobResourceTest.java
@@ -41,6 +41,7 @@ import marquez.api.models.JobRunRequest;
 import marquez.api.models.JobRunResponse;
 import marquez.api.models.JobsResponse;
 import marquez.api.resources.JobResource;
+import marquez.common.models.NamespaceName;
 import marquez.service.JobService;
 import marquez.service.NamespaceService;
 import marquez.service.exceptions.MarquezServiceException;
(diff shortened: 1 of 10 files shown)


src/test/java/marquez/api/JobResourceTest.java
```diff
@@ -56,7 +57,7 @@
   private static final NamespaceService MOCK_NAMESPACE_SERVICE = mock(NamespaceService.class);
   private static final JobResource JOB_RESOURCE =
       new JobResource(MOCK_NAMESPACE_SERVICE, MOCK_JOB_SERVICE);
-  private static final String NAMESPACE_NAME = "someNamespace";
+  private static final NamespaceName NAMESPACE_NAME = NamespaceName.fromString("test");
```
sshah-wework: we'll want a generator for `NamespaceName`s ... might be a good opportunity to start a `marquez.common.models.Generator`

wslulciuc: You're right. Let's open an epic around testing changes (i.e. clean up tasks for test suites, providing solid test examples for new comers to our codebase, etc). So, I'll hold off for now, but something we will be revisiting soon.

PR #417, opened 2019-03-24
Default Description.NO_VALUE to null

### LINKED ISSUE #311: Default column value to null when description not present
We currently default the description column for `Namespace`, `Job`, `Dataset`, etc  to an empty string (`""`).  Let's use `null` instead to signify an unknown or not given description. 

see: https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/common/models/Description.java#L14

wslulciuc:
Closes #311


src/main/java/marquez/api/models/DatasetVersionRequest.java
```diff
@@ -26,7 +27,7 @@
 @JsonSubTypes({@JsonSubTypes.Type(value = DbTableVersionRequest.class, name = "DB")})
 public abstract class DatasetVersionRequest {
   @Getter @NonNull private final DatasetType type;
-  private final String description;
+  @Nullable private final String description;
```
sshah-wework: this is an important behavior change that is not captured in the description for this PR -- descriptions are now nullable. This PR should probably update the OpenAPI.yaml simultaneously:

https://github.com/MarquezProject/marquez/blob/master/docs/openapi.yml#L337

sshah-wework: Woops, that's not the right line I linked to, will fix.

sshah-wework: nvm, looks like it's not actually marked required in the OpenAPI.yaml :)

sshah-wework: we're good here, but now I'm noticing that the OpenAPI.yaml is a bit behind with required fields. Will file a ticket to review that and make sure we're more consistent across our endpoints.

src/main/java/marquez/common/models/Description.java
```diff
@@ -14,18 +14,18 @@
...
 import lombok.ToString;
 
-@RequiredArgsConstructor(staticName = "fromString")
+@AllArgsConstructor(staticName = "fromString")
 @EqualsAndHashCode
 @ToString
 public final class Description {
-  public static final String NO_VALUE = "";
+  public static final String NO_VALUE = null;
   public static final Description NO_DESCRIPTION = new Description(NO_VALUE);
```
sshah-wework: Hm, not sure I understand `NO_DESCRIPTION`.. it's less clear what's happening elsewhere in the code. `new Description(NO_VALUE)` is almost easier to understand

sshah-wework: If we're going to avoid using `null` to signify that something is not set, then I think that value should live outside of `marquez.common.models.Description` and should be used universally in place of `null`. I haven't worked with a ton of java codebases, but to me it feels very natural to use `null` when something is not set. Is this considered bad form?

wslulciuc: > Hm, not sure I understand NO_DESCRIPTION.. it's less clear what's happening elsewhere in the code. new Description(NO_VALUE) is almost easier to understand

The static instance `Description.NO_DESCRIPTION` is more for convenience and expressibility. It's a type that may or may not be present (`null`, empty / blank string), so we want to handle the optional type differently, or at least attempt to avoid the occurrences of `new Description(null)` throughout our codebase.

For example, let's look at an example `Row` class:

```java
public final class Row {
  @Getter private final UUID uuid;
  @Getter private final Instant createdAt;
  @Nullable private final Description description;

  public Optional<String> getDescription() {
    return Optional.ofNullable(description);
  }
}
```

It's clear `Row.Description` maybe `null`. Let's say we want to read the `DatasetRow.Description`. We can either:

```java
// Option 1
final Row row = ...
final Description description = row.getDescription().orElse(new Description(null))
```

or use a statically defined instance given the common occurrence of attempting the access a row description

```java
import static marquez.common.models.Description.NO_DESCRIPTION;

// Option 1
final Row row = ...
final Description description = row.getDescription().orElse(NO_DESCRIPTION)
```

> If we're going to avoid using null to signify that something is not set, then I think that value should live outside of marquez.common.models.Description

I don't think `null`s in a codebase is a bad thing, and in most cases unavoidable. To begin replacing all `null`s with `NO_VALUE` seems a bit much and am ok with removing the constant. That is, 

```java
public final class Description {
  ...
  public static final Description NO_DESCRIPTION = new Description(null);
}
```

sshah-wework: Gotcha.
I'd be ok with keeping `NO_DESCRIPTION`, but would rather stick with `null`s over `NO_VALUE`s in the code.

PR #418, opened 2019-03-24
Use @AllArgsConstructor and standardized on lombok annotations

wslulciuc:


src/test/java/marquez/api/JobResourceTest.java
```diff
@@ -122,10 +122,9 @@ public void testCreateJobBadInputs() throws MarquezServiceException {
     when(MOCK_NAMESPACE_SERVICE.get(any())).thenReturn(Optional.of(Generator.genNamespace()));
 
     JobResponse jobForJobCreationRequest = generateApiJob();
-    jobForJobCreationRequest.setLocation(null);
 
     Response res = insertJob(jobForJobCreationRequest);
-    assertEquals(UNPROCESSABLE_ENTRY_STATUS_CODE, res.getStatus());
+    assertEquals(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), res.getStatus());
```
ashulmanWeWork: Can we do a static import here?

wslulciuc: Yeah, I thought about this! But, all status codes are referenced non-statically in `JobResourceTest`, so wanted to be consistent. I'll will be opening an epic around tests that will make sure to capture the change across all tests.

src/test/java/marquez/api/mappers/NamespaceResponseMapperTest.java
```diff
@@ -44,7 +44,8 @@ public void testSingleNamespace() {
     assertThat(Timestamp.valueOf(mappedResponse.getCreatedAt()))
         .isEqualTo(descriptionNamespace.getCreatedAt());
     assertThat(mappedResponse.getOwner()).isEqualTo(descriptionNamespace.getOwnerName());
-    assertThat(mappedResponse.getDescription().orElse(null)).isEqualTo(descriptionNamespace.getDescription());
+    assertThat(mappedResponse.getDescription().orElse(null))
```
ashulmanWeWork: Since we know that description here is not null, shouldn't we add a test for mapperResponse.getDescription.isPresent() ?

src/test/java/marquez/api/mappers/NamespaceResponseMapperTest.java
```diff
@@ -44,7 +44,8 @@ public void testSingleNamespace() {
     assertThat(Timestamp.valueOf(mappedResponse.getCreatedAt()))
         .isEqualTo(descriptionNamespace.getCreatedAt());
```
ashulmanWeWork: Can't directly comment on this b/c it's not part of the review, but we can make all the member variables `private`, `final`, and `static`.

src/main/java/marquez/api/models/JobRunResponse.java
```diff
@@ -14,23 +14,34 @@
...
-@NoArgsConstructor
+@EqualsAndHashCode
+@ToString
 public final class JobRunResponse {
-  @NotBlank private UUID runId;
+  @Getter @NonNull private final UUID runId;
+  @Nullable private final String nominalStartTime;
+  @Nullable private final String nominalEndTime;
+  @Nullable private final String runArgs;
+  @Getter @NonNull private final String runState;
```
sshah-wework: ^^ seems like we're going beyond just lombok annotations here. But since you're adding JobRun fields in this PR, what about `createdAt`?

wslulciuc: Sure. But I only annotated the fields and no new fields were added. The current field list is: `runId`, `nominalStartTime`, `runArgs`, and `runState`. I'll added `createdAt` and update the openapi spec

wslulciuc:
I'm closing this PR as there are better ways we should be annotating our classes with lombok that I have in mind :)

PR #420, opened 2019-03-25
Add continuous autogen deployment

ashulmanWeWork:


release_autogen_client/common_packaging_utils.sh
```diff
@@ -0,0 +1,99 @@
...
+{
+  clone_dir=${MARQUEZ_CLONE_DIR}
+  rm -rf ${clone_dir} || true
+
+  git clone --depth=1 https://github.com/MarquezProject/marquez.git ${clone_dir}
+}
+
+update_config_file()
+{
+  sed -i '' "s/packageVersion.*/packageVersion\": \"${version}\",/g" ${CONFIG_FILE_LOCATION}
```
ashulmanWeWork: @wslulciuc Updated to use the existing config file. Verified that the produced diff in the `marquez-python-codegen` project contains `config.json` as well, so that there will be a record of what was used to auto-generate the rest of the client.

release_autogen_client/common_packaging_utils.sh
```diff
@@ -0,0 +1,99 @@
...
+  # Please see bumpversion --help for more information.
+  cd ${MARQUEZ_PYTHON_CODEGEN_CLONE_DIR}
+  current_version=$(python ${MARQUEZ_PYTHON_CODEGEN_CLONE_DIR}/setup.py --version)
+  if [ -n "${new_version}" ]; then
+    echo "Upping the version to be ${new_version}"
+    # Note: the type here, patch, does not matter as we are just upgrading
+    # to a specified version. It's required to satisfy the arg parsing.
+    bumpversion --current-version ${current_version} --new-version ${new_version} --commit patch ./setup.py
+  else
+    bumpversion --current-version ${current_version} --commit ${type} ./setup.py
```
wslulciuc: We'll want to add a check ensuring `bumpversion` is installed. This will make sure we don't get into a weird state on client gen (i.e. the script won't execute the main logic until all requirements are met).

release_autogen_client/common_packaging_utils.sh
```diff
@@ -0,0 +1,99 @@
...
+export CONFIG_FILE_LOCATION="${MARQUEZ_PYTHON_CODEGEN_CLONE_DIR}/config.json"
+
+clone_marquez_python_client_codegen()
+{
+  clone_dir=${MARQUEZ_PYTHON_CODEGEN_CLONE_DIR}
+  rm -rf ${clone_dir} || true
+  git clone https://github.com/MarquezProject/marquez-python-codegen.git ${clone_dir}
+}
+
+clone_marquez()
```
wslulciuc: Is there a reason we are also cloning the `marquez` repo as well? The added _precaution_ step makes the flow harder to follow.

release_autogen_client/RELEASING.md
```diff
@@ -0,0 +1,22 @@
...
+
+Example usage:
+1. We want to release a Marquez-Python-Codegen client with version 1.2.0, perhaps because Marquez is now versioned to that value:
+`./update_auto_gen_client_specific_version.sh 1.2.0`
+
+2. We want to do a patch increment of the Marquez-Python-Codegen client to incorporate some new, non-breaking functionality:
+`./update_auto_gen_client_semantic_version.sh patch`
+
+3. We want to do a patch increment of the Marquez-Python-Codegen client to incorporate a new breaking change:
+`./update_auto_gen_client_semantic_version.sh major`
```
wslulciuc: I'd add the script usage to the header of the script.

wslulciuc: Is it necessary to have two scripts (`update_auto_gen_client_specific_version`, `update_auto_gen_client_semantic_version `) that handle generating our python client from the openAPI spec? Feel like the logic can be self contained in a single script.

sshah-wework:
Could this be simplified down to one script which does:

1) check out `marquez-python-autogen` in a tmp dir
2) set the version in `marquez-python-autogen`'s config.json
3) generate the client
4) commit to `marquez-python-autogen`
5) push to `marquez-python-autogen` to pypi?

release_autogen_client/update_auto_gen_client_specific_version.sh
```diff
@@ -0,0 +1,31 @@
...
+
+export new_version=${1}
+echo "new_version is ${1}"
+if [ -z "${new_version}" ]; then
+  echo "required parameter 'version' not provided. Please specify in form {major}.{minor}.{patch}"
+  exit 1
+fi
+
+setup_repos
+refresh_codegen
```
sshah-wework: Couldn't all of these scripts be simplified down to one script which does:

check out marquez-python-autogen in a tmp dir
set the version in marquez-python-autogen's config.json
generate the client
commit to marquez-python-autogen
push to marquez-python-autogen to pypi?

release_autogen_client/common_packaging_utils.sh
ashulmanWeWork: If we've settled on Marquez being the repo to hold the scripts, I'm happy to use the existing cloned repo.

release_autogen_client/update_auto_gen_client_specific_version.sh
ashulmanWeWork: Sure, happy to do that. The only exception is the last step, which I feel is better handled by CircleCI, in part b/c it very conveniently keeps the credentials safe, but also b/c I feel more comfortable with CircleCI doing the push to PyPi since it will keep a record about exactly what happened. Does that seem okay?

scripts/RELEASING.md
```diff
@@ -0,0 +1,22 @@
...
+
+Example usage:
+1. We want to release a Marquez-Python-Codegen client with version 1.2.0, perhaps because Marquez is now versioned to that value:
+`./autogen.sh --set-version 1.2.0`
+
+2. We want to do a patch increment of the Marquez-Python-Codegen client to incorporate some new, non-breaking functionality:
+`./autogen.sh --semantic-bump patch`
+
+3. We want to do a major increment of the Marquez-Python-Codegen client to incorporate a new breaking change:
+`./autogen.sh --semantic-bump major`
```
wslulciuc: I'd move these steps to `openapi/`

ashulmanWeWork: so `./scripts/openapi/autogen.sh` ?

ashulmanWeWork: Oh, moving releasing.md. Will move it.

release_autogen_client/update_auto_gen_client_specific_version.sh
sshah-wework: Yep, that's fine.

PR #428, opened 2019-03-28
Adjust datasource idempotency

ashulmanWeWork:


src/main/java/marquez/db/DatasourceDao.java
```diff
@@ -16,22 +16,26 @@
...
-import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 
 @RegisterRowMapper(DatasourceRowMapper.class)
 public interface DatasourceDao {
-  @SqlUpdate(
+  @SqlQuery(
       "INSERT INTO datasources (guid, urn, name, connection_url) "
-          + "VALUES (:uuid, :urn, :name, :connectionUrl)")
-  void insert(@BindBean DatasourceRow datasourceRow);
+          + "VALUES (:uuid, :urn, :name, :connectionUrl) ON CONFLICT DO NOTHING RETURNING * ")
```
wslulciuc: Minor: I'd make the following changes to `@SqlQuery`:

1. Remove `ON CONFLICT DO NOTHING`. Since we now check to ensure a given datasource row already exists before insertion, we'll want to log occurrences where conflicts do happen rather than handling the conflict silently.
2. Move `RETURNING` to separate lines

```java
@SqlQuery(
    "INSERT INTO datasources (guid, urn, name, connection_url) "
        + "VALUES (:uuid, :urn, :name, :connectionUrl) "
        + "RETURNING * ")
```

src/main/java/marquez/service/DatasourceService.java
```diff
@@ -43,10 +43,14 @@ public Datasource create(@NonNull ConnectionUrl connectionUrl, @NonNull Datasour
     final DatasourceRow row = DatasourceRowMapper.map(connectionUrl, name);
 
     try {
-      datasourceDao.insert(row);
-      final DatasourceUrn urn = DatasourceUrn.fromString(row.getUrn());
-      final Optional<DatasourceRow> rowIfFound = datasourceDao.findBy(urn.getValue());
-      return rowIfFound.map(DatasourceMapper::map).orElseThrow(MarquezServiceException::new);
+      // Check if its already there based on its name. If so, return
+      final Optional<DatasourceRow> existingRowIfFound = datasourceDao.findBy(name);
```
wslulciuc: I think we can use `DatasourceService.get(DatasourceUrn)`? You have everything needed to build the URN.

PR #429, opened 2019-03-28
Add DB and common model generators

wslulciuc:
This PR kicks off our new approach to defining / using generators for tests!

### DIFF
--- a/src/test/java/marquez/common/models/CommonModelGenerator.java
+++ b/src/test/java/marquez/common/models/CommonModelGenerator.java
@@ -1,3 +1,17 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package marquez.common.models;
 
 import java.util.Random;
--- a/src/test/java/marquez/common/models/CommonModelGenerator.java
+++ b/src/test/java/marquez/common/models/CommonModelGenerator.java
@@ -1,0 +1,99 @@
+package marquez.common.models;
+
+import java.util.Random;
+
+public final class CommonModelGenerator {
+  private CommonModelGenerator() {}
+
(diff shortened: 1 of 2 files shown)


src/test/java/marquez/common/models/CommonModelGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+  private CommonModelGenerator() {}
+
+  private static final int BOUND = DatasourceType.values().length;
+  private static final Random RANDOM = new Random();
+
+  public static OwnerName newOwnerName() {
+    return newOwnerNameWith("test_owner" + newId());
+  }
+
+  public static OwnerName newOwnerNameWith(String value) {
```
ashulmanWeWork: Should we add validation, such as `@NotBlank` or `@NonNull`? Should we also make it final?

src/test/java/marquez/db/models/DbModelGenerator.java
```diff
@@ -0,0 +1,82 @@
+package marquez.db.models;
```
ashulmanWeWork: Please add the license info.

src/test/java/marquez/common/models/CommonModelGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+
+  public static Description newDescription() {
+    return newDescriptionWith("test_desciption" + newId());
+  }
+
+  public static Description newDescriptionWith(String value) {
+    return Description.fromString(value);
+  }
+
+  private static int newId() {
```
sshah-wework: we will have something similar in all generators, so let's find a way to standardize.

src/test/java/marquez/db/models/DbModelGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+
+    if (wasUpdated) {
+      builder.updatedAt(newTimestamp());
+      builder.currentVersion(UUID.randomUUID());
+    }
+
+    return builder.build();
+  }
+
+  public static Instant newTimestamp() {
```
sshah-wework: potentially something to make common as well across Generators

src/test/java/marquez/common/models/CommonModelGenerator.java
wslulciuc: For test code we can relax input validation. Let's say an invalid `value` is passed to `newOwnerNameWith`, an exception will be thrown by `wnerName.fromString()`. So, we're covered :)

src/test/java/marquez/common/models/CommonModelGenerator.java
wslulciuc: I think we might be okay just keeping `newId()` in this class as it's used to provide unique names. As we expand the usage of generators, we can begin moving common helper methods into a base generator.

PR #436, opened 2019-04-02
Adjusting enum types and adding additional enums for job run state in…

ashulmanWeWork:
… openapi.yml

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -574,8 +574,8 @@ components:
           format: uuid
         runState:
           description: The current state of the job.
-          type: enum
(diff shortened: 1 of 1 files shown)


docs/openapi.yml
```diff
@@ -574,8 +574,8 @@ components:
           format: uuid
         runState:
           description: The current state of the job.
-          type: enum
-          enum: [STARTED, COMPLETED, FAILED]
+          type: string
+          enum: [NEW, RUNNING, COMPLETED, FAILED, ABORTED]
```
wslulciuc: 💯

PR #438, opened 2019-04-02
Add started at ended at to job run response fixes #437

### LINKED ISSUE #437: Remove startedAt and endedAt from jobrun in openAPI and DB layer

ashulmanWeWork:


ashulmanWeWork:
Spoke with Shawn and Willy and they mentioned we're moving away from using these fields.

PR #439, opened 2019-04-03
Removing startedAt and endedAt from openApi and database - fixes #437

### LINKED ISSUE #437: Remove startedAt and endedAt from jobrun in openAPI and DB layer

ashulmanWeWork:


sshah-wework:
I think there are other references to those columns that need to be cleaned up:

```
src/main/java/marquez/db/Columns.java:  public static final String STARTED_AT = "started_at";
src/main/java/marquez/db/Columns.java:  public static final String ENDED_AT = "ended_at";
```

docs/openapi.yml
```diff
@@ -579,14 +579,6 @@ components:
...
           description: The runtime arguments of the job. JSON encoded.
           type: string
-        startedAt:
-          description: An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the date/time the job started.
-          type: string
-          format: date-time
-        endedAt:
-          description: An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the date/time the job ended.
-          type: string
-          format: date-time
```
wslulciuc: 💯

ashulmanWeWork:
Fixes #437

src/main/java/marquez/db/Columns.java
```diff
@@ -35,9 +35,6 @@ private Columns() {}
   // Namespace ownership column names
 
   public static final String NAMESPACE_UUID = "namespace_guid";
-  public static final String STARTED_AT = "started_at";
-  public static final String ENDED_AT = "ended_at";
-  public static final String OWNER_UUID = "owner_uuid";
```
sshah-wework: this doesn't seem related :(

sshah-wework: try to keep relevant things in the same PR, otherwise it will seem like `owner_uuid` was somehow part of cleaning up `startedAt` and `endedAt`.

sshah-wework:
Will approve, but also remember to stick to the branch name convention. Thanks!

src/main/java/marquez/db/Columns.java
ashulmanWeWork: Thanks, fixed!

ashulmanWeWork:
@wslulciuc I had the fix in the title, but i've also added it into the comment section. I think that was your only objection, so merging in now :)

PR #444, opened 2019-04-11
Add dataset creation and retrieval endpoints in openAPI spec

wslulciuc:
Related PR #395

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -265,6 +265,23 @@ paths:
   /namespaces/{namespace}/datasets:
     parameters:
       - $ref: '#/components/parameters/namespace'
+    post:
+      summary: Create a dataset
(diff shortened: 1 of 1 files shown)


docs/openapi.yml
```diff
@@ -370,6 +404,17 @@ components:
...
+    datasetUrn:
+      name: urn
+      in: path
+      description: The URN of the dataset.
+      required: true
+      schema:
+        type: string
+        format: URN
+        maxLength: 1024
+        example: urn:datasource:postgresql:financials_db
```
ashulmanWeWork: This looks more like a datasource urn?

wslulciuc: Good catch! Getting a bit lazy with my copy/paste :)

PR #452, opened 2019-04-13
Throw NamespaceNotFoundException on NamespaceResource.get()

wslulciuc:


src/main/java/marquez/api/resources/NamespaceResource.java
```diff
@@ -70,13 +70,12 @@ public Response createOrUpdate(
   @Path("/namespaces/{namespace}")
   public Response get(@PathParam("namespace") NamespaceName namespaceName)
```
ashulmanWeWork: Do we need `@Valid` here?

wslulciuc: `@Valid` is used to validate object properties annotated with `@NotBlank`, etc. So, not necessary for **NamespaceName** but very much required for HTTP request payloads.

PR #457, opened 2019-04-16
Add marquezproject.github.io

### LINKED ISSUE #242: Publish design doc
Let's publish the [Marquez draft design](https://drive.google.com/open?id=1zxvp-6jv4Gs7eAGFmK0fbKbYL9cbEQrRJKLk3ez4PRA) to this repo to allow public access to the doc.

wslulciuc:
This PR introduces an overview of Marquez to our website: https://marquezproject.github.io/marquez 

Closes #242

/cc @sshah-wework @ashulmanWeWork @roaraya8


wslulciuc:
@julienledem @hougs @villasv @whugoh, would love to get your feedback on the overview of Marquez in [`index.md`](https://github.com/MarquezProject/marquez/blob/feature/marquezproject.github.io/docs/index.md)

villasv:
👏 specially liked the Design, Data Model, and Talks sections

docs/index.md
```diff
@@ -1 +1,95 @@
+---
+layout: index
+---
 
+## Overview
+
+Marquez is an open source **metadata service** for the **collection**, **aggregation**, and **visualization** of a data ecosystem's metadata. It maintains the [provenance](https://en.wikipedia.org/wiki/Provenance#Data_provenance) of how datasets are consumed and produced, provides global visibility into job runtime and frequency of dataset access, centralization of dataset lifecycle management, and much more. Marquez<sup>[1](#1)</sup> was released and open sourced by [The We Company](https://www.we.co).
```
alexlingww: This is definitely due to my lack of background, but in this overview it states Marquez is about data provenance, but in the link it links to data lineage which is also mentioned below.

I guess my confusion is that data lineage and data provenance seem like the same thing?

docs/index.md
```diff
@@ -1 +1,95 @@
...
+
+## Why Marquez?
```
alexlingww: Seems to me there needs to be a merger of this section and the tiny section below. The points in the below section seem to map to some of the points in this section. 

Maybe even going as far as making diagrams for this section too?

docs/index.md
```diff
@@ -1 +1,95 @@
...
+
+## Why manage and utilize metadata?
+
+<figure align="center">
+  <img src="./assets/images/ecosystem.png">
+</figure>
+
+## Design
+
+Marquez is a modular system and has been designed as a highly scalable, highly extensible platform-agnostic solution for [metadata management](https://en.wikipedia.org/wiki/Metadata_management). It consists of the following system components:
```
alexlingww: Metadata management was brought up above in the Features list, maybe link it up there instead of here?

whugoh:
Looks great, @wslulciuc.

One aspect the doc is light on from my perspective is the metadata storage implementation. In particular, we currently have a PostgreSQL dependency, which I suppose is not strictly needed. Some discussion of it would be good.

Not a blocker for a first version though IMO.

docs/index.md
wslulciuc: Agreed! I'll link the definition when the _metadata management_ concept is first introduced.

docs/index.md
wslulciuc: Great feedback, @alexlingww!

> Seems to me there needs to be a merger of this section and the tiny section below. The points in the below section seem to map to some of the points in this section.

The `Features` section is actually a subsection of `Why Marquez?`. In the current layout, it might not be immediately obvious that the two sections are linked. I actually feel the `Features` section is a more appropriate subsection of `Overview`.

> Maybe even going as far as making diagrams for this section too?

Added a diagram illustrating a sample lineage query. More details can be added in future iterations.

wslulciuc:
@whugoh, very much agree! We should also provided a deep dive into the  _schema_ of the metadata repository. But, these are details that can be added in future iterations of the doc. I opened:

- Add overview of metadata storage implementation #471
- Add overview of metadata repository schema #470

Thanks again for your feedback 💯

docs/index.md
wslulciuc: > I guess my confusion is that data lineage and data provenance seem like the same thing?

@alexlingww, exactly! I use the terms data lineage and data provenance interchangeable. In other words:

> Data lineage includes the concept of an origin for the data—its original source or provenance—and the movement and change of the data as it passes through systems and is adopted for different uses (the sequence of steps within the data chain through which data has passed).

source: https://www.sciencedirect.com/topics/computer-science/data-lineage

Issue #448, opened 2019-04-12
Test dataset idempotence at the DAO layer


ashulmanWeWork:
Need to codify the behavior using tests.

Issue #447, opened 2019-04-12
Add empty result tests for get and list for dataset API


ashulmanWeWork:
In the DAO and service layers we need to testing to make sure that empty results are handled properly.

PR #459, opened 2019-04-18
Testing/add more dataset dao tests

### LINKED ISSUE #446: Follow-on Testing Work for Dataset APIs
There were a few test cases for Datasets that we need to add - generally speaking the cases are where findBy does not return anything, where list returns an empty set, and the idempotency properties of both dataset and namespace. Example test cases include (but not limited to):
- List datasets when there are no datasets (empty list)
- In the dataset dao, test the double insert and verify idempotence
- In the dataset dao, test the case where findBy does not find anything
- Test how mappers handle `updatedAt` 
- Verify the behavior when the namespace service returns an empty set

**Subtasks**
- [ ] Add empty result tests for get and list for dataset API  #447
- [ ] Test dataset idempotence at the DAO layer #448

ashulmanWeWork:
Fixes #446

### DIFF
--- a/src/main/java/marquez/db/DatasetDao.java
+++ b/src/main/java/marquez/db/DatasetDao.java
@@ -69,7 +69,7 @@ public interface DatasetDao {
   @SqlUpdate(
       "UPDATE datasets "
           + "SET updated_at = NOW(), "
-          + "current_version_uuid = :currentVersionUuid "
+          + "    current_version_uuid = :currentVersionUuid "
           + "WHERE guid = :uuid")
   void updateCurrentVersionUuid(UUID uuid, UUID currentVersionUuid);
 
--- a/src/main/java/marquez/db/DatasetDao.java
+++ b/src/main/java/marquez/db/DatasetDao.java
@@ -41,8 +41,7 @@ public interface DatasetDao {
 
   @SqlUpdate(
       "INSERT INTO datasets (guid, namespace_guid, datasource_uuid, urn, description, name) "
-          + "VALUES (:uuid, :namespaceUuid, :datasourceUuid, :urn, :description, :name) "
-          + "ON CONFLICT (urn) DO NOTHING")
+          + "VALUES (:uuid, :namespaceUuid, :datasourceUuid, :urn, :description, :name) ")
   void insert(@BindBean DatasetRow datasetRow);
 
   @SqlQuery(
--- a/src/main/java/marquez/db/DatasetDao.java
+++ b/src/main/java/marquez/db/DatasetDao.java
@@ -41,7 +41,7 @@ public interface DatasetDao {
 
   @SqlUpdate(
       "INSERT INTO datasets (guid, namespace_guid, datasource_uuid, urn, description, name) "
-          + "VALUES (:uuid, :namespaceUuid, :datasourceUuid, :urn, :description, :name) ")
+          + "VALUES (:uuid, :namespaceUuid, :datasourceUuid, :urn, :description, :name)")
   void insert(@BindBean DatasetRow datasetRow);
 
   @SqlQuery(
@@ -69,7 +69,7 @@ public interface DatasetDao {
   @SqlUpdate(
       "UPDATE datasets "
           + "SET updated_at = NOW(), "
-          + "    current_version_uuid = :currentVersionUuid "
+          + "current_version_uuid = :currentVersionUuid "
           + "WHERE guid = :uuid")
   void updateCurrentVersionUuid(UUID uuid, UUID currentVersionUuid);
 
--- a/src/main/resources/db/migration/V16__set_default_dataset_updated_at.sql
+++ b/src/main/resources/db/migration/V16__set_default_dataset_updated_at.sql
@@ -1,0 +1,1 @@
+ALTER TABLE datasets ALTER COLUMN updated_at SET DEFAULT current_timestamp;
--- a/src/test/java/marquez/api/resources/NamespaceResourceTest.java
+++ b/src/test/java/marquez/api/resources/NamespaceResourceTest.java
@@ -103,6 +103,19 @@ public class NamespaceResourceTest extends NamespaceBaseTest {
   }
 
   @Test
+  public void testListNamespaceWithNoResults() throws MarquezServiceException {
+    final List<Namespace> existingCoreModelNamespaces = Collections.emptyList();
+    NamespaceService namespaceService = mock(NamespaceService.class);
+    NamespaceResource namespaceResource = new NamespaceResource(namespaceService);
+    when(namespaceService.getAll()).thenReturn(existingCoreModelNamespaces);
+
+    Response res = namespaceResource.list();
+    NamespacesResponse responseBody = (NamespacesResponse) res.getEntity();
+
+    assertThat(responseBody.getNamespaces()).isEmpty();
+  }
+
+  @Test
   public void testListNamespaceWithSingleResultSet() throws MarquezServiceException {
     final List<Namespace> existingCoreModelNamespaces = Collections.singletonList(TEST_NAMESPACE);
     NamespaceService namespaceService = mock(NamespaceService.class);
--- a/src/test/java/marquez/db/DatasetDaoTest.java
+++ b/src/test/java/marquez/db/DatasetDaoTest.java
@@ -21,6 +21,7 @@ import static marquez.db.models.DbModelGenerator.newDatasourceRow;
 import static marquez.db.models.DbModelGenerator.newNamespaceRow;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.time.Instant;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
@@ -32,12 +33,12 @@ import marquez.db.models.DatasetRow;
 import marquez.db.models.DatasourceRow;
 import marquez.db.models.NamespaceRow;
 import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 import org.jdbi.v3.testing.JdbiRule;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
@@ -99,7 +100,7 @@ public class DatasetDaoTest {
     assertThat(rowsAfter).isEqualTo(rowsBefore + 1);
   }
 
-  @Test
+  @Test(expected = UnableToExecuteStatementException.class)
   public void testIdempotentInsert() {
     final int rowsBefore = datasetDao.count();
 
@@ -111,12 +112,9 @@ public class DatasetDaoTest {
     assertThat(rowsAfter).isEqualTo(rowsBefore + 1);
 
     datasetDao.insert(newDatasetRow);
-    final int rowsAfterIdempotentInsert = datasetDao.count();
-    assertThat(rowsAfter).isEqualTo(rowsAfterIdempotentInsert);
   }
 
-  @Ignore("Ignore for now while we try to reach parity betwen insert and insertAndGet")
-  @Test
+  @Test(expected = UnableToExecuteStatementException.class)
   public void testIdempotentInsertAndGet() {
     final int rowsBefore = datasetDao.count();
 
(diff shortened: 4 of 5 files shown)


ashulmanWeWork:
Fixes #447 
Fixes #448

src/main/resources/db/migration/V16__set_default_dataset_updated_at.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE datasets ALTER COLUMN updated_at SET DEFAULT current_timestamp;
```
wslulciuc: We should only set `updated_at` when the row has been modified. Meaning, a `null` value for `updated_at` in `datasets` signifies the row has been _inserted_, but not yet _updated_. Why change this behavior?

src/test/java/marquez/api/resources/NamespaceResourceTest.java
```diff
@@ -102,6 +102,19 @@ public void testGetNamespaceRequestErrorHandling() throws Throwable {
...
+    NamespaceService namespaceService = mock(NamespaceService.class);
+    NamespaceResource namespaceResource = new NamespaceResource(namespaceService);
+    when(namespaceService.getAll()).thenReturn(existingCoreModelNamespaces);
+
+    Response res = namespaceResource.list();
+    NamespacesResponse responseBody = (NamespacesResponse) res.getEntity();
+
+    assertThat(responseBody.getNamespaces()).isEmpty();
+  }
+
```
wslulciuc: Minor: We'll need to revisit the test suite for `NamespaceResource`, so would have been nice to see this test added then vs this PR on `DatasetDao`

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -61,13 +66,31 @@ public static void setUpOnce() {
     datasetDao = jdbi.onDemand(DatasetDao.class);
   }
 
+  @Before
+  public void setup() {
```
wslulciuc: Not every tests requires rows to be inserted. I think if that was the case, a setup method like this one would make a lot of sense. To avoid confusion and created test rows unnecessarily, I'd move these inserts into their respective tests.

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -61,13 +66,31 @@ public static void setUpOnce() {
...
+    datasourceRow = newDatasourceRow();
+    namespaceRow = newNamespaceRow();
+
+    namespaceDao.insert(namespaceRow);
+    datasourceDao.insert(datasourceRow);
+  }
+
+  @Test
+  public void testGetEmptySetOfRowsByUUID() {
```
wslulciuc: _Minor_: `testFindBy_uuid_emptyRows`. And a more natural flow, for readability, would be to move this test below `testFindBy_uuid`

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -61,13 +66,31 @@ public static void setUpOnce() {
...
+    Optional<DatasetRow> returnedRow = datasetDao.findBy(UUID.randomUUID());
+    assertThat(returnedRow).isEmpty();
```
wslulciuc: We can shorten this into a single line:

```java
assertThat(datasetDao.findBy(UUID.randomUUID())).isNotPresent()

```

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -61,13 +66,31 @@ public static void setUpOnce() {
...
+  }
+
+  @Test
+  public void testFindAllWithEmptySetOfRows() {
```
wslulciuc: _Minor_: `testFindAll_noRowsFound`. And a more natural flow, for readability, would be to move this test below `testFindAll`

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -61,13 +66,31 @@ public static void setUpOnce() {
...
+    List<DatasetRow> returnedRow =
```
wslulciuc: We can shorten this into a single line:

```java
assertThat(datasetDao.findAll(newNamespaceName(), LIMIT, 0)).isNotPresent()
```

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -77,13 +100,36 @@ public void testInsert() {
     assertThat(rowsAfter).isEqualTo(rowsBefore + 1);
   }
 
+  @Test(expected = UnableToExecuteStatementException.class)
+  public void testIdempotentInsert() {
```
wslulciuc: _Minor_: A better test name would be `testInsert_throwsException_onDuplicateRow`

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -77,13 +100,36 @@ public void testInsert() {
...
+    final int rowsBefore = datasetDao.count();
+
+    final DatasetRow newDatasetRow =
+        newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid());
+    datasetDao.insert(newDatasetRow);
+
+    final int rowsAfter = datasetDao.count();
+    assertThat(rowsAfter).isEqualTo(rowsBefore + 1);
+
+    datasetDao.insert(newDatasetRow);
```
wslulciuc:
```java
assertThatExceptionOfType(UnableToExecuteStatementException.class)
    .isThrownBy(() -> datasetDao.insert(newDatasetRow));
```

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -77,13 +100,36 @@ public void testInsert() {
...
+  }
+
+  @Test(expected = UnableToExecuteStatementException.class)
+  public void testIdempotentInsertAndGet() {
+    final int rowsBefore = datasetDao.count();
+
+    final DatasetRow newDatasetRow =
+        newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid());
+    datasetDao.insert(newDatasetRow);
```
wslulciuc: You're testing `DatasetDao.insert()`. Is this test even necessary for `DatasetDao.insertAndGet()`?

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -93,51 +139,38 @@ public void testInsertAndGet() {
...
-    final NamespaceRow namespaceRow = newNamespaceRow();
-    final DatasourceRow datasourceRow = newDatasourceRow();
-    namespaceDao.insert(namespaceRow);
-    datasourceDao.insert(datasourceRow);
-
     final DatasetRow newDatasetRow =
         newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid());
     final DatasetRow datasetRow = datasetDao.insertAndGet(newDatasetRow).orElse(null);
     assertThat(datasetRow.getCurrentVersionUuid()).isNull();
+    final Instant previousUpdatedAt = datasetRow.getUpdatedAt();
```
wslulciuc: Why would a new row have `UpdatedAt` set?

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -93,51 +139,38 @@ public void testInsertAndGet() {
...
 
     final UUID currentVersionUuid = UUID.randomUUID();
     datasetDao.updateCurrentVersionUuid(datasetRow.getUuid(), currentVersionUuid);
 
     final DatasetRow datasetRowWithVersion = datasetDao.findBy(datasetRow.getUuid()).orElse(null);
     assertThat(datasetRowWithVersion.getCurrentVersionUuid()).isNotNull();
     assertThat(datasetRowWithVersion.getCurrentVersionUuid()).isEqualTo(currentVersionUuid);
+    assertThat(datasetRowWithVersion.getUpdatedAt()).isAfter(previousUpdatedAt);
```
wslulciuc: Not sure the variable adds much, would be simpler to assert:

```java
assertThat(datasetRowWithVersion.getUpdatedAt()).isAfter(datasetRow.getUpdatedAt());
```

src/test/java/marquez/api/resources/NamespaceResourceTest.java
ashulmanWeWork: Yea, this was something of a non-sequitur. It got bundled into the issue b/c I think it somehow came up in the code review. But yea, better if it was part of a different PR.

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: true - fixed.

src/test/java/marquez/service/DatasetServiceTest.java
```diff
@@ -232,7 +232,16 @@ public void testGet() throws MarquezServiceException {
   }
 
   @Test
-  public void testGet_throwsException_onDbError() throws MarquezServiceException {
+  public void testGetWithNoResults() throws MarquezServiceException {
+    when(datasetDao.findBy(DATASET_URN)).thenReturn(Optional.empty());
+
+    final Dataset dataset = datasetService.get(DATASET_URN).orElse(null);
```
wslulciuc: assertThat(datasetService.get(DATASET_URN)).isNotPresent()

src/test/java/marquez/service/DatasetServiceTest.java
```diff
@@ -232,7 +232,16 @@ public void testGet() throws MarquezServiceException {
   }
 
   @Test
-  public void testGet_throwsException_onDbError() throws MarquezServiceException {
+  public void testGetWithNoResults() throws MarquezServiceException {
```
wslulciuc: Minor: `testGet_notPresent`

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: This is part of the setup and not what's under test. Since we're not validating the data that's inserted, there's no purpose to get the data. Seems confusing to `get` anything and then not store it.

That said, this test is necessary because `insert` and `insertAndGet` follow different code paths and issue different queries. It was that divergent behavior (one had `ON CONFLICT DO NOTHING` and one did not) that led me to write this test case in the first place. While it's true that similar now, but if they diverged in an erroneous way, having a test for each would help detect that, where just one test could miss it.

wslulciuc: Make sense 👍

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -77,13 +100,36 @@ public void testInsert() {
...
+    final int rowsBefore = datasetDao.count();
+
+    final DatasetRow newDatasetRow =
+        newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid());
+    datasetDao.insert(newDatasetRow);
+
+    final int rowsAfter = datasetDao.count();
+    assertThat(rowsAfter).isEqualTo(rowsBefore + 1);
+
+    datasetDao.insertAndGet(newDatasetRow);
```
wslulciuc:
```java
assertThatExceptionOfType(UnableToExecuteStatementException.class)
    .isThrownBy(() -> datasetDao.insertAndGet(newDatasetRow));
```

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: Yup :) Fixed.

src/test/java/marquez/db/DatasetDaoTest.java
wslulciuc: Are you expecting `datasetDao.insertAndGet()` to throw an exception?

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: Nice - I like this approach more b/c it makes the test more precise by indicating just where it's supposed to throw the exception.

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -41,8 +41,7 @@
 
   @SqlUpdate(
       "INSERT INTO datasets (guid, namespace_guid, datasource_uuid, urn, description, name) "
-          + "VALUES (:uuid, :namespaceUuid, :datasourceUuid, :urn, :description, :name) "
-          + "ON CONFLICT (urn) DO NOTHING")
+          + "VALUES (:uuid, :namespaceUuid, :datasourceUuid, :urn, :description, :name) ")
```
wslulciuc: Minor: You can remove the trailing white space in the query

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: Done

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: Done

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: Done

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: Done

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: Yup

src/test/java/marquez/db/DatasetDaoTest.java
ashulmanWeWork: There are only two test cases that don't rely on `datasourceRow` and `namespaceRow`. I've modified one of them to use the `namespaceRow` created in setup (which it should have been doing anyway), so that leaves just one.
I'm okay with one test method not needing the setup code but having it run anyway. 

The alternative would be to re-add 22 lines of duplicate code, which would take away from the readability and break the DRY principle, since we're trying to accomplish the same setup each time.

src/main/java/marquez/db/DatasetDao.java
ashulmanWeWork: Fixed.

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -70,7 +69,7 @@ default void insertAll(
   @SqlUpdate(
       "UPDATE datasets "
           + "SET updated_at = NOW(), "
-          + "    current_version_uuid = :currentVersionUuid "
```
wslulciuc: The space here was added for alignment and readability. Why remove it?

ashulmanWeWork: Ohh ok, looked a little unnatural to me, but I can move it back.

src/test/java/marquez/api/resources/NamespaceResourceTest.java
```diff
@@ -102,6 +102,19 @@ public void testGetNamespaceRequestErrorHandling() throws Throwable {
...
   }
 
+  @Test
+  public void testListNamespaceWithNoResults() throws MarquezServiceException {
+    final List<Namespace> existingCoreModelNamespaces = Collections.emptyList();
+    NamespaceService namespaceService = mock(NamespaceService.class);
+    NamespaceResource namespaceResource = new NamespaceResource(namespaceService);
+    when(namespaceService.getAll()).thenReturn(existingCoreModelNamespaces);
+
+    Response res = namespaceResource.list();
```
ashulmanWeWork: todo: One the next pass, let's make what can be final, final

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -61,13 +66,19 @@ public static void setUpOnce() {
     datasetDao = jdbi.onDemand(DatasetDao.class);
   }
 
+  @Before
+  public void setup() {
```
ashulmanWeWork: `setUp` ... will fix

PR #461, opened 2019-04-19
Support dashes (-) in the namespaceName

roaraya8:


src/test/java/marquez/common/models/NamespaceNameTest.java
```diff
@@ -33,6 +33,12 @@ public void testNewNamespace() {
     assertEquals(value, NamespaceName.fromString(value).getValue());
   }
 
+  @Test
```
ashulmanWeWork: I don't anticipate issues, but please also add a resource-level test.

roaraya8: I just modified the namespace name use in the NameSpaceResourceTest to make sure.

ashulmanWeWork: That works too! Thanks :)

src/main/java/marquez/common/models/NamespaceName.java
```diff
@@ -30,15 +30,15 @@
...
+          String.format("^[a-zA-Z0-9_-]{%d,%d}$", NAMESPACE_MIN_SIZE, NAMESPACE_MAX_SIZE));
 
   @Getter private final String value;
 
   private NamespaceName(@NonNull final String value) {
     checkArgument(
         NAMESPACE_PATTERN.matcher(value).matches(),
         "A namespaces must contain only letters (a-z, A-Z), numbers (0-9), or "
-            + "underscores (_) with a maximum length of 1024 characters.");
+            + "underscores (_) and dashes (-) with a maximum length of 1024 characters.");
```
wslulciuc: _very_ _very_ _minor_: we'll want the wording of the error to be:

> A namespaces must contain only letters (a-z, A-Z), numbers (0-9), underscores (_) or dashes (-) with a maximum length of 1024 characters.

src/test/java/marquez/common/models/NamespaceNameTest.java
```diff
@@ -33,6 +33,12 @@ public void testNewNamespace() {
...
+  public void testNewNamespace_withDashesAndUnderscore() {
+    final String nameWithDashesAndUnderscore = "test_a-b_c";
```
wslulciuc: 💯

Issue #183, opened 2018-12-06
Migrate from java.sql.Timestamp / java.util.Date to java.util.Instant


sshah-wework:
java.util.Date is apparently [no bueno](https://codeblog.jonskeet.uk/2017/04/23/all-about-java-util-date/).

sshah-wework:
We seem to use a few different date formats. In some cases it's a long representing the UNIX timestamp, in other cases it's a Timestamp instance. All should be switched over to Instant.

wslulciuc:
Fixed #484

PR #484, opened 2019-05-08
Return timestamp strings as ISO format

### LINKED ISSUE #189: Remove timestamp from the API models and replace with String type

wslulciuc:
This PR introduces the following changes:

1. Sets server timezone to UTC using the JVM argument `user.timezone`
2. Drops timezones from timestamps stored in DB
3. Formats all API timesamps as ISO format
4. ~Renames `db_table_infos` to `db_table_info`~

Closes #189 #183


src/test/java/marquez/db/mappers/DbTableVersionRowMapperTest.java
```diff
@@ -40,7 +40,10 @@
 
   @Test
   public void testMap() throws SQLException {
+    final Object exists = mock(Object.class);
     final ResultSet results = mock(ResultSet.class);
+    when(results.getObject(Columns.CREATED_AT)).thenReturn(exists);
```
ashulmanWeWork: How does this reconcile with line 48?
```
when(results.getTimestamp(Columns.CREATED_AT)).thenReturn(Timestamp.from(CREATED_AT));
```

wslulciuc: Sure, see [`Columns.timestampOrThrow()`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/db/Columns.java#L93)

src/main/resources/db/migration/V19__add_created_at_column_to_db_info_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE db_table_infos RENAME TO db_table_info;
```
sshah-wework: this seems unrelated -- does the `DbTableInfoDao` have to change now? Also seems like we don't have `DbTableInfoDao` tests to catch this. Filed #485 

The DAO method below might need to be updated?
```
  @SqlUpdate(
      "INSERT INTO db_table_infos (uuid, db_name, db_schema_name) VALUES (:uuid, :db, :dbSchema)")
  void insert(@BindBean DbTableInfoRow dbTableInfoRow);
```

src/main/resources/db/migration/V20__add_update_at_column_to_job_runs_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE job_runs ADD COLUMN updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
```
sshah-wework: needs to be `NOT NULL`able?

src/main/resources/db/migration/V19__add_created_at_column_to_db_info_table.sql
wslulciuc: Sure, let's update the table name when we fix #485. Thanks for filling the issue, @sshah-wework!

src/main/resources/db/migration/V20__add_update_at_column_to_job_runs_table.sql
wslulciuc: Yes!

PR #486, opened 2019-05-09
Set timestamp strings as ISO format in openAPI spec

wslulciuc:
This PR ensures all timestamp strings are ISO format and removes `updatedAt` from the example Job API call in the spec as it's currently not being returned.


docs/openapi.yml
```diff
@@ -449,7 +449,7 @@ components:
           type: string
       example:
         name: wework
-        createdAt: 2018-10-04T15:01:23.045Z
+        createdAt: 2019-05-09T00:24:57.609125Z
```
ashulmanWeWork: Stop working past midnight :)!

wslulciuc: Actually, I should provided better timestamps here. Also, will try not to :)

PR #488, opened 2019-05-09
Add class api.mappers.NamespaceMapper

wslulciuc:


src/test/java/marquez/api/mappers/NamespaceMapperTest.java
```diff
@@ -0,0 +1,67 @@
...
+import marquez.api.models.NamespaceRequest;
+import marquez.common.models.NamespaceName;
+import marquez.service.models.Namespace;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(UnitTests.class)
+public class NamespaceMapperTest {
+  private final NamespaceName NAMESPACE_NAME = newNamespaceName();
+  private final NamespaceRequest REQUEST = newNamespaceRequest();
```
ashulmanWeWork: For performance reasons, these can be static. JUnit creates a new instance of the class for every test method, so each time, we'd need to instantiate these. By making it static, we avoid those extra instantiations.
```
Every time we run a JUnit test class, a new instance of it is created. There is even a new instance before each test method execution. JUnit framework provides following basic lifecycle annotations (all are method level annotations).
```
source: https://www.logicbig.com/tutorials/unit-testing/junit/lifecycle.html

wslulciuc: They should be defined as constants, just forgot to make them static. Nice catch!

Issue #508, opened 2019-05-14
DatasetServiceTest should be adjusted to instantiate test mocks just once


ashulmanWeWork:
In `DatasetServiceTest` the mock DAOs are instantiated once per test method. To improve test execution time, this can be change to only occur once at the class level, by making the member variables `private static final` with inline initialization.

We can then adjust the per-method setup code to just reset the mocks.

The `DatasetService` member can also be can be instantiated once per class as well, which allows for making it `private static final`.

Please see `DatasourceServiceTest` for an example of how this can be done.

ashulmanWeWork:
It can also be argued that having each test instantiate its own mocks is better for clarity and to avoid any potential interactions, which is more important in a test. Closing this issue.

PR #509, opened 2019-05-14
Make test fixtures in JobServiceTest and DatasetServiceTest static to…

ashulmanWeWork:
… avoid re-instantiation


ashulmanWeWork:
Fixes #508

PR #513, opened 2019-05-14
Handle Flyway.repair() error

wslulciuc:


src/main/java/marquez/MarquezApp.java
```diff
@@ -102,13 +102,20 @@ private void migrateDbOrError(@NonNull MarquezConfig config) {
...
+        log.info("Repairing failed database migration...");
+        flyway.repair();
+        log.info("Successfully repaired database.");
+      } catch (FlywayException errorOnDbRepair) {
+        log.error("Failed to apply repair to database.", errorOnDbRepair);
+      }
+
+      log.info("Stopping app...");
       // The throwable is not propagating up the stack.
       onFatalError(); // Signal app termination.
```
ashulmanWeWork: I think a few unit tests would be good to verify error behavior and what is called when specific exceptions are thrown.

PR #517, opened 2019-05-16
 Change datasetdao.findall() to order by dataset.name 

soumyasmruti:


src/test/java/marquez/db/models/DbModelGenerator.java
```diff
@@ -15,14 +15,7 @@
...
 import static java.util.stream.Collectors.toList;
-import static marquez.common.models.CommonModelGenerator.newConnectionUrl;
-import static marquez.common.models.CommonModelGenerator.newDatasetName;
-import static marquez.common.models.CommonModelGenerator.newDatasetUrn;
-import static marquez.common.models.CommonModelGenerator.newDatasourceName;
-import static marquez.common.models.CommonModelGenerator.newDatasourceUrn;
-import static marquez.common.models.CommonModelGenerator.newDescription;
-import static marquez.common.models.CommonModelGenerator.newNamespaceName;
-import static marquez.common.models.CommonModelGenerator.newOwnerName;
+import static marquez.common.models.CommonModelGenerator.*;
```
ashulmanWeWork: @wslulciuc, what's our recommended best practice on star imports? I've seen it where if there are more than 6 imports from a library, we default to `*`, for example, but it varies team by team.

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -221,8 +232,18 @@ public void testFindAll_limitOnly() {
 
   @Test
   public void testFindAll_offsetOnly() {
-    final List<DatasetRow> newDatasetRows =
-        newDatasetRowsWith(namespaceRow.getUuid(), datasourceRow.getUuid(), 4);
+    List<DatasetRow> newDatasetRows = new ArrayList<>();
```
ashulmanWeWork: nit: this can be made final.

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -221,8 +232,18 @@ public void testFindAll_limitOnly() {
...
+
+    Stream.<String>builder()
+        .add("a_test_dataset")
```
ashulmanWeWork: Instead of hard-coding these, it's recommended to use `CommonModelGenerator.newDatasetName()`.

ashulmanWeWork: In the generation code we have, to generate a list of n of something, the pattern is `new<Object>s`. In this situation, I don't see a `newDatasetNames(limit n)`, so if you have a minute, I'd recommend adding it.

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -221,8 +232,18 @@ public void testFindAll_limitOnly() {
...
+        .add("b_test_dataset")
+        .add("c_test_dataset")
+        .add("d_test_dataset")
+        .build()
+        .forEach(
+            name ->
+                newDatasetRows.add(
+                    newDatasetRowWith(namespaceRow.getUuid(), datasourceRow.getUuid(), name)));
```
ashulmanWeWork: Actually, I realized that naming (specifically the dataset name) is highly important to what you're looking to demonstrate in the test, which is why you opted to set them yourself :)

src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -207,8 +209,17 @@ public void testFindAll_noRowsFound() {
 
   @Test
   public void testFindAll_limitOnly() {
-    final List<DatasetRow> newDatasetRows =
-        newDatasetRowsWith(namespaceRow.getUuid(), datasourceRow.getUuid(), 4);
+    List<DatasetRow> newDatasetRows = new ArrayList<>();
```
ashulmanWeWork: Can we add a test case where we insert datasets in a non-sorted order and verify that they come back sorted?

soumyasmruti: Sure

soumyasmruti: I took care of this case in both the test cases which inserts the randomly generated list first and then sorts it and compares.

src/test/java/marquez/db/DatasetDaoTest.java
soumyasmruti: Went with the sorting approach. Looks clean now.

src/test/java/marquez/db/DatasetDaoTest.java
soumyasmruti: Done

src/test/java/marquez/db/models/DbModelGenerator.java
soumyasmruti: Removed *. This was generate by default from the IDE. Can I know the default number so that I can set it in my IDE.

PR #528, opened 2019-05-20
Allow dash ('-') in URN fixes #476

### LINKED ISSUE #476: Allow dash ('-') in URN
Let's allow dashes (**`-`**) in [`UrnPattern`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/common/models/UrnPattern.java#L41)

overlinden:
Allow dash ('-') in URN
Fixes #476


overlinden:
@wslulciuc: Thank you. I just added the testcase to the commit. Doing that I also updated the exception text to reflect current regex pattern.
Looking forward for merge now.

src/main/java/marquez/common/models/UrnPattern.java
```diff
@@ -48,7 +48,7 @@ void throwIfNoMatch(@NonNull String value) {
       throw new IllegalArgumentException(
           String.format(
               "urn (%s) must contain only letters (a-z, A-Z), numbers (0-9), periods (.), "
-                  + "or underscores (_) and be sperated by colons (:) with each part "
+                  + " underscores (_) or dashes (-) and be sperated by colons (:) with each part "
```
wslulciuc: _Minor_: Great to see that you updated the description as well! You'll want to remove the extra spacing:

```
" underscores (_) or dashes (-) ...",  should be "underscores (_) or dashes (-) ..."
```

wslulciuc:
@overlinden Awesome! Looks like CI is failing formatting though. Easy fix, just run:

```
$ ./gradlew spotlessApply
```

Then commit the formatting changes.

src/main/java/marquez/common/models/UrnPattern.java
overlinden: done

overlinden:
@wslulciuc Done. No everything should be fine. ;)

wslulciuc:
@overlinden Looks great, but CI is still failing because of formatting violation. Mind running the following cmd (as described above):

```
$ ./gradlew spotlessApply
```

Once fixed, we can land these sweet changes into master!

overlinden:
@wslulciuc I get struggled with the formatting. The line endings seem to be the problem. After running `$ ./gradlew spotlessApply` the whole file is replaced in the diff and still the format seems to be wrong. Do you have any advise what I can do to fix this?

wslulciuc:
Interesting! Just to make sure we handle formatting issues the same way, mind following these steps?

1. `./gradlew spotlessApply`
2. `git commit -am "Apply formatting"`
3. `git push origin [feature-branch]`

Rather than doing a _force push_, let's apply formatting in a separate commit. That should fix things (hopefully).

overlinden:
@wslulciuc Problem was the `core.autocrlf` setting in git. I manually set this to `input` and it seems to work now.

Issue #490, opened 2019-05-10
JdbiRule.embeddedPostgres().migrateWithFlyway() deprecated


wslulciuc:
Jdbi has deprecated [`migrateWithFlyway()`](http://jdbi.org/apidocs/org/jdbi/v3/testing/JdbiRule.html#migrateWithFlyway--). We should use [`withMigration(Migration)`](http://jdbi.org/apidocs/org/jdbi/v3/testing/JdbiRule.html#withMigration-org.jdbi.v3.testing.Migration-) instead.

ravikamaraj:
Hi @wslulciuc 
I can pick this task, can you assign it to me. thanks

wslulciuc:
@ravikamaraj sure and thanks!

wslulciuc:
Fixed: #531

PR #531, opened 2019-05-20
Resolves Issue #490

ravikamaraj:
@wslulciuc 
Can you review, thanks


wslulciuc:
@ravikamaraj, great work! I'm wondering, can we get away with initializing `Migration` inline? That is,

```java
JdbiRule.embeddedPostgres()
  .withPlugin(new SqlObjectPlugin())
  .withMigration(Migration.before().withDefaultPath());
```

Since we're created a new embedded postgres instance on each DAO test suite, we can probably get away with allowing the teardown of the DB objects to be handled transparently behind the scenes. Thoughts?

ravikamaraj:
@wslulciuc, sure, yes that makes sense

ravikamaraj:
@wslulciuc, thanks for the feedback. I have updated the code, can you review

PR #538, opened 2019-05-26
Issue 533: Update test classes on Mock usage

ravikamaraj:
@wslulciuc 
Mock usage optimization was applied to all the test classes. For below test classes, where it had mock usage at @ClassRule level, simplified mock usage was not applied, as it required to change object scopes. Kindly review, and let me know your feedback, thanks

- JobResourceTest
- NamespaceResourceTest
- DatasourceResourceTest

### DIFF
--- a/src/test/java/marquez/api/resources/NamespaceResourceTest.java
+++ b/src/test/java/marquez/api/resources/NamespaceResourceTest.java
@@ -41,11 +41,21 @@ import marquez.service.exceptions.MarquezServiceException;
 import marquez.service.models.Namespace;
 import org.junit.Before;
 import org.junit.ClassRule;
+import org.junit.Rule;
 import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
 
 public class NamespaceResourceTest extends NamespaceBaseTest {
 
-  NamespaceName namespaceName = NamespaceName.fromString(NAMESPACE_NAME);
+  @Rule public MockitoRule rule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
+
+  @Mock private NamespaceService namespaceService;
+
+  private NamespaceName namespaceName = NamespaceName.fromString(NAMESPACE_NAME);
+
   private static final NamespaceService NAMESPACE_SERVICE = mock(NamespaceService.class);
 
   @ClassRule
@@ -78,7 +88,7 @@ public class NamespaceResourceTest extends NamespaceBaseTest {
   @Test
   public void testValidNamespace() throws MarquezServiceException {
     Optional<Namespace> returnedOptionalNamespace = Optional.of(TEST_NAMESPACE);
-    NamespaceService namespaceService = mock(NamespaceService.class);
+
     NamespaceResource namespaceResource = new NamespaceResource(namespaceService);
 
     when(namespaceService.get(namespaceName)).thenReturn(returnedOptionalNamespace);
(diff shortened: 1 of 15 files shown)


src/test/java/marquez/api/resources/NamespaceResourceTest.java
```diff
@@ -41,11 +41,21 @@
...
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
 
 public class NamespaceResourceTest extends NamespaceBaseTest {
 
-  NamespaceName namespaceName = NamespaceName.fromString(NAMESPACE_NAME);
+  @Rule public MockitoRule rule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
+
+  @Mock private NamespaceService namespaceService;
```
wslulciuc: We can remove the usage of `NAMESPACE_SERVICE` and replace all references with the instance created with `@Mock`

src/test/java/marquez/db/ColumnsTest.java
```diff
@@ -27,16 +26,25 @@
...
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
 
 @Category(UnitTests.class)
 public class ColumnsTest {
+  @Rule public MockitoRule rule = MockitoJUnit.rule().strictness(Strictness.WARN);
```
wslulciuc: Should we be using `Strictness.STRICT_STUBS` instead?

src/test/java/marquez/service/DatasourceServiceTest.java
```diff
@@ -33,24 +31,23 @@
...
-    datasourceService = new DatasourceService(datasourceDao);
-  }
+  @Mock private DatasourceDao datasourceDao;
 
-  @After
-  public void tearDown() {
-    reset(datasourceDao);
+  @Before
+  public void setUp() {
+    datasourceService = new DatasourceService(datasourceDao);
```
wslulciuc: 👍

src/test/java/marquez/api/resources/NamespaceResourceTest.java
wslulciuc: Nvd, you commented on this case in your PR description 👍

wslulciuc:
Overall, this looks great @ravikamaraj! Left a minor comment, but we can merge once you provide some clarification on mock strictness.

src/test/java/marquez/db/ColumnsTest.java
ravikamaraj: @wslulciuc 
In each of the test cases, different values has been used for column argument in ResultSet mock method, so had to choose Strictness.WARN. If we had to use Strictness.STRICT_STUBS, then each of the test cases has to be updated to use Mockito.anyString() as argument in ResultSet mock methods. Please let me know, how do we want to handle this.

Note: Strictness.WARN has also been used in below test classes, due to above reason
NamespaceRowMapperTest
JobServiceTest

wslulciuc: Great, thanks for clarifying @ravikamaraj! Looks like `Strictness.WARN` has been the default behavior since Mockito [2.x](https://static.javadoc.io/org.mockito/mockito-core/2.27.0/org/mockito/quality/Strictness.html) allowing us to define the rule simply as:

```java
@Rule public MockitoRule rule = MockitoJUnit.rule()
```

ravikamaraj: yes, that's correct, I would update the code shortly

ravikamaraj: done, can you review

PR #540, opened 2019-05-31
Endpoint to get job runs by job

sshah-wework:
New endpoint to fetch job runs by job:
- Returns job runs in reverse chronological order
- Supports `LIMIT` and `OFFSET` for pagination

### DIFF
--- a/src/main/java/marquez/api/resources/JobResource.java
+++ b/src/main/java/marquez/api/resources/JobResource.java
@@ -150,10 +150,23 @@ public final class JobResource {
   }
 
   @Timed
-  @ResponseMetered
-  @ExceptionMetered
+  @Path("/namespaces/{namespace}/jobs/{job}/runs")
+  public Response getJobRuns(
+      @PathParam("namespace") final NamespaceName namespaceName, @PathParam("job") final String job)
+      throws MarquezServiceException {
+    if (!namespaceService.exists(namespaceName)) {
+      return Response.status(Response.Status.NOT_FOUND).entity("Namespace not found").build();
+    }
+    final Optional<List<JobRun>> jobRuns = jobService.getAllRunsOfJob(namespaceName, job);
+    if (jobRuns.isPresent()) {
+      return Response.ok().entity(coreJobRunToApiJobRunMapper.map(jobRuns.get())).build();
+    }
+    return Response.status(Response.Status.NOT_FOUND).build();
+  }
+
   @GET
   @Path("/namespaces/{namespace}/jobs/{job}/runs")
+  @Produces(APPLICATION_JSON)
   public Response getJobRuns(
       @PathParam("namespace") final NamespaceName namespaceName, @PathParam("job") final String job)
       throws MarquezServiceException {
@@ -167,6 +180,21 @@ public final class JobResource {
 
   @GET
   @Produces(APPLICATION_JSON)
+  @Path("/namespaces/{namespace}/jobs/{job}/runs")
+  public Response getJobRuns(
+      @PathParam("namespace") final NamespaceName namespaceName, @PathParam("job") final String job)
+      throws MarquezServiceException {
+    if (!namespaceService.exists(namespaceName)) {
+      return Response.status(Response.Status.NOT_FOUND).entity("Namespace not found").build();
+    }
+    final Optional<List<JobRun>> jobRuns = jobService.getAllRunsOfJob(namespaceName, job);
+    if (jobRuns.isPresent()) {
+      return Response.ok().entity(coreJobRunToApiJobRunMapper.map(jobRuns.get())).build();
+    }
+    return Response.status(Response.Status.NOT_FOUND).build();
+  }
+
+  @GET
   @Timed
   @Path("/jobs/runs/{id}")
   public Response get(@PathParam("id") final UUID runId) throws MarquezServiceException {
--- a/src/main/java/marquez/api/resources/JobResource.java
+++ b/src/main/java/marquez/api/resources/JobResource.java
@@ -153,8 +153,24 @@ public final class JobResource {
   @ResponseMetered
   @ExceptionMetered
   @GET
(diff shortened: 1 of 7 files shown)


src/main/java/marquez/api/resources/JobResource.java
```diff
@@ -149,12 +151,25 @@ public Response create(
...
+    if (!namespaceService.exists(namespaceName)) {
+      return Response.status(Response.Status.NOT_FOUND).entity("Namespace not found").build();
+    }
+    final List<JobRun> jobRuns = jobService.getAllRunsOfJob(namespaceName, job, limit, offset);
+    return Response.ok().entity(JobRunResponseMapper.map(jobRuns)).build();
+  }
+
+  @GET
+  @Timed
+  @Path("/jobs/runs/{id}")
```
wslulciuc: _Minor_: We'll want to keep the metric annotations for `GET` `/jobs/runs/{id}`. May have been removed when merging with `master`:

```java
@Timed
@ResponseMetered
@ExceptionMetered
@GET
@Path("/jobs/runs/{id}")
@Produces(APPLICATION_JSON)
public Response get(@PathParam("id") final UUID runId) throws MarquezServiceException {
  . . .
}
``

File: src/main/java/marquez/api/resources/JobResource.java [hunk 2]
```diff
@@ -149,12 +151,25 @@ public Response create(
         .build();
   }
 
-  @Timed
-  @ResponseMetered
-  @ExceptionMetered
   @GET
-  @Path("/jobs/runs/{id}")
   @Produces(APPLICATION_JSON)
+  @Path("/namespaces/{namespace}/jobs/{job}/runs")
```
wslulciuc: We'll also want to add the following metric annotations:

```java
@Timed
@ResponseMetered
@ExceptionMetered
@GET
@Path("/namespaces/{namespace}/jobs/{job}/runs")
@Produces(APPLICATION_JSON)
public Response getRunsForJob(...) {
  ...
}
```

src/main/java/marquez/db/JobRunDao.java
```diff
@@ -68,4 +69,15 @@ default void updateState(UUID jobRunID, Integer state) {
...
           + "WHERE jr.guid = :guid")
   JobRun findJobRunById(UUID guid);
+
+  @SqlQuery(
+      "SELECT jr.*, jra.args_json "
+          + "FROM job_runs jr "
+          + "INNER JOIN job_versions jv "
+          + " ON (jr.job_version_guid = jv.guid AND jv.job_guid = :jobUuid)"
+          + "LEFT JOIN job_run_args jra "
+          + " ON (jr.job_run_args_hex_digest = jra.hex_digest) "
```
wslulciuc: _Minor_: Just realized we need to update the table schemas for `job_runs` and `job_runs_args` (which drops `job_run_args_hex_digest`, etc)

- Update job_run_args schema #541
- Update job_runs schema #542

src/main/java/marquez/db/JobRunDao.java
```diff
@@ -68,4 +69,15 @@ default void updateState(UUID jobRunID, Integer state) {
...
+          + "ORDER BY created_at DESC "
+          + "LIMIT :limit OFFSET :offset")
```
wslulciuc: 👍

src/test/java/marquez/api/resources/JobResourceTest.java
```diff
@@ -322,6 +326,69 @@ public void testJobRunCompleteWithUnparseableRunId() throws Exception {
...
+
+    Response response =
+        JOB_RESOURCE.getRunsForJob(
+            NAMESPACE_NAME, "nonexistent_job", DEFAULT_LIMIT, DEFAULT_OFFSET);
+
+    assertEquals(Response.Status.NOT_FOUND.getStatusCode(), response.getStatus());
+  }
+
+  @Test
+  public void testGetAllRunsOfJob_noJobRuns() throws MarquezServiceException {
```
wslulciuc: If no job runs are found for a given job, I think I'd be more ideal to return an empty set of runs rather then a `404`?

```
{
  "runs": []
}
```

src/test/java/marquez/api/resources/JobResourceTest.java
```diff
@@ -322,6 +326,69 @@ public void testJobRunCompleteWithUnparseableRunId() throws Exception {
...
+    JobResponse job = generateApiJob();
+    when(MOCK_NAMESPACE_SERVICE.exists(NAMESPACE_NAME)).thenReturn(false);
```
wslulciuc: Also, I think we want to this mocked call to be:

```java
when(MOCK_NAMESPACE_SERVICE.exists(NAMESPACE_NAME)).thenReturn(true);
``

src/main/java/marquez/db/JobRunDao.java
sshah-wework: good reminder, thanks

src/test/java/marquez/api/resources/JobResourceTest.java
sshah-wework: fixed

src/test/java/marquez/api/resources/JobResourceTest.java
sshah-wework: typo on my part, fixed

src/main/java/marquez/api/resources/JobResource.java
sshah-wework: yah was a bad merge, should've checked the diff. fixing it

sshah-wework: fixed

src/main/java/marquez/api/resources/JobResource.java
sshah-wework: totally missed this good catch

PR #554, opened 2019-06-07
Auto updates version in quickstart.md

ronthalanki-wework:
Using the latest tag name from github in the quickstart.md, modified link as well


docs/_config.yml
```diff
@@ -7,3 +7,4 @@ theme: jekyll-theme-cayman
 plugins:
   - jekyll-seo-tag
   - jekyll-sitemap
+
```
wslulciuc: _Minor_: Looks like the newlines may have been checked in by mistake?

PR #562, opened 2019-06-16
Use CommonModelGenerator in commons tests

alulc:


src/test/java/marquez/common/models/CommonModelGenerator.java
```diff
@@ -92,8 +92,12 @@ public static DbName newDbNameWith(String value) {
   }
 
   public static ConnectionUrl newConnectionUrl() {
+    return newConnectionUrlWithDatasourceType(newDatasourceType());
+  }
+
+  public static ConnectionUrl newConnectionUrlWithDatasourceType(DatasourceType datasourceType) {
```
wslulciuc: _Minor:_ The naming convention we have for generators is `newClazzWith(Overrides...)`, where `Clazz` is the class you'd like to instantiate and `Overrides...` are any overrides you'd like to provide.

So a generator method for `ConnectionUrl` that overrides the `DatasourceType` would be:

```java
newConnectionUrlWith(DatasourceType datasourceType)
```

src/test/java/marquez/common/models/ConnectionUrlTest.java
```diff
@@ -22,7 +22,7 @@
 
 @Category(UnitTests.class)
 public class ConnectionUrlTest {
-  private static final DatasourceType DATASOURCE_TYPE = DatasourceType.POSTGRESQL;
+  private static final DatasourceType DATASOURCE_TYPE = CommonModelGenerator.newDatasourceType();
```
wslulciuc: _Minor_: I would statically import `newDatasourceType()`

```java
import static marquez.common.models.CommonModelGenerator.newDatasourceType;

private static final DatasourceType DATASOURCE_TYPE = newDatasourceType()
```

src/test/java/marquez/common/models/CommonModelGenerator.java
alulc: Ok.

src/test/java/marquez/common/models/ConnectionUrlTest.java
alulc: Will do.

Issue #19, opened 2018-08-09
API versioning / evolution?


sshah-wework:
What are everyone's thoughts regarding how to version the API? Versioning (via URI or headers) or [evolution](https://philsturgeon.uk/api/2018/05/02/api-evolution-for-rest-http-apis/)? Or something else?

wslulciuc:
+1 for URI-based versioning (e.g. `https://api.marquez.com/v1`). My preference would be to use a versioning scheme that is common and well known. Makes testing and evolving the API more predictable.

[API Versioning Has No “Right Way”](https://blog.apisyouwonthate.com/api-versioning-has-no-right-way-f3c75457c0b7) is a great blog outlining possible alternatives.

@hougs @julienledem : Any preference?

sshah-wework:
:thumbs up: URI-based feels like the most conventional approach. Have worked with an evolution approach and deprecation felt a bit more difficult as the API got more consumers.

wslulciuc:
We'll prefix all API endpoints with `/api/v1` #104

PR #566, opened 2019-06-27
Add job type along with tests

ashulmanWeWork:


src/main/java/marquez/service/models/JobType.java
```diff
@@ -0,0 +1,7 @@
+package marquez.service.models;
+
+public enum JobType {
```
wslulciuc: I would define the job type with the `Job` class along for `Job.Type.BATCH`, `Job.Type.STREAM`, and `Job.Type. SERVICE`.

src/main/java/marquez/db/mappers/JobRowMapper.java
```diff
@@ -40,6 +40,7 @@ public Job map(@NonNull ResultSet results, @NonNull StatementContext context)
         arrayOrThrow(results, Columns.INPUT_DATASET_URNS),
         arrayOrThrow(results, Columns.OUTPUT_DATASET_URNS),
         timestampOrThrow(results, Columns.CREATED_AT),
-        timestampOrThrow(results, Columns.UPDATED_AT));
+        timestampOrThrow(results, Columns.UPDATED_AT),
+        Job.Type.valueOf(results.getString(Columns.JOB_TYPE)));
```
wslulciuc: I would use `Columns.stringOrThrow()` to access the job type column.

src/main/java/marquez/service/JobService.java
```diff
@@ -44,6 +45,8 @@
   private final JobRunDao jobRunDao;
   private final JobRunArgsDao jobRunArgsDao;
 
+  private static final Job.Type DEFAULT_JOB_TYPE = Job.Type.BATCH;
```
wslulciuc: Would we want to default to a job type to `BATCH`?

wslulciuc: We should make the job type required when registered a job.

ashulmanWeWork: This was the recommendation of @sshah-wework (specifically that it be optional and default to BATCH). 
(from slack):
> we can start w/ it being optional, and it defaults to BATCH

src/main/java/marquez/db/mappers/JobRowMapper.java
ashulmanWeWork: Yea agreed - fixed.

src/main/java/marquez/db/JobDao.java
```diff
@@ -32,12 +32,12 @@
   JobVersionDao createJobVersionDao();
 
   @SqlUpdate(
-      "INSERT INTO jobs (guid, name, namespace_guid, description, input_dataset_urns, output_dataset_urns) "
-          + " VALUES (:guid, :name, :namespaceGuid, :description, :inputDatasetUrns, :outputDatasetUrns)")
-  public void insert(@BindBean Job job);
+      "INSERT INTO jobs (guid, name, namespace_guid, description, input_dataset_urns, output_dataset_urns, type) "
+          + " VALUES (:guid, :name, :namespaceGuid, :description, :inputDatasetUrns, :outputDatasetUrns, :type)")
+  void insert(@BindBean Job job);
```
ashulmanWeWork: Redundant scope modifier - interface methods are always public.

src/main/java/marquez/service/JobService.java
ashulmanWeWork: Filed #567 to adjust the behavior. Likely a Q3 task.

src/main/java/marquez/api/resources/JobResource.java
```diff
@@ -77,6 +78,11 @@ public Response create(
     if (!namespaceService.exists(namespaceName)) {
       return Response.status(Response.Status.NOT_FOUND).build();
     }
+    if (request.getType().isPresent() && !jobService.isValidType(request.getType().get())) {
+      return Response.status(HttpStatus.BAD_REQUEST_400)
+          .entity("Bad value for Job type " + request.getType().get())
```
sshah-wework: seems like this `JobType` should be a common model which could do this validation for you instead of requiring you to call into the job service for validation. Don't have to fix it here, but please address #19 afterwards.

src/main/java/marquez/db/JobDao.java
```diff
@@ -32,12 +32,12 @@
...
 
   @SqlUpdate("UPDATE jobs SET current_version_guid = :currentVersionGuid WHERE guid = :jobGuid")
-  public void setCurrentVersionGuid(UUID jobGuid, UUID currentVersionGuid);
+  void setCurrentVersionGuid(UUID jobGuid, UUID currentVersionGuid);
```
sshah-wework: why did this change? does it have something to do with job types?

src/main/resources/db/migration/V22__add_type_to_job.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TYPE type AS ENUM ('BATCH', 'SERVICE', 'STREAM');
+
+ALTER TABLE jobs
+ADD COLUMN if not exists type TYPE;
+
+ALTER TABLE jobs
+ALTER COLUMN type SET DEFAULT 'BATCH';
```
sshah-wework: we don't need to set this at the database level, the code should already be setting this.

src/main/java/marquez/db/JobDao.java
ashulmanWeWork: Oh, I left a review comment down below about this but basically it's unnecessary and the IDE gives a warning saying so. Figured it would be nice clean up little things.

src/main/java/marquez/api/resources/JobResource.java
ashulmanWeWork: Yea that's a good idea. I felt a little weird about the service doing validation for a specific type - this sounds better. Are you sure #19 is the issue though?

src/main/java/marquez/service/JobService.java
```diff
@@ -238,6 +225,48 @@ private static JobVersion createJobVersion(Job job) {
...
+    if (type == null) {
+      return false;
+    }
+    if (Enums.getIfPresent(Job.Type.class, type).orNull() != null) {
+      return true;
+    }
+    return false;
+  }
+
+  protected static Job generateNewJob(Job job) {
```
sshah-wework: the function name `generateNewJob` is really confusing

PR #592, opened 2019-07-17
Issue 584

ravikamaraj:
@wslulciuc , can you review. thanks

Also, I have noticed ApiModelGenerator::newDatasetsResponse method is not referenced/used, if this method is not going to be used, shall I go ahead and cleanup the method


ravikamaraj:
looks like old commits are pulled in, will close and open new PR

PR #593, opened 2019-07-17
namespace request/response tests

ravikamaraj:
@wslulciuc , can you review. thanks

Also, I have noticed ApiModelGenerator::newDatasetsResponse method is not referenced/used, if this method is not going to be used, shall I go ahead and cleanup the method


wslulciuc:
> Also, I have noticed ApiModelGenerator::newDatasetsResponse method is not referenced/used, if this method is not going to be used, shall I go ahead and cleanup the method

Actually, nice catch! `ApiModelGenerator.newDatasetResponses()` will never be used. Did you want to open a PR to remove the method from the generator?

ravikamaraj:
> > Also, I have noticed ApiModelGenerator::newDatasetsResponse method is not referenced/used, if this method is not going to be used, shall I go ahead and cleanup the method
> 
> Actually, nice catch! `ApiModelGenerator.newDatasetResponses()` will never be used. Did you want to open a PR to remove the method from the generator?

sure

Issue #571, opened 2019-06-28
Update OpenAPI spec for Create a Job Run


ronthalanki-wework:
The method returns a JobRun object but is documented to return a runId

ronthalanki-wework:
Closed by #595

PR #595, opened 2019-07-17
Feature/update openapi

### LINKED ISSUE #570: Update OpenAPI spec for all list methods
Include limit and offset query parameters in the OpenAPI spec for all list methods

ronthalanki-wework:
Update open api spec with limit and offset query parameters and with JobRun for the response of createJobRun
Closes #570 and #571

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -49,6 +49,9 @@ paths:
 
   /namespaces:
     get:
+      parameters:
+        - $ref: '#/components/parameters/limit'
+        - $ref: '#/components/parameters/offset'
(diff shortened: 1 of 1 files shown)


docs/openapi.yml
```diff
@@ -102,6 +105,9 @@ paths:
     parameters:
       - $ref: '#/components/parameters/namespace'
     get:
+      parameters:
```
wslulciuc: _Minor:_ The namespace endpoint doesn't currently support `limit` and `offset`, but I'll be working on introducing that functionality in the next few days

PR #600, opened 2019-07-21
Add to and from json tests for API models

wslulciuc:
This PR tests the (de)serialization of **classes** `NamespaceRequest`, `DatasetRequest`, and `JobRequest` to/from Json.


src/test/resources/fixtures/namespace/request_no_description.json
```diff
@@ -0,0 +1,4 @@
+{
+  "ownerName": "analytics",
+  "description": null
```
ashulmanWeWork: This is a really good set of tests, particularly for user input very close to the edge. Glad we're doing these!

That said, itseems like this file is intended to be what the user sends over. If so, for me, when I'd like to create a request using cUrl or another REST client, and I don't have a description for something, I usually just omit instead of sending over null or even something blank. This is different than the case you're testing here.

I think this set of tests would definitely benefit from testing the case of the description field being omitted entirely (both in this particular example and in other situations w/o a description).

For example, in this case, I think a more representative test case would be to just send over the ownerName and verify the result is what you expect.

wslulciuc: I've removed the static json fixtures representing req/resp payloads in favor of using [`ApiModelGenerator`](https://github.com/MarquezProject/marquez/blob/feature/to-from-json-http-req-tests/src/test/java/marquez/api/models/ApiModelGenerator.java) to ensure valid values and inlining json within each test.

> I don't have a description for something, I usually just omit instead of sending over null or even something blank. This is different than the case you're testing here.

It would be difficult to differentiate between a field that was set as _null_ vs one that was omitted all together. We'd be testing how Jackson handles such a case and personally don't find much value in adding such tests (`Optional` fields already default to _null_)

src/test/java/marquez/api/models/JobRequestTest.java
```diff
@@ -56,4 +66,38 @@ public void testNewRequest_noDescription() {
...
+
+  @Test
+  public void testNewRequest_fromJson_noDescription() throws Exception {
+    final ArrayNode array0 = MAPPER.valueToTree(REQUEST_NO_DESCRIPTION.getInputDatasetUrns());
+    final ArrayNode array1 = MAPPER.valueToTree(REQUEST_NO_DESCRIPTION.getOutputDatasetUrns());
+    final ObjectNode obj =
+        MAPPER
+            .createObjectNode()
+            .put("location", REQUEST_NO_DESCRIPTION.getLocation())
+            .put("description", REQUEST_NO_DESCRIPTION.getDescription().orElse(null));
```
ashulmanWeWork: I'm wondering what the behavior will be in the case that description is `null`. From what I understand, the simulated json here will contain the description field, with the value `null`. This is great for testing machine-to-machine interactions, where if there is no description, it'll just be set to `null`.

However, I'm concerned that this isn't testing the case where a user just omits the field altogether, because by default, Jackson still serialized fields that are null.

wslulciuc: It's a fair point. But as I mentioned in the response above (see [link](https://github.com/MarquezProject/marquez/pull/600#discussion_r308016493)), we'd be testing how Jackson handles such cases. Also, `Optional` fields already default to _null_.

wslulciuc:
Thanks @ronthalanki-wework for inspiring / introducing [`JsonGenerator`](https://github.com/MarquezProject/marquez/blob/master/src/test/java/marquez/api/models/JsonGenerator.java) 💯

PR #610, opened 2019-08-05
 Add limit and offset support to NamespaceAPI

wslulciuc:
The PR adds support for limiting results when listing namespaces:

#### GET `/namespaces?limit={limit}&offset=offset={offset}`

### DIFF
--- a/src/test/java/marquez/api/resources/NamespaceResourceTest.java
+++ b/src/test/java/marquez/api/resources/NamespaceResourceTest.java
@@ -49,6 +49,8 @@ import org.mockito.junit.MockitoRule;
 import org.mockito.quality.Strictness;
 
 public class NamespaceResourceTest extends NamespaceBaseTest {
+  private static final int LIMIT = 100;
+  private static final int OFFSET = 0;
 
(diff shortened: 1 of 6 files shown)


src/test/java/marquez/api/resources/NamespaceResourceTest.java
```diff
@@ -115,9 +117,9 @@ public void testGetNamespaceRequestErrorHandling() throws Throwable {
   public void testListNamespaceWithNoResults() throws MarquezServiceException {
     final List<Namespace> existingCoreModelNamespaces = Collections.emptyList();
     NamespaceResource namespaceResource = new NamespaceResource(namespaceService);
-    when(namespaceService.getAll()).thenReturn(existingCoreModelNamespaces);
+    when(namespaceService.getAll(LIMIT, OFFSET)).thenReturn(existingCoreModelNamespaces);
 
-    Response res = namespaceResource.list();
+    Response res = namespaceResource.list(LIMIT, OFFSET);
```
ashulmanWeWork: nit: we can make this field final.

wslulciuc: I only wanted to make sure the tests passed. But, yeah, when we refactor the test suite, we can address felids that are set as `final`

Issue #176, opened 2018-12-01
Add timeout to wait-for-db.sh


wslulciuc:
Let's add an optional `timeout` param (with a reasonable default) to specify the time limit (in seconds) to establish a DB connection:

```
./wait-for-db.sh <host> <port> [timeout]
```

ajaymuppuri:
https://github.com/MarquezProject/marquez/pull/626

ajaymuppuri:
Please review

wslulciuc:
Fixed #626

PR #626, opened 2019-10-16
Add timeout for the wait script

ajaymuppuri:


wslulciuc:
@ajaymuppuri, thanks for your help! Note that PR #176, which I haven't reviewed yet also introducing a `timeout` param for `wait-for-db.sh`. I didn't assign the issue to @soumyasmruti, which is usually the process, but forgot to in this case.

Here's the proposed plan: I'll review both approaches later this week and see if we can combine efforts or simply favor one over another. How does that sound?

docker/wait-for-db.sh
```diff
@@ -18,15 +18,18 @@ set -eu
 
 host="${1}"
 port="${2}"
+timeout="${3:-20}"
+NEXT_WAIT_TIME=0
 
-until PGPASSWORD="${POSTGRES_PASSWORD}" psql \
+until [ PGPASSWORD="${POSTGRES_PASSWORD}" psql \
```
wslulciuc: _Minor_: `[[ ... ]]` is preferred over `[`, see [Google Shell Style Guide](https://google.github.io/styleguide/shell.xml?showone=Test,_%5B_and_%5B%5B#Test,_%5B_and_%5B%5B).

docker/wait-for-db.sh
```diff
@@ -18,15 +18,18 @@ set -eu
...
         --host="${host}" \
         --port="${port}" \
         --username "${POSTGRES_USER}" \
         --dbname "${POSTGRES_DB}" \
-        --command '\q' > /dev/null 2>&1; do
+        --command '\q' > /dev/null 2>&1 || $NEXT_WAIT_TIME -eq "${timeout}" ]; do
   echo "Waiting for postgres to become available..."
   sleep 1
+  let NEXT_WAIT_TIME=NEXT_WAIT_TIME+1
```
wslulciuc: We can simply the timeout logic by decrementing the `timeout` variable by `1`. For example,

```
sleep 1
timeout=$(( timeout-1 ))
```

This eliminates the need to define the extra variable `NEXT_WAIT_TIME`. Now, on each connection attempt, we simply check:

 ```
[[ ... || $timeout -eq 0 ]]
```

wslulciuc:
Hey, @ajaymuppuri! It'd be great if we can merged this sweet PR soon :)

Issue #632, opened 2019-10-26
Add Job.context


wslulciuc:
Let's add a **context** field defined as a `Map` to [Job](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/service/models/Job.java). A context can be used for getting additional details about a job. The context should be associated with a _job version_ and used in the versioning function applied to job metadata, see [`JobMeta.version()`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/service/models/JobMeta.java#L58).

### Example

Take the simple case where we collect metadata on a batch job that executes the following SQL:

```
SELECT * FROM room_bookings WHERE room_num = '2H';
```

The SQL value above would be stored in `Job.context` under the key `sql`.

**`METADATA`**

```
{
  "type": "BATCH",
  “name”: “room_bookings_7_days”,
  "inputs": [
    "wedata.room_bookings"
  ],
  "outputs": [],
  "location": "https://github.com/wework/jobs/commit/124f6089ad4c5fcbb1d7b33cbb5d3a9521c5d32c",
  "context": {
    "sql": "SELECT * FROM room_bookings WHERE room_num = '2H';"
  }
  "description": "Determine weekly room booking occupancy patterns."
}
```

### Open questions

1. Should we limited the number of allowed key/value pairs in the context (maybe <= `100`?)

/cc @julienledem @roaraya8 @fanrukun

julienledem:
job context should probably be limited in total size rather that property count. There's going to be underlying storage limitations

julienledem:
this looks good to me.

wslulciuc:
Sure thing, @julienledem! I'll add validation around total size.

sshah-wework:
Neat! Is the plan to store as KV pairs in the DB?

wslulciuc:
@sshah-wework, when I first thought about how we might store the KV pairs representing `Job.context`, I immediately wanted to use `hstore`. But, there are some limitations. You first need to run the following on our PG instance to install the package:

```
CREATE EXTENSION hstore;
```
see, http://postgresguide.com/cool/hstore.html

Then, you need to enable it for [Jdbi]( http://jdbi.org/#_hstore).

I feel we're fine with `JSON`. This also allows the application to be less dependent on the choice of DB. Thoughts?

wslulciuc:
Fixed #636

PR #636, opened 2019-10-31
Add Job.context

wslulciuc:
This PR updates the [Job API](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/api/JobResource.java) allowing an optional `context` to be provided when collecting metadata for a given job #632. A context is associated with a _job version_. Therefore, to limit storage costs, we have introduced the following table:

### Table `job_contexts`

| COLUMNS                            |                                                                 |
|------------------------------------|-----------------------------------------------------------------|
| **`uuid`** <br> `uuid`             | _Primary Key_                                                   |
| **`created_at `** <br> `timestamp` | An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp |
| **`context `** <br> `string`       | The list of key/value pairs as `JSON`                           |
| **`checksum `** <br> `string`      | The checksum of the context                                     |

This allows us to associate a row in `job_versions` with a row in `job_contexts`.  For more details, see [Marquez Data Model](https://www.lucidchart.com/documents/edit/f918ce01-9eb4-4900-b266-49935da271b8/DUI5AzPyU0jc).

## API Example

Using the newly added field `context` in [JobRequest](https://github.com/MarquezProject/marquez/blob/feature/job-context/src/main/java/marquez/api/models/JobRequest.java), we can now store, for example, the SQL for _batch_ jobs under the key `sql`:

```
SELECT * FROM room_bookings WHERE room = '2H';
```

**`REQUEST`**

#### `PUT` /api/v1/namespaces/`wedata`/jobs/`room_bookings_7_days`

```
{
  "type": "BATCH",
  "inputs": ["wedata.room_bookings"],
  "outputs": [],
  "location": "https://github.com/wework/jobs/commit/124f6089ad4c5fcbb1d7b33cbb5d3a9521c5d32c",
  "context": {
    "sql": "SELECT * FROM room_bookings WHERE room = '2H';"
  }
  "description": "Weekly email of room bookings occupancy patterns."
}
```

**`RESPONSE`**

```
{
  "type": "BATCH",
  "name": "room_bookings_7_days",
  "createdAt": "2019-10-31T16:04:58.758Z",
  "updatedAt": "2019-10-31T16:04:58.848Z",
  "inputs": ["wedata.room_bookings"],
  "outputs": [],
  "location": "https://github.com/wework/jobs/commit/124f6089ad4c5fcbb1d7b33cbb5d3a9521c5d32c",
  "context": {
    "sql": "SELECT * FROM room_bookings WHERE room = '2H';"
  },
  "description": "Determine weekly room booking occupancy patterns."
}
```

## Related Issues

* Define valid Job.context, #637


src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -74,7 +75,7 @@ default void insertAndUpdate(JobVersionRow row) {
   void update(UUID rowUuid, Instant updatedAt, UUID latestRunUuid);
 
   @SqlQuery(
-      "SELECT j.uuid AS job_uuid, j.namespace_uuid, jv.*, "
+      "SELECT j.uuid AS job_uuid, j.namespace_uuid, jv.*, jc.uuid AS job_context_uuid, jc.context, "
```
fanrukun: it's bit risky to return  explicit columns mixed with * of some other tables. to maintain in the long run,  we usually specify all columns explicitly.  so when project goes very large, it's easy to find out by searching for the column name.  But we probably can leave this kind of refactoring work into future.

src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -106,17 +109,32 @@ default void insertAndUpdate(JobVersionRow row) {
...
+          + "  ON (job_context_uuid = jc.uuid) "
           + "ORDER BY created_at DESC "
           + "LIMIT 1")
-  Optional<JobVersionRow> findLatest(String namespaceName, String jobName);
+  Optional<ExtendedJobVersionRow> findLatest(String namespaceName, String jobName);
 
   @SqlQuery(
-      "SELECT j.* FROM jobs AS j "
+      "SELECT j.uuid AS job_uuid, j.namespace_uuid, jv.*, jc.uuid AS job_context_uuid, jc.context "
+          + "ARRAY(SELECT dataset_uuid "
```
fanrukun: I assume the j.uuid is unique in the query result, otherwise the Array might get mixed up with multiple jobs.

src/main/java/marquez/service/JobService.java
```diff
@@ -82,12 +89,17 @@ public Job createOrUpdate(
...
         final NamespaceRow namespaceRow = namespaceDao.findBy(namespaceName.getValue()).get();
         final JobRow newJobRow = Mapper.toJobRow(namespaceRow.getUuid(), jobName, meta);
-
         jobDao.insert(newJobRow);
       }
 
       final UUID version = meta.version(namespaceName, jobName);
       if (!versionDao.exists(version)) {
+        final String checksum = Utils.checksumFor(meta.getContext());
+        if (!contextDao.exists(checksum)) {
```
fanrukun: are we saying if the checksum is the same and job version is different, different job version share the same job context ?  I thought each job version will come with a new job context.

src/main/java/marquez/db/JobVersionDao.java
wslulciuc: > I assume the j.uuid is unique in the query result, otherwise the Array might get mixed up with multiple jobs.

Yep, ` j.uuid` is unique!

src/main/java/marquez/api/models/JobResponse.java
```diff
@@ -65,6 +70,10 @@ public JobResponse(
     return ImmutableList.copyOf(new ArrayList<>(outputs));
   }
 
+  public Map<String, String> getContext() {
```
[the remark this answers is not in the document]
wslulciuc: Given that `context` is _optional_, we apply the same logic for accessing the job context map in `JobResponse`.

src/main/resources/db/migration/V2__add_job_contexts.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE job_contexts (
+  uuid       UUID PRIMARY KEY,
+  created_at TIMESTAMP NOT NULL,
+  context    TEXT NOT NULL,
```
wslulciuc: @julienledem @fanrukun, have any strong options on what type we should use for storing `JSON`? That is, `TEXT` vs `VARCHAR`

fanrukun: https://stackoverflow.com/questions/25300821/difference-between-varchar-and-text-in-mysql

I use varchar most of the time (as it can specify the length limit for efficiency).  I guess in your case both looks fine.

wslulciuc: Given we're unsure of what the minimum size (in _bytes_) for `context` will be (see #637 ), using `TEXT` provides us the flexibility for now. We can enforce the size constraint in the app layer.

PR #645, opened 2019-11-08
Make Job.location optional

wslulciuc:
This PR makes the `location` of a given job _optional_.


src/main/java/marquez/service/models/JobMeta.java
```diff
@@ -74,7 +78,7 @@ public UUID version(@NonNull NamespaceName namespaceName, @NonNull JobName jobNa
                     .map(output -> output.getValue())
                     .collect(joining(VERSION_DELIM)),
                 KV_JOINER.join(getContext()),
-                location.toString())
+                getLocation().map(location -> location.toString()).orElse(null))
```
roaraya8: you could do `URL::toString` here

src/main/java/marquez/api/mappers/Mapper.java
```diff
@@ -184,7 +184,7 @@ public static JobResponse toJobResponse(@NonNull final Job job) {
         ISO_INSTANT.format(job.getUpdatedAt()),
         job.getInputs().stream().map(input -> input.getValue()).collect(toImmutableList()),
         job.getOutputs().stream().map(output -> output.getValue()).collect(toImmutableList()),
-        job.getLocation().toString(),
+        job.getLocation().map(location -> location.toString()).orElse(null),
```
roaraya8: you could use `.map(URL::toString)` here as well

src/main/java/marquez/service/models/JobMeta.java
wslulciuc: Yes! And hand no idea this was an option 💯

src/main/java/marquez/api/mappers/Mapper.java
wslulciuc: Done!

PR #649, opened 2019-11-12
Add dataset field versioning support

wslulciuc:


src/main/java/marquez/db/DbTableVersionDao.java
```diff
@@ -0,0 +1,61 @@
...
+import marquez.db.mappers.DbTableVersionRowMapper;
+import marquez.db.models.DbTableVersionRow;
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.sqlobject.SqlObject;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+@RegisterRowMapper(DbTableVersionRowMapper.class)
+public interface DbTableVersionDao extends SqlObject {
+  default void insert(DbTableVersionRow row) {
```
fanrukun: I assume this insert function lives in a transaction context. correct ?

src/main/resources/db/migration/V4__add_db_table_versions.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE db_columns (
+  uuid         UUID PRIMARY KEY,
+  type         VARCHAR(64) NOT NULL,
+  created_at   TIMESTAMP NOT NULL,
+  updated_at   TIMESTAMP NOT NULL,
+  dataset_uuid UUID REFERENCES datasets(uuid),
+  name         UUID NOT NULL,
+  desscription UUID,
```
fanrukun: typo ?

fanrukun: alos should the type of name/description be UUID ?  shouldn't they be varchar ?

wslulciuc: Great catch!

src/main/java/marquez/db/DbTableVersionDao.java
wslulciuc: Yep! It definitely should 👍

PR #661, opened 2019-12-02
adding schema required by tagging feature

fanrukun:
This is to add the table schema for tagging required feature.
It creates tags table and the mapping with dataset and data_set field.


src/main/resources/db/migration/V5__add_tag.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE tags (
+  uuid        UUID PRIMARY KEY,
+  created_at  TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
wslulciuc: We'll want the application layer to set the timestamp (for example, [`Mapper.toDatasetRow()`](https://github.com/MarquezProject/marquez/blob/master/src/main/java/marquez/service/mappers/Mapper.java#L188)). Meaning, we can remove defaulting the timestamp to `CURRENT_TIMESTAMP`.

src/main/resources/db/migration/V5__add_tag.sql
```diff
@@ -0,0 +1,19 @@
...
+  updated_at  TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  name        VARCHAR(255) NOT NULL UNIQUE,
+  description TEXT
+);
+CREATE TABLE dataset_tag_mapping (
```
wslulciuc: _Minor_: For readability, let's be consistent with how we format (spacing, etc) our SQL statements :) see, [`V1__initial_schema.sql`](https://github.com/MarquezProject/marquez/blob/master/src/main/resources/db/migration/V1__initial_schema.sql#L1)

src/main/resources/db/migration/V5__add_tag.sql
fanrukun: fixed.

src/main/resources/db/migration/V5__add_tag.sql
fanrukun: fixed

PR #662, opened 2019-12-03
add tagging Resource and related support class

fanrukun:
add tagging Resource and related support class. 
The newly added resource will allow listing of exiting tags.


src/main/java/marquez/api/mappers/Mapper.java
```diff
@@ -16,52 +16,23 @@
...
+import static java.util.Collections.unmodifiableList;
+import static java.util.stream.Collectors.toList;
 
 import java.net.URI;
 import java.net.URL;
 import java.time.Instant;
 import java.util.List;
 import java.util.UUID;
 import lombok.NonNull;
-import marquez.api.models.DatasetRequest;
```
fanrukun: all the imports issue will be fixed in next iteration

wslulciuc:
Close PR. The tagging feature is being introduced in #664

PR #663, opened 2019-12-03
Link dataset versions with run inputs

wslulciuc:
This PR links input dataset versions to a given run using the table [`runs_input_mapping`](https://github.com/MarquezProject/marquez/blob/master/src/main/resources/db/migration/V1__initial_schema.sql#L119).

### DIFF
--- a/src/main/java/marquez/service/JobService.java
+++ b/src/main/java/marquez/service/JobService.java
@@ -30,6 +30,7 @@ import marquez.common.models.DatasetName;
 import marquez.common.models.JobName;
 import marquez.common.models.NamespaceName;
 import marquez.db.DatasetDao;
+import marquez.db.DatasetVersionDao;
 import marquez.db.JobContextDao;
 import marquez.db.JobDao;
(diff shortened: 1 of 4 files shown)


src/main/java/marquez/service/JobService.java
```diff
@@ -267,23 +274,32 @@ public Run createRun(
...
       final RunRow newRunRow = Mapper.toRunRow(versionRow.getUuid(), runArgsRow.getUuid(), runMeta);
+      final List<UUID> inputVersionUuids =
+          datasetVersionDao.findAllInUuidList(versionRow.getInputUuids()).stream()
+              .map(DatasetVersionRow::getUuid)
+              .collect(toImmutableList());
 
-      runDao.insert(newRunRow);
-      markRunAs(newRunRow.getUuid(), Run.State.NEW);
+      runDao.insertWith(newRunRow, inputVersionUuids);
       log.info(
```
fanrukun: seems the log.info should be after markRunas new ?

PR #665, opened 2019-12-05
add support for initSql setting in Flyway

roaraya8:


src/main/java/marquez/MarquezConfig.java
```diff
@@ -30,4 +30,8 @@
   @Getter
   @JsonProperty("flyway")
   private final FlywayFactory flywayFactory = new FlywayFactory();
+
+  @Getter
+  @JsonProperty("flywayInitSql")
+  private final String flywayInitSql = null;
 }
```
wslulciuc: We can override the getter generated by lombok to return an `Optional<String>` instead, making it more explicit that `MarquezConfig.flywayInitSql` is nullable. For example:

```java
// MarquezConfig.java
public Optional<String> getFlywayInitSql() {
  return Optional.ofNullable(flywayInitSql);
}

// MarquezApp.java
private static final String NO_INIT_SQL = "";
private void migrateDbOrError(@NonNull MarquezConfig config) {
  ...
  final String initSqlOrNull = config.getFlywayInitSql().orElse(NO_INIT_SQL);
  flyway.setDataSource(db.getUrl(), db.getUser(), db.getPassword(), initSql);
}
```

src/main/java/marquez/MarquezConfig.java
```diff
@@ -30,4 +30,8 @@
   @Getter
   @JsonProperty("flyway")
   private final FlywayFactory flywayFactory = new FlywayFactory();
+
+  @Getter
+  @JsonProperty("flywayInitSql")
+  private final String flywayInitSql = null;
```
wslulciuc: _Minor_: `null` is the default, so and this can just be:

``` java
private final String flywayInitSql;
```

src/main/java/marquez/MarquezConfig.java
roaraya8: Agree that with Optional it makes it a lot safer, but ideally we can soon move to using the FlywayFactory.

wslulciuc: Sure, sounds good!

PR #666, opened 2019-12-05
Add Dataset.lastModified

wslulciuc:
This PR introduces the column `last_modified` to the `datasets` table.


src/main/java/marquez/service/JobService.java
```diff
@@ -129,13 +130,12 @@ public Job createOrUpdate(
         final NamespaceRow namespaceRow = namespaceDao.findBy(namespaceName.getValue()).get();
         final JobRow newJobRow = Mapper.toJobRow(namespaceRow.getUuid(), jobName, jobMeta);
         jobDao.insert(newJobRow);
+        jobs.labels(namespaceName.getValue(), jobMeta.getType().toString()).inc();
```
fanrukun: what does this labels line do ?

wslulciuc: Labels are a Prometheus concept, see [labels](https://prometheus.io/docs/practices/naming/#labels). But, essentially, they add an extra dimension to metrics. Super helpful!

Issue #699, opened 2020-01-11
Add `Location` header to job run creation


nkijak:
To use the job run API you must first create a run to generate the ID to use.  This ID is included in the returned content, but it would be helpful to include a Location header with the full URL to the run resource given that the runs API is quite different from the creation or other APIs. 

**Currently**
```
$ http -v localhost:5000/api/v1/namespaces/dataeng/jobs/edw-pull/runs < fake_run.json
POST /api/v1/namespaces/dataeng/jobs/edw-pull/runs HTTP/1.1
Accept: application/json, */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 109
Content-Type: application/json
Host: localhost:5000
User-Agent: HTTPie/1.0.3

{
    "runArgs": {
        "end": "2020-01-02T00:00:00",
        "env": "dev",
        "start": "2020-01-02T00:00:00"
    }
}

HTTP/1.1 201 Created
Content-Length: 271
Content-Type: application/json
Date: Sat, 11 Jan 2020 19:34:52 GMT

{
    "createdAt": "2020-01-11T19:34:52.432Z",
    "nominalEndTime": null,
    "nominalStartTime": null,
    "runArgs": {
        "end": "2020-01-02T00:00:00",
        "env": "dev",
        "start": "2020-01-02T00:00:00"
    },
    "runId": "261202b3-8fd7-42dc-8446-9575f39e2dc1",
    "runState": "NEW",
    "updatedAt": "2020-01-11T19:34:52.437Z"
}
```

**Requested response**
_Note `Location` header_
```
HTTP/1.1 201 Created
Content-Length: 271
Content-Type: application/json
Location: localhost:5000/api/v1/jobs/runs/261202b3-8fd7-42dc-8446-9575f39e2dc1
Date: Sat, 11 Jan 2020 19:34:52 GMT

{
    "createdAt": "2020-01-11T19:34:52.432Z",
    "nominalEndTime": null,
    "nominalStartTime": null,
    "runArgs": {
        "end": "2020-01-02T00:00:00",
        "env": "dev",
        "start": "2020-01-02T00:00:00"
    },
    "runId": "261202b3-8fd7-42dc-8446-9575f39e2dc1",
    "runState": "NEW",
    "updatedAt": "2020-01-11T19:34:52.437Z"
}
```

wslulciuc:
[...]

PR #701, opened 2020-01-14
Added Location header to run creation

nkijak:
Fix for #699

### DIFF
--- a/src/main/java/marquez/api/JobResource.java
+++ b/src/main/java/marquez/api/JobResource.java
@@ -16,11 +16,11 @@
 package marquez.api;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
-import static javax.ws.rs.core.Response.Status.CREATED;
 
 import com.codahale.metrics.annotation.ExceptionMetered;
 import com.codahale.metrics.annotation.ResponseMetered;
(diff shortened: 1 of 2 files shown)


nkijak:
First time using JAX-RS, so not sure if this is the best way to get the location header.  Didn't want to hardcode the value but the reflection step to get the method seems like it might be bad.

src/main/java/marquez/service/models/Run.java
```diff
@@ -33,6 +33,10 @@
   @NonNull Run.State state;
   @Nullable Map<String, String> args;
 
+  public UUID getId() {
```
wslulciuc: _Minor:_ [Lombok](https://projectlombok.org) will generate getters/setters for us. So, we don't need to explicitly define one.

nkijak:
@wslulciuc made the update but the commits are messy. Should I close this and open another?

wslulciuc:
@nkijak My PRs get messy as well. I recommend [squashing](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) your commits into one. Then, we can merge this sweet feature!

nkijak:
@wslulciuc squashed

Issue #709, opened 2020-01-20
Add qualifier field to SourceMeta and Source


wslulciuc:

julienledem:
This is deprecated in favor of a less constrained Datasource type fields

Issue #707, opened 2020-01-20
Add sources to config.yml


wslulciuc:
Let's add `sources` to `config.yml` to enabling the configuration of sources. The introduction of qualifiers has the advantage of contextualize the source while still maintaining the mapping `DB`**->** `Set<DataType>` needed for validation:

| Source type | Qualifier  | Data Types                                |
|-------------|------------|-------------------------------------------|
| MYSQL       | DB         | [`INTEGER`,`SMALLINT`,`DECIMAL`, **...**] |
| POSTGRESQL  | DB         | [`BIGINT`,`VARCHAR`,`TEXT`, **...**]      |
| KAFKA       | STREAM     | **-**                                     |
| KINESIS     | STREAM     | **-**                                     |
| S3          | FILESYSTEM | **-**                                     |
| GCP         | FILESYSTEM | **-**                                     |

Below we provide an example configuration for the table above: 

### `config.yml`

```yml
# Configure sources
sources:
  # Add databases and allowed data types
  databases:
    - type: MySQL
      dataTypes:
        - INTEGER
        - SMALLINT
        - DECIMAL
        # ...
    - type: PostgreSQL
      dataTypes:
        - BIGINT
        - VARCHAR
        - TEXT
        # ...
  # Add streams
  streams:
    - type: Kafka
    - type: Kinesis
  # Add file systems
  filesystems:
    - type: S3
    - type: GCP
```

julienledem:
This is deprecated in favor of an unconstrained String DataSource type.

Issue #706, opened 2020-01-20
Add SourceRequest.qualifier


wslulciuc:
Let's add the field `qualifier` to **class** `SourceRequest`:

```
{
  "type": ENUM,
  "qualifier": ENUM,
  "connectionUrl": STRING,
  "description": STRING
}  
```

julienledem:
this is deprecated in favor of a less constrained DataSource type field

PR #713, opened 2020-02-11
Add source type config

### LINKED ISSUE #705: Add column qualifier to table sources
Let's add the column `qualifier` to **table** `sources`:

| COLUMN         | TYPE      |
|----------------|-----------|
| uuid           | `UUID`    |
| type           | `ENUM`    |
| qualifier       | `ENUM`    |
| created_at     | `DATE`    |
| updated_at     | `DATE`    |
| name           | `VARCHAR` |
| connection_url | `VARCHAR` |
| description    | `TEXT`    |

wslulciuc:
This PR adds support for defining source types in `config.yml`. To simplify type configuration, the following sources are predefined: 

- `MYSQL`
- `POSTGRESQL`
- `REDSHIFT`
- `SNOWFLAKE`
- `KAFKA`
- `KINESIS`
- `S3`
- `GCS`

Closes #705, #709, #706, #707

**Other features include:**

- **`400`** when attempting to link a dataset type to a source that is not supported (ex: a stream dataset should only be linked to sources like Kafka, Kinesis, etc and not PostgreSQL.)
- **`400`** when a data type is not supported by a source
- **`400`** when a source type is not supported
- **`400`** when attempting to link a dataset to an unknown source

### Create source

Below we create the source `snowflake_db` with type `SNOWFLAKE` using the qualifier `DB`:

```
PUT /api/v1/sources/snowflake_db

{
  "type":"SNOWFLAKE",
  "qualifier":"DB",
  "connectionUrl":"jdbc:snowflake://myorg.snowflakecomputing.com"
}
```

**`200`**

### Add new source type

Marquez will return a **`400`** for sources not defined in `config.yml`. For example:

```
PUT /api/v1/sources/my_db

{
  "type":"MYDB",
  "qualifier":"DB",
  "connectionUrl":"jdbc:mydb://myorg.mydb.com"
}
```

```
{
  "code": 400,
  "message": "Source type 'MyDb' not supported. Please see your configuration to make sure the type has been enabled for 'DB'."
}
```

Below are the steps to add support for the type `MYDB`:

1. Update `config.yml`:

   ```yml
   sourceTypes:
   dbs:
     - type: MYDB
       dataTypes:
         - INTEGER
         - SMALLINT
         - DECIMAL
         # ...
   ```
2. Restart HTTP API server, and retry:

    ```
    PUT /api/v1/sources/my_db
    
    {
      "type":"MYDB",
      "qualifier":"DB",
      "connectionUrl":"jdbc:mydb://myorg.mydb.com"
    }
    ```
    
    ```
    {
      "type": "MYDB",
      "qualifier": "DB",
      "name": "my_db",
      "createdAt": "2020-02-11T18:51:32.936Z",
      "updatedAt": "2020-02-11T18:51:32.936Z",
      "connectionUrl": "jdbc:mydb://myorg.mydb.com",
      "description": null
    }
    ```
    **`200`** 🎉


src/main/java/marquez/common/models/SourceQualifier.java
```diff
@@ -14,10 +14,8 @@
...
-public enum SourceType {
-  MYSQL,
-  POSTGRESQL,
-  REDSHIFT,
-  SNOWFLAKE,
-  KAFKA;
+public enum SourceQualifier {
+  DB,
+  STREAM,
+  FILESYSTEM;
```
julienledem: I would say that the SourceQualifier enum is redundant with the DatasetType enum. We can keep only one. The SourceType can contain datasets of type X (as define by DatasetType). In this PR there is a 1-1 relation between the elements of SourceQualifier and DatasetType. I would suggest to add a field in dataset type for the name of the sourceQualifier if necessary and just remove SourceQualifier altogether.

src/main/java/marquez/db/Columns.java
```diff
@@ -48,6 +48,7 @@ private Columns() {}
   public static final String OWNER_UUID = "owner_uuid";
 
   /* SOURCE ROW COLUMNS */
+  public static final String QUALIFIER = "qualifier";
```
julienledem: this would become dataset type.

src/main/java/marquez/common/models/Field.java
```diff
@@ -27,18 +27,18 @@
 @Value
 public class Field {
   String name;
-  FieldType type;
```
julienledem: if there's a way to cleanly wrap/unwrap those in json I would replace the enum with a value type rather than remove FieldType here.

src/main/java/marquez/service/models/Source.java
```diff
@@ -35,4 +37,17 @@
...
+    switch (qualifier) {
+      case DB:
+        return datasetType == DatasetType.DB_TABLE;
+      case STREAM:
+        return datasetType == DatasetType.STREAM;
+      case FILESYSTEM:
+        return datasetType == DatasetType.OBJECT;
+      default:
+        throw new AssertionError();
+    }
```
julienledem: this becomes unnecessary if we keep only one of those two enums. (side note, you can add fields and methods to enums so that you don't need a switch statement)

src/main/java/marquez/service/models/SourceTypes.java
```diff
@@ -0,0 +1,348 @@
...
+    public Filesystem(@NonNull final String type) {
+      this.type = checkNotBlank(type, "type must not be blank");
+    }
+  }
+
+  private static final boolean NO_SUPPORT = false;
+
+  private Map<String, Db> dbMap;
+  private Map<String, Stream> streamMap;
+  private Map<String, Filesystem> filesystemMap;
```
julienledem: these should be final

src/main/java/marquez/service/models/SourceTypes.java
```diff
@@ -0,0 +1,348 @@
...
+        });
+    // Overrides
+    if (overrides != null) {
+      overrides.forEach(
+          override -> {
+            dbMap.put(override.getType().toUpperCase(), override);
+          });
+    }
+    return ImmutableMap.copyOf(dbMap);
+  }
```
julienledem: I'd use a Type parameter to factor out those 3 methods.

src/main/java/marquez/service/models/SourceTypes.java
```diff
@@ -0,0 +1,348 @@
...
+
+    final String typeAsUpperCase = type.toUpperCase();
+    switch (qualifier) {
+      case DB:
+        return dbMap.containsKey(typeAsUpperCase)
+            ? dbMap.get(typeAsUpperCase).supports(dataType)
+            : NO_SUPPORT;
+      default:
+        // We currently only support data types for a database source.
+        return NO_SUPPORT;
```
julienledem: I would put that case first:
if (!qualifier.equals(DB)) return false;
then you don't need a switch statement anymore

src/main/java/marquez/service/models/SourceTypes.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+  }
+
+  private static final Db MYSQL =
```
julienledem: I would add a comment to clarify those are defaults that can be overridden in the conf

src/main/java/marquez/api/DatasetResource.java
```diff
@@ -202,6 +214,27 @@ private void throwIfNotExists(
     }
   }
 
+  private void throwIfNotSupported(
+      @NonNull String sourceString, @NonNull DatasetType datasetType, @NonNull List<Field> fields)
+      throws MarquezServiceException {
+    final SourceName sourceName = SourceName.of(sourceString);
+    final Source source =
+        sourceService.get(sourceName).orElseThrow(() -> new SourceNotFoundException(sourceName));
```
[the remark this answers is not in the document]
wslulciuc: Great idea! 👍

frankcash:
I almost wonder instead of S3 it should just be `ObjectStorage`, which would then just fit cozy with all the cloud providers (GCP, DO, AWS) with less work?

wslulciuc:
> I almost wonder instead of S3 it should just be ObjectStorage, which would then just fit cozy with all the cloud providers (GCP, DO, AWS) with less work?

@frankcash  I think it would be helpful to know the object storage type as it would help with abstracting away all the data consumption details from the job. For example, let's say we have a job that processes logs from cloud storage with some logs existing in `S3` or `GCS`. The job would: 1) make a call to Marquez to get the connection details for the source 2) depending on the source (`S3`, `GCP`, etc), it would know which client to instantiate to read the files.

frankcash:
I think that could be handled by URI formatting for the connection url though, but I do get what you're saying!

MrNeocore:
Hi,

I'm new to this project, it looks very promising !
I was curious what's the status on this issue / MR ?

Thank you :)

PR #732, opened 2020-03-27
add RunTransitionListener to JobService

julienledem:


.gitignore
```diff
@@ -17,3 +17,4 @@ build/
 
 # Marquez configuration
 config.yml
+node_modules
```
wslulciuc: _Minor_: We don't need to ignore `node_modules`

julienledem: ah sorry about that

wslulciuc: No problem at all!

PR #733, opened 2020-03-28
add unique identifiers DatasetId to Dataset and JobId to Job

julienledem:
Signed-off-by: Julien Le Dem <julien@apache.org>

### DIFF
--- a/src/main/java/marquez/api/models/DatasetResponse.java
+++ b/src/main/java/marquez/api/models/DatasetResponse.java
@@ -26,6 +26,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.ToString;
+import marquez.common.models.DatasetName;
 import marquez.common.models.Field;
 import marquez.service.models.DatasetId;
 
@@ -38,7 +39,7 @@ import marquez.service.models.DatasetId;
 })
 public abstract class DatasetResponse {
   @Getter private final DatasetId id;
-  @Getter private final String name;
+  @Getter private final DatasetName name;
   @Getter private final String physicalName;
   @Getter private final String createdAt;
   @Getter private final String updatedAt;
@@ -50,7 +51,7 @@ public abstract class DatasetResponse {
 
   public DatasetResponse(
       @NonNull final DatasetId id,
-      @NonNull final String name,
+      @NonNull final DatasetName name,
       @NonNull final String physicalName,
       @NonNull final String createdAt,
       @NonNull final String updatedAt,
@@ -60,7 +61,7 @@ public abstract class DatasetResponse {
       @Nullable final String lastModifiedAt,
       @Nullable final String description) {
     this.id = id;
-    this.name = checkNotBlank(name);
+    this.name = name;
     this.physicalName = checkNotBlank(physicalName);
     this.createdAt = checkNotBlank(createdAt);
     this.updatedAt = checkNotBlank(updatedAt);
--- a/src/main/java/marquez/api/models/DatasetResponse.java
+++ b/src/main/java/marquez/api/models/DatasetResponse.java
@@ -27,6 +27,7 @@ import lombok.Getter;
 import lombok.NonNull;
 import lombok.ToString;
 import marquez.common.models.Field;
+import marquez.service.models.DatasetId;
 
 @EqualsAndHashCode
 @ToString
@@ -36,6 +37,7 @@ import marquez.common.models.Field;
   @JsonSubTypes.Type(value = StreamResponse.class, name = "STREAM")
 })
 public abstract class DatasetResponse {
+  @Getter private final DatasetId id;
   @Getter private final String name;
   @Getter private final String physicalName;
   @Getter private final String createdAt;
@@ -47,6 +49,7 @@ public abstract class DatasetResponse {
   @Nullable private final String description;
 
   public DatasetResponse(
+      @NonNull final DatasetId id,
       @NonNull final String name,
       @NonNull final String physicalName,
       @NonNull final String createdAt,
@@ -56,6 +59,7 @@ public abstract class DatasetResponse {
       @Nullable final List<String> tags,
       @Nullable final String lastModifiedAt,
       @Nullable final String description) {
+    this.id = id;
     this.name = checkNotBlank(name);
     this.physicalName = checkNotBlank(physicalName);
     this.createdAt = checkNotBlank(createdAt);
--- a/src/main/java/marquez/api/models/DatasetResponse.java
+++ b/src/main/java/marquez/api/models/DatasetResponse.java
@@ -28,6 +28,7 @@ import lombok.NonNull;
 import lombok.ToString;
 import marquez.common.models.DatasetName;
 import marquez.common.models.Field;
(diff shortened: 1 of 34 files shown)


src/main/java/marquez/service/models/JobId.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.NonNull;
+import lombok.Value;
+import marquez.common.models.JobName;
+import marquez.common.models.NamespaceName;
+
+@Value
+public class JobId {
+
+  @NonNull NamespaceName namespace;
+  @NonNull JobName name;
```
wslulciuc: What are your thoughts on having the underlying value of the ID be a _string_? That is, we'd define a factory class to create instances of `JobId `:

```java
NamespaceName NamespaceName = NamespaceName.of("my_namespace");
JobName jobName = JobName.of("my_job");

JobId id = JobId.of(NamespaceName, JobName)`
log.info(id.getValue())
```
```
>> my_namespace:my_job
```

julienledem: As the url of a Job resource is ```/namespaces/{namespceName}/jobs/{name}``` it makes sense to me that the ID calls those out its components explicitly. That being said we could have something more compact. I'd use a separator (like ```:```) only if we have a non ambiguous way of parsing it back (for example namespace and name don't contain ```:```) . If we wanted this, I'd keep JobId as is but would add converters to serialize it as a string using the JsonSerializer annotation.

julienledem:
I added an ID field because it is useful to represent the unique identifier of a resource. It is more intuitive IMO than explaining that the namespace and name together are unique. It then also used to refer to them, for example referring to the DatasetId as inputs and outputs of a job.

src/main/java/marquez/service/JobService.java
```diff
@@ -226,6 +216,39 @@ public Job createOrUpdate(
...
+              String namespace = e.getKey().getValue();
+              List<String> names =
+                  e.getValue().stream()
+                      .map((id) -> id.getName().getValue())
+                      .collect(toImmutableList());
+              List<DatasetRow> results = datasetDao.findAllIn(namespace, names);
+              if (results.size() < names.size()) {
+                List<String> actual =
+                    results.stream().map(DatasetRow::getName).collect(toImmutableList());
+                throw new DatasetNotFoundException(
```
wslulciuc: We'll want to avoid throwing an exception in the service layer that is defined in the API layer. For now, maybe we just throw the catch-all exception `MarquezServiceException`?

src/main/java/marquez/service/models/JobId.java
wslulciuc: I think this makes a lot of sense. A job or dataset is unique within a namespace and the composite of the two (`dataset|job`+ `namespace`) should make up the ID itself. By introducing a separator, we are present with other concerns. So, making each component explicit avoids us having to define constraints on what valid characters can be present in a name.

wslulciuc:
> It then can also be used to refer to them, for example referring to the DatasetId as inputs and outputs of a job.

💯

src/main/java/marquez/api/models/DatasetResponse.java
```diff
@@ -56,7 +60,8 @@ public DatasetResponse(
       @Nullable final List<String> tags,
       @Nullable final String lastModifiedAt,
       @Nullable final String description) {
-    this.name = checkNotBlank(name);
+    this.id = id;
+    this.name = name;
     this.physicalName = checkNotBlank(physicalName);
     this.createdAt = checkNotBlank(createdAt);
     this.updatedAt = checkNotBlank(updatedAt);
```
wslulciuc: _Minor_: Let's also return the namespace in the response?

src/main/java/marquez/api/models/DatasetsResponse.java
```diff
@@ -14,14 +14,17 @@
...
 import lombok.Value;
 
 @Value
 public class DatasetsResponse {
-  @NonNull
-  @JsonProperty("datasets")
-  List<DatasetResponse> responses;
+  @NonNull List<DatasetResponse> datasets;
+
+  @JsonCreator
```
wslulciuc: Do we need `@JsonCreator` if we are only serializing the response?

src/main/java/marquez/api/models/JobRequest.java
```diff
@@ -28,12 +30,20 @@ public JobRequest(
...
+      throw new IllegalArgumentException(
+          "Either use inputs/outputs or inputIds/outputIds but not both at the same time");
+    }
+    this.inputs = inputs == null ? null : ImmutableList.copyOf(inputs);
+    this.outputs = outputs == null ? null : ImmutableList.copyOf(outputs);
+    this.inputIds = inputIds == null ? null : ImmutableList.copyOf(inputIds);
+    this.outputIds = outputIds == null ? null : ImmutableList.copyOf(outputIds);
     this.location = location;
     this.context = (context == null) ? ImmutableMap.of() : ImmutableMap.copyOf(context);
     this.description = description;
```
wslulciuc: _Minor_: Like the `DatasetResponse`, let's also return the namespace in the response?

src/main/java/marquez/api/models/JobResponse.java
```diff
@@ -26,38 +27,49 @@
...
     this.updatedAt = checkNotBlank(updatedAt);
-    this.inputs = ImmutableList.copyOf(inputs);
-    this.outputs = ImmutableList.copyOf(outputs);
+    this.inputIds = ImmutableList.copyOf(inputIds);
+    this.outputIds = ImmutableList.copyOf(outputIds);
+    this.inputs = inputIds.stream().map((i) -> i.getName()).collect(toImmutableList());
+    this.outputs = outputIds.stream().map((i) -> i.getName()).collect(toImmutableList());
     this.location = location;
     this.context = (context == null) ? ImmutableMap.of() : ImmutableMap.copyOf(context);
     this.description = description;
```
wslulciuc: _Minor_: Like `DatasetResponse`, let's also return the namespace in the response?

src/main/java/marquez/common/models/DatasetName.java
```diff
@@ -32,4 +39,18 @@ private DatasetName(final String value) {
   public static DatasetName of(final String value) {
     return new DatasetName(value);
   }
+
+  public static class DatasetToString extends StdConverter<DatasetName, String> {
```
wslulciuc: 💯

src/main/java/marquez/common/models/DatasetName.java
```diff
@@ -32,4 +39,18 @@ private DatasetName(final String value) {
...
+    @Override
+    public String convert(DatasetName value) {
+      return value.getValue();
+    }
+  }
+
+  public static class StringToDataset extends StdConverter<String, DatasetName> {
```
wslulciuc: 💯

src/main/java/marquez/api/models/DatasetsResponse.java
julienledem: I use it to deserialize the response in the integration test.

wslulciuc: Ahh, 👍

PR #746, opened 2020-04-29
Updates openapi to show correct key

frankcash:
Updates openapi to show correct key.  

I tried creating a along the lines of:

```bash
curl -X PUT http://localhost:5000/api/v1/namespaces/data_engineering/jobs/airflow.pim_agg \
  -H 'Content-Type: application/json' \
  -d '{
        "type": "BATCH",
        "inputs": ["frank_dev.items"],
        "outputs": ["frank_dev.items_aggregate"],
        "location": "https://github.com/...",
        "description": "thing.",
        "context": {
          "sql":"INSERT INTO frank_dev.items_aggregate..."
        }
      }'
```

Per the Docs discussion
Screen Shot 2020-04-29 at 9 57 04 AM

This results in an empty sql field when running marquez-web.  To resolve this someone needs to send 

```bash
curl -X PUT http://localhost:5000/api/v1/namespaces/data_engineering/jobs/airflow.pim_agg \
  -H 'Content-Type: application/json' \
  -d '{
        "type": "BATCH",
        "inputs": ["frank_dev.items"],
        "outputs": ["frank_dev.items_aggregate"],
        "location": "https://github.com/...",
        "description": "thing.",
        "context": {
          "SQL":"INSERT INTO frank_dev.items_aggregate..."
        }
      }'
```

Thus I felt it necessary to update the openapi examples

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -573,7 +573,7 @@ components:
         inputs: ["wedata.room_bookings", "analytics.scheduled_events"]
         outputs: ["wedata.room_bookings_7_days"]
         location: https://github.com/wework/jobs/commit/124f6089ad4c5fcbb1d7b33cbb5d3a9521c5d32c
-        context: {'sql': "SELECT * FROM room_bookings WHERE room_num = '2H';"}
+        context: {'SQL': "SELECT * FROM room_bookings WHERE room_num = '2H';"}
         description: Determine weekly room booking occupancy patterns.
         latestRun: null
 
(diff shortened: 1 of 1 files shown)


frankcash:
Alternatively this could be changed here: https://github.com/MarquezProject/marquez-web/blob/493e787aa99ed8bac4b02af2f93091ba1243c138/src/types/api.ts#L37

frankcash:
it looks like circle failed for some reason..?

wslulciuc:
@frankcash The _keys_ for a job context are case-sensitive and wasn't aware that `marquez-web` was strictly looking for `SQL`. Thanks for pointing this out!

> Alternatively this could be changed here: https://github.com/MarquezProject/marquez-web/blob/493e787aa99ed8bac4b02af2f93091ba1243c138/src/types/api.ts#L37

You're correct and maybe support for lower case and upper case keys should be supported in `marquez-web`. That said, storing SQL under the `SQL` key was a convention we decided on at WeWork.

Otherwise PR looks great! Updating the API spec will at least allow the UI to correctly fetch any SQL associated with a job version.

wslulciuc:
> it looks like circle failed for some reason..?

Odd, I kicked off CI and all tests passed. The failed build might be because of a flaky test which I look into.

frankcash:
Happy to help!

PR #779, opened 2020-07-29
Bumping the shadow plugin version to 6.0.0. 

Meehau:
Fixes the ./gradlew runShadow error: "Execution failed for task ':runShadow'. > The value for task ':runShadow' property 'mainClass' is final and cannot be changed any further."

Signed-off-by: Michal Januszewski <michal.januszewski@gmail.com>


Meehau:
I guess this is one of the first PRs after the CI change :) I think we need to lower the code coverage standards for now, at least until we add more tests.

PR #787, opened 2020-08-03
Adding type for `int`

phixMe:


wslulciuc:
This looks great! Thanks, @phixMe! Mind also adding a few tests? You can checkout some examples for `Columns.stringOrThrow()` here: https://github.com/MarquezProject/marquez/blob/main/src/test/java/marquez/db/ColumnsTest.java#L147

Issue #795, opened 2020-08-26
Add support for external run ID


wslulciuc:
Often, frameworks / systems have run IDs that they manage internally. For example, Airflow uses a [run ID](https://airflow.apache.org/docs/stable/_api/airflow/models/dagrun/index.html#airflow.models.dagrun.DagRun.run_id) to manage the state of a DAG run. Currently, our [`marquez-airflow`](https://github.com/MarquezProject/marquez-airflow) lib keeps track of both the Airflow run ID and Marquez run ID. To simply our integration efforts with systems like Airflow, let's extend the Job API to _also_ allow for an external run ID on [run creation](https://github.com/MarquezProject/marquez/blob/main/src/main/java/marquez/api/JobResource.java#L126) in Marquez. One way of adding such support would be to simply add the _optional_ query parameter `external_id` to the API call. For example:

```
POST /api/v1/namespaces/{namespace}/jobs/{job}/runs?external_id={external-id}
```

The run ID can then be used to update the states of a given run as if the ID was created by the Marquez backend:

```
POST /jobs/runs/{external-id}/start
POST /jobs/runs/{external-id}/complete
```

**Note:** The caller _must_ provide an external run ID that is a valid UUID.

wslulciuc:
@julienledem: Assigning to you! and let me know if more clarity is needed

wslulciuc:
Fixed #797

PR #797, opened 2020-08-26
add optional runId when creating a run

julienledem:
per #795
Adding support for client provided runId
Signed-off-by: Julien Le Dem <julien@apache.org>


docs/openapi.yml
```diff
@@ -1028,6 +1028,10 @@ components:
     newRun:
       type: object
       properties:
+        runId:
```
wslulciuc, line 1031: _Minor:_ Thoughts on naming the ID for a run `externalId` to make it explicit that it's user provided?

docs/openapi.yml
```diff
@@ -1028,6 +1028,10 @@ components:
...
+          description: An optional client-provided UUID for this run (this could be a Random or name based UUID). If not provided a random UUID will be generated for the run
+          type: string
+          format: uuid
```
wslulciuc, line 1034: You'll also want to render the `html` using `redoc-cli bundle`, see https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#api-docs

src/test/java/marquez/api/JobResourceTest.java
```diff
@@ -160,6 +161,31 @@ public void testCreateRun() throws MarquezServiceException {
...
+    when(namespaceService.exists(NAMESPACE_NAME)).thenReturn(true);
+    when(jobService.exists(NAMESPACE_NAME, JOB_NAME)).thenReturn(true);
+    when(jobService.createRun(NAMESPACE_NAME, JOB_NAME, runMeta)).thenReturn(run);
+
+    doReturn(runLocation).when(jobResource).locationFor(uriInfo, run);
+
+    final Response response = jobResource.createRun(NAMESPACE_NAME, JOB_NAME, runMeta, uriInfo);
+    assertThat(response.getStatus()).isEqualTo(201);
+    assertThat(response.getLocation()).isEqualTo(runLocation);
+    assertThat((Run) response.getEntity()).isEqualTo(run);
```
wslulciuc, line 186: 👍

src/test/java/marquez/db/JobServiceTest.java
```diff
@@ -198,5 +204,31 @@ public void testRun() throws MarquezServiceException, MalformedURLException {
...
+                ImmutableMap.of(),
+                "description"));
+
+    RunId run2Id = RunId.of(UUID.randomUUID());
+    Run run2 = jobService.createRun(
+        NAMESPACE_NAME, job.getName(), new RunMeta(run2Id, null, null, null));
+    assertThat(run2.getId()).isEqualTo(run2Id);
+    Optional<Run> run2Again = jobService.getRun(run2Id);
+    assertThat(run2Again.isPresent()).isTrue();
+    assertThat(run2Again.get().getId()).isEqualTo(run2Id);
```
wslulciuc, line 232: 💯

docs/openapi.yml
julienledem: I would tend to call it runId here since it becomes the actual ID for the run. I would keep externalId to add in the metadata the actual external id defined by a scheduler. That one could be numerical or a string and may not be a uuid.
In this case, it is provided in the request by the client, so I think it is clear it is user provided. Just like Job and Dataset ids are user provided.

docs/openapi.yml
julienledem: thanks for reminding me.

docs/openapi.yml
wslulciuc: Yep, that totally makes sense. Since it will be the ID of the run returned to the caller, should we keep naming in sync? That is, in the request it's `runId`, but for the response it's just `id`. For example,

**Request**
```
POST /api/v1/namespaces/my-namespace/jobs/my-job/runs
{
  "runId": "cbd759f7-1948-4013-a14b-7043c0d06f35", <-- use 'id' here to match response?
  "args": {...}
}
```

**Response**
```
{
  "id": "cbd759f7-1948-4013-a14b-7043c0d06f35",
  "args": {...}
}
```

julienledem: that makes sense, I'll change to id for consistency.

wslulciuc: Thanks! 👍

PR #800, opened 2020-08-29
add missing parameter documentation

julienledem:


docs/openapi.yml
```diff
@@ -401,6 +405,15 @@ paths:
 
 components:
   parameters:
+    at:
+      name: at
+      in: query
+      description: The time the event happened at
```
wslulciuc, line 411: _Minor_: Given that the `at` query param is used to set the transition time of a given run, what are your thoughts on the description below?

```
An ISO-8601 timestamp representing the time when the run transitioned.
```

PR #810, opened 2020-09-13
Add more dataset resource tests

wslulciuc:
This PR adds tests to ensure a `400` is thrown when tagging a dataset or field that does not exits. This PR also reduces the DB calls made on tagging by using the `Dataset` object retuned by `DatasetService.tagWith()` and `DatasetService.tagFieldWith()`. `DatasetService.fieldExist()` has also been renamed from  `DatasetService.exist()` to make it more obvious on the _field_ existence check that is being made.


julienledem:
as codecov puts it, the diff coverage is 💯 %

julienledem (approved):

PR #818, opened 2020-10-16
Add run ID support on job meta

wslulciuc:
This PR updates the [Job API](https://marquezproject.github.io/marquez/openapi.html#tag/Jobs/paths/~1namespaces~1{namespace}~1jobs~1{job}/put) to support providing an optional run ID along with the job meta. When the metadata for a job changes, Marquez creates a new job version on the backend. The optional run ID will be used to link a newly created job version to an existing job run, while supporting updating the run state and avoiding having to create another run. This PR also bumps `marquez-java` to `0.4.3-rc.1` and updates CI to use `jdk11`


src/main/java/marquez/api/JobResource.java
```diff
@@ -264,9 +265,11 @@ void throwIfExists(
     }
   }
 
-  void throwIfNotExists(@NonNull RunId runId) throws MarquezServiceException {
-    if (!jobService.runExists(runId)) {
-      throw new RunNotFoundException(runId);
+  void throwIfNotExists(@Nullable RunId runId) throws MarquezServiceException {
+    if (runId != null) {
```
julienledem, line 269: this seems unintuitive to me. If we want this method to throw an error when the runid is not found, we should probably expect to call this only in a context where the run exists. I would call `if (!null)` before calling throwIfNotExists.

julienledem (approved):
One minor comment.
THis looks good to me

src/main/java/marquez/api/JobResource.java
wslulciuc: Good point. We can do something more like:

```
jobMeta.getRunId().ifPresent(this::throwIfNotExists);
```

PR #822, opened 2020-11-06
Allow job updates to associate new input datasets.

henneberger:
For workflows like big query, where the input dataset is not available during run creation, we need to associate the input datasets of a completed job. This is done via the job update resources that takes a run id as a context key.

### DIFF
--- a/src/main/java/marquez/db/DatasetVersionDao.java
+++ b/src/main/java/marquez/db/DatasetVersionDao.java
@@ -74,11 +74,11 @@ public interface DatasetVersionDao {
   void updateFields(UUID datasetVersionUuid, UUID datasetFieldUuid);
 
   String SELECT =
-      "SELECT *, "
+      "SELECT dv.*, "
           + "ARRAY(SELECT dataset_field_uuid "
           + "      FROM dataset_versions_field_mapping "
-          + "      WHERE dataset_version_uuid = uuid) AS field_uuids "
-          + "FROM dataset_versions ";
+          + "      WHERE dataset_version_uuid = dv.uuid) AS field_uuids "
+          + "FROM dataset_versions dv ";
 
   String EXTENDED_SELECT =
       "SELECT dv.*, d.name as dataset_name, n.name as namespace_name, "
@@ -91,6 +91,12 @@ public interface DatasetVersionDao {
 
   @SqlQuery(SELECT + "WHERE uuid = :uuid")
   Optional<DatasetVersionRow> findBy(UUID uuid);
+
+  @SqlQuery(
+      SELECT
+          + "INNER JOIN datasets AS d ON d.uuid = dv.dataset_uuid AND d.current_version_uuid = dv.uuid "
+          + "WHERE dv.dataset_uuid = :datasetUuid")
+  Optional<DatasetVersionRow> findLatestDatasetByUuid(UUID datasetUuid);
 
   default Optional<DatasetVersionRow> find(String typeString, @Nullable UUID uuid) {
     if (uuid == null) {
--- a/src/main/java/marquez/service/JobService.java
+++ b/src/main/java/marquez/service/JobService.java
@@ -26,15 +26,18 @@ import com.google.common.collect.ImmutableSet;
 import io.prometheus.client.Counter;
 import io.prometheus.client.Gauge;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
+import marquez.api.exceptions.RunNotFoundException;
 import marquez.common.Utils;
 import marquez.common.models.DatasetId;
 import marquez.common.models.DatasetName;
@@ -54,6 +57,7 @@ import marquez.db.RunArgsDao;
 import marquez.db.RunDao;
 import marquez.db.RunStateDao;
 import marquez.db.models.DatasetRow;
+import marquez.db.models.DatasetVersionRow;
 import marquez.db.models.ExtendedDatasetVersionRow;
 import marquez.db.models.ExtendedJobVersionRow;
 import marquez.db.models.ExtendedRunRow;
@@ -196,8 +200,12 @@ public class JobService {
         }
         final JobRow jobRow = jobDao.find(namespaceName.getValue(), jobName.getValue()).get();
         final JobContextRow contextRow = jobContextDao.findBy(checksum).get();
-        final List<UUID> inputUuids = findUuids(jobMeta.getInputs());
-        final List<UUID> outputUuids = findUuids(jobMeta.getOutputs());
+        final List<DatasetRow> inputRows = findDatasetRows(jobMeta.getInputs());
+        final List<UUID> inputUuids =
+            inputRows.stream().map(DatasetRow::getUuid).collect(Collectors.toList());
+        final List<DatasetRow> outputRows = findDatasetRows(jobMeta.getOutputs());
+        final List<UUID> outputUuids =
+            outputRows.stream().map(DatasetRow::getUuid).collect(Collectors.toList());
         final JobVersionRow newJobVersionRow =
             Mapper.toJobVersionRow(
                 jobRow.getUuid(),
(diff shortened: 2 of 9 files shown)


src/main/java/marquez/db/DatasetVersionDao.java
```diff
@@ -92,6 +92,11 @@ default void insertWith(DatasetVersionRow row, List<DatasetFieldRow> fieldRows)
   @SqlQuery(SELECT + "WHERE uuid = :uuid")
   Optional<DatasetVersionRow> findBy(UUID uuid);
 
+  @SqlQuery(
+      SELECT
+          + "WHERE dataset_uuid = :datasetUuid AND created_at = (select max(created_at) from dataset_versions where dataset_uuid = :datasetUuid group by dataset_uuid) limit 1")
+  Optional<DatasetVersionRow> findLatestDatasetByUuid(UUID datasetUuid);
```
julienledem, line 98: the datasets table has a current_version_uuid so you can find the current version of the dataset with something along the lines of:
SELECT + " INNER JOIN datasets AS d ON d.uuid = dv.dataset_uuid AND d.current_version_uuid = dv.uuid WHERE dv.dataset_uuid = :uuid"

src/main/java/marquez/service/JobService.java
```diff
@@ -242,13 +260,52 @@ public Job createOrUpdate(
...
+    }
+    ExtendedRunRow run = runRow.get();
+    notify(
+        new JobInputUpdate(
+            RunId.of(runUuid),
+            new RunMeta(
+                RunId.of(runUuid),
+                run.getNominalStartTime().orElse(null),
+                run.getNominalEndTime().orElse(null),
+                ImmutableMap.of()),
```
julienledem, line 297: what was this last parameter for?

src/main/java/marquez/service/JobService.java
```diff
@@ -242,13 +260,52 @@ public Job createOrUpdate(
...
+        DatasetVersionRow row = version.get();
+        runDao.updateInputVersions(runUuid, version.get().getUuid());
+        runInputs.add(
+            new RunInput(
+                new DatasetVersionId(
+                    namespaceName, DatasetName.of(inputDataset.getName()), row.getUuid())));
+      }
+    }
+    Optional<ExtendedRunRow> runRow = runDao.findBy(runUuid);
+    if (runRow.isEmpty()) {
```
julienledem, line 286: should we log an error in that case?

src/main/java/marquez/service/JobService.java
```diff
@@ -242,13 +260,52 @@ public Job createOrUpdate(
...
+      JobName jobName,
+      JobVersionRow jobVersion) {
+    if (inputRows.isEmpty()) {
+      return;
+    }
+    List<RunInput> runInputs = new ArrayList<>();
+    for (DatasetRow inputDataset : inputRows) {
+      Optional<DatasetVersionRow> version =
+          datasetVersionDao.findLatestDatasetByUuid(inputDataset.getUuid());
+      if (version.isPresent()) {
```
julienledem, line 276: I think in the existing contract those datasets have to exists. 
So we should log an error if the dataset does not exist.
In the future we may want to implicitly create those datasets.

src/test/java/marquez/MarquezAppIntegrationTest.java
```diff
@@ -461,6 +461,99 @@ public void testApp_createJobAndMarkRunAsComplete() {
...
+            .context(JOB_CONTEXT)
+            .description(JOB_DESCRIPTION)
+            .build();
+    client.createJob(NAMESPACE_NAME, jobName, jobUpdateMeta);
+
+    // Assure datasets are associated
+    Job finalJob = client.getJob(NAMESPACE_NAME, jobName);
+    assertThat(finalJob.getInputs()).hasSize(2);
+
+    // todo: Assure new input dataset has been associated.
```
julienledem, line 554: I think you could pass a mock listener through the config.
Or just add a separate test where you mock the listener in the JobServiceTest

julienledem (commented):
This looks correct.
My main comments are on the logic to get the current version of a dataset and how to test that the proper change has been propagated to the listener.

src/main/java/marquez/service/JobService.java
henneberger: Arguments of the run. I feel like its unnecessary so I would lean more towards removing RunMeta completely rather than reconstructing the object with dao calls.

src/main/java/marquez/db/DatasetVersionDao.java
henneberger: Ah good call

src/main/java/marquez/service/JobService.java
```diff
@@ -196,8 +200,12 @@ public Job createOrUpdate(
         }
         final JobRow jobRow = jobDao.find(namespaceName.getValue(), jobName.getValue()).get();
         final JobContextRow contextRow = jobContextDao.findBy(checksum).get();
-        final List<UUID> inputUuids = findUuids(jobMeta.getInputs());
-        final List<UUID> outputUuids = findUuids(jobMeta.getOutputs());
+        final List<DatasetRow> inputRows = findUuids(jobMeta.getInputs());
+        final List<UUID> inputUuids =
+            inputRows.stream().map(e -> e.getUuid()).collect(Collectors.toList());
+        final List<DatasetRow> outputRows = findUuids(jobMeta.getOutputs());
```
wslulciuc, line 206: _Minor:_ Given the method now returns _row_ objects, I'd rename to `findDatasetRows()`

src/main/java/marquez/service/JobService.java
```diff
@@ -196,8 +200,12 @@ public Job createOrUpdate(
         }
         final JobRow jobRow = jobDao.find(namespaceName.getValue(), jobName.getValue()).get();
         final JobContextRow contextRow = jobContextDao.findBy(checksum).get();
-        final List<UUID> inputUuids = findUuids(jobMeta.getInputs());
-        final List<UUID> outputUuids = findUuids(jobMeta.getOutputs());
+        final List<DatasetRow> inputRows = findUuids(jobMeta.getInputs());
+        final List<UUID> inputUuids =
+            inputRows.stream().map(e -> e.getUuid()).collect(Collectors.toList());
```
wslulciuc, line 205: _Minor:_ You can use `.map(DatasetRow::getUuid)` if you wanted to make the code a bit more compact.

src/main/java/marquez/service/JobService.java
```diff
@@ -210,6 +218,16 @@ public Job createOrUpdate(
         versions
             .labels(namespaceName.getValue(), jobMeta.getType().toString(), jobName.getValue())
             .inc();
+
+        if (jobMeta.getRunId().isPresent()) {
```
wslulciuc, line 222: You can avoid the `if` check and directly called `get()` on the optional by using [`ifPresent()`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ifPresent-java.util.function.Consumer-):

```java
   jobMeta
     .getRunId()
     .ifPresent( ... )
```

We have a similar logic when [creating a job with a run ID](https://github.com/MarquezProject/marquez/blob/main/src/main/java/marquez/service/JobService.java#L219)

src/main/java/marquez/service/JobService.java
wslulciuc: I think we'd want to return the run args as well. The **class** [`ExtendedRunRow`](https://github.com/MarquezProject/marquez/blob/main/src/main/java/marquez/db/models/ExtendedRunRow.java) returns the run args as a Json string (key / value pairs) and can be deserialized using `Utils.fromJson()`. For example https://github.com/MarquezProject/marquez/blob/main/src/main/java/marquez/service/mappers/Mapper.java#L382

We'd might want to add a `toRunArgs()` method to the mapper that factors our the  deserialization logic.

gradle.properties
```diff
@@ -1,4 +1,4 @@
 org.gradle.internal.http.connectionTimeout=60000
 org.gradle.internal.http.socketTimeout=60000
 
-version=0.11.3
+version=0.11.4-SNAPSHOT
```
wslulciuc, line 4: _Minor:_ Your properties file might be out of sync? (`0.11.4-SNAPSHOT` is the latest).

src/main/java/marquez/service/JobService.java
julienledem: actually on second thoughts we might want to throw an error if the run is not found here (same for comment bellow).
The api does that validation too so it wouldn't be a reachable case. It is still better to throw an exception so that if things change in the future it is obvious what's happening:
https://github.com/MarquezProject/marquez/blob/c927c88d17edd1b65f57d5428d441f6c0886ec13/src/main/java/marquez/api/JobResource.java#L87

wslulciuc (approved):
LGTM 👍

julienledem (approved):
👍

PR #823, opened 2020-11-11
Fix connection pool exhaustion when inserting

henneberger:
The getHandle method was obtaining a connection from the pool and not freeing it


wslulciuc (approved):
Thanks, @henneberger! 🥇

PR #828, opened 2020-11-14
Separate Job and Run into separate service/resource layers

henneberger:
Separate job and run into separate service layer. Also, take a more declarative approach to job service layer.


wslulciuc (approved):
This looks great, @henneberger! Nice and clean separation of responsibility between services 👍

PR #844, opened 2020-11-19
Fix docker/up.sh to allow for multiple flags

wslulciuc:
This PR fixes `docker/up.sh` to allow for setting both `--build` and `--seed`. A usage flags is also introduced:

```
 $ ./docker/up.sh --help
usage: ./up.sh [--build] [--seed]
```

This PR also uses the latest [`marquez-web`](https://hub.docker.com/r/marquezproject/marquez-web) image when running `docker/up.sh` and builds from source when using the `--build` flag.


phixMe (approved):
Thanks for fixing this up. Works on my side with both, either, and no flags.

PR #853, opened 2020-11-29
Allow override of marquez port in webpack.dev.js

wslulciuc:
This PR defaults the marquez API port to `8080` in `webpack.dev.js` and adds supports for overriding the port. For example,

```bash
$  MARQUEZ_PORT=5000 npm run dev
```

### DIFF
--- a/web/webpack.dev.js
+++ b/web/webpack.dev.js
@@ -14,7 +14,7 @@ const webpackDev = {
     },
     proxy: {
       '/api': {
-        target: `http://localhost:5000/`,
+        target: `http://localhost:${process.env.MARQUEZ_PORT || 8080}/`,
         secure: false,
         logLevel: 'debug',
         headers: {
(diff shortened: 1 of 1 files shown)


web/webpack.dev.js
```diff
@@ -14,7 +14,7 @@ const webpackDev = {
     },
     proxy: {
       '/api': {
-        target: `http://localhost:5000/`,
+        target: `http://localhost:${process.env.MARQUEZ_PORT || 8080}/`,
```
phixMe, line 17: I think it might make sense to include the host as well here if we want to go back to this approach.
```javascript
target: `http://${process.env.MARQUEZ_HOST}:${process.env.MARQUEZ_PORT || 8080}}/`,
```

wslulciuc: You got it 👍

phixMe (approved):
That works, thanks @wslulciuc

PR #856, opened 2020-12-01
Fix bug where all runs endpoint does not return all runs & job inputs not associated with run

henneberger:


wslulciuc (approved):
LGTM 👍

PR #873, opened 2020-12-15
Adding run indices

phixMe:
This adds indices for the following:
`runs`: `created_at:desc`, `current_run_state:asc`
`rubs`: `created_at:desc`


src/main/resources/marquez/db/migration/V15__index_created_at_and_current_run_state_on_runs.sql
```diff
@@ -0,0 +1,2 @@
+create index runs_created_at_current_run_state_index
+    on runs (created_at desc, current_run_state asc);
```
wslulciuc, line 2: Do we need to add `asc` to the `current_run_state` column?

wslulciuc (approved):
Left a minor comment, otherwise 💯 💯 💯

PR #880, opened 2020-12-29
Add open lineage endpoint

henneberger:


src/main/java/marquez/db/MapperUtils.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public static Set<String> getColumnNames(ResultSetMetaData metaData) {
+    try {
+      Set<String> columns = new HashSet<>();
+      for (int i = 1; i <= metaData.getColumnCount(); i++) {
+        columns.add(metaData.getColumnName(i));
+      }
+      return columns;
+    } catch (SQLException e) {
+      log.error("Unable to get column names", e);
```
julienledem, line 21: Why do we not throw here?

src/main/java/marquez/api/model/LineageEvent.java
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Requires jackson serialization features: mapper.registerModule(new JavaTimeModule());
+ * mapper.setSerializationInclusion(Include.NON_NULL);
+ * mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+ * mapper.disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE);
+ */
+@Valid
+@ToString
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class LineageEvent {
```
julienledem, line 25: Nice

src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -0,0 +1,431 @@
...
+      if (event.job.facets.sourceCodeLocation != null) {
+        if (event.job.facets.sourceCodeLocation.type != null) {
+          args.put("job.facets.sourceCodeLocation.type", event.job.facets.sourceCodeLocation.type);
+        }
+        if (event.job.facets.sourceCodeLocation.url != null) {
+          args.put("job.facets.sourceCodeLocation.url", event.job.facets.sourceCodeLocation.url);
+        }
+      }
+      if (event.job.facets.sql != null) {
+        args.put("job.facets.sql.query", event.job.facets.sql.query);
```
julienledem, line 383: should this just be "sql"

src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  default PGobject createJsonArray(List<LineageDataset> dataset, ObjectMapper mapper) {
+    try {
+      PGobject jsonObject = new PGobject();
+      jsonObject.setType("json");
+      jsonObject.setValue(mapper.writeValueAsString(dataset));
+      return jsonObject;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
julienledem, line 428: maybe add a message here? `"Can't write json datasets to db: " + dataset`

src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -0,0 +1,431 @@
...
+                runId)
+            .getBytes(UTF_8);
+    return UUID.nameUUIDFromBytes(bytes);
+  }
+
+  default String versionField(String fieldName, String type, String description) {
+    return VERSION_JOINER.join(fieldName, type, description);
+  }
+
+  default PGobject createJsonArray(List<LineageDataset> dataset, ObjectMapper mapper) {
```
julienledem, line 421: s/dataset/datasets ?

src/main/java/marquez/service/OpenLineageService.java
```diff
@@ -0,0 +1,40 @@
...
+            () ->
+                openLineageDao.createLineageEvent(
+                    event.eventType == null ? "" : event.eventType,
+                    event.eventTime.withZoneSameInstant(ZoneId.of("UTC")).toInstant(),
+                    event.run.runId,
+                    event.job.name,
+                    event.job.namespace,
+                    openLineageDao.createJsonArray(event.inputs, mapper),
+                    openLineageDao.createJsonArray(event.outputs, mapper),
+                    event.producer));
```
julienledem, line 36: why don't we pass the run and job facets as well?

src/main/resources/marquez/db/migration/V17__open_lineage.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE lineage_event (
+  event_time timestamp with time zone,
+  event_type text,
+  run_id text,
+  job_name text,
+  job_namespace text,
+  inputs jsonb,
+  outputs jsonb,
```
julienledem, line 8: don't we want to store the whole event as json?

src/test/java/marquez/OpenLineageIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    assertEquals(mapper.readTree(in), mapper.readTree(out));
+  }
+
+  public ObjectMapper getMapper() {
+    ObjectMapper mapper = new ObjectMapper();
+    mapper.setSerializationInclusion(Include.NON_NULL);
+    mapper.registerModule(new JavaTimeModule());
+    mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+    mapper.disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE);
```
julienledem, line 92: Shouldn't we get this from the MarquezContext?

src/test/resources/open_lineage/event_unicode.json
```diff
@@ -0,0 +1,28 @@
+{
+  "eventType": "COMPLETE",
+  "eventTime": "2020-12-28T19:52:00.001+10:00",
+  "run": {
+    "runId": "41fb5137-f0fd-4ee5-ba5c-56f8571d1bd7"
+  },
+  "job": {
+    "namespace": "my-scheduler-namespace",
+    "name": "myjob.buggy_task🕷"
```
julienledem, line 9: nice :)
I see 🕷 is 2 surrogates in utf-16 :)

julienledem (commented):
This looks good to me.
My main comment is on the the lineage_event table which seems to no be saving all the info in the event.

src/main/java/marquez/MarquezContext.java
```diff
@@ -134,49 +153,17 @@ private MarquezContext(
...
   }
 
-  public void registerListener(@NonNull RunTransitionListener listener) {
-    runTransitionListeners.add(listener);
-  }
-
-  public static Builder builder() {
-    return new Builder();
-  }
+  public ObjectMapper getLineageObjectMapper() {
```
wslulciuc, line 160: I would move creating a new instance of `ObjectMapper` for _OpenLineage_ to [`Utils`](https://github.com/MarquezProject/marquez/blob/main/src/main/java/marquez/common/Utils.java) as we keep things shared across boundaries (_resource_, _service_, etc) related to (de)serialization, etc under the `common` pkg.

We can also be explicit here and use the full name for _OpenLineage_ :)

```
Utils.newOpenLineageObjectMapper()
```

src/main/java/marquez/MarquezContext.java
```diff
@@ -134,49 +153,17 @@ private MarquezContext(
...
-    public Builder tags(@NonNull ImmutableSet<Tag> tags) {
-      this.tags = tags;
-      return this;
-    }
-
-    public Builder runTransitionListener(@NonNull RunTransitionListener runTransitionListener) {
-      return runTransitionListeners(Lists.newArrayList(runTransitionListener));
-    }
-
-    public Builder runTransitionListeners(
```
wslulciuc, line 172: I think both `runTransitionListener()` and `runTransitionListeners()` an be helpful. Do we want to remove them?

src/main/java/marquez/MarquezContext.java
```diff
@@ -39,46 +47,55 @@
 import marquez.service.models.Tag;
 import org.jdbi.v3.core.Jdbi;
 
+@Getter
 public final class MarquezContext {
-  @Getter private final NamespaceDao namespaceDao;
```
wslulciuc, line 43: For readability and consistency, mind if we keep instances variables annotated with `@Getter`? which is the approach taken for all of our [service models](https://github.com/MarquezProject/marquez/tree/main/src/main/java/marquez/service/models).

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -32,6 +33,8 @@
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import org.jdbi.v3.sqlobject.transaction.Transaction;
 
+@RegisterRowMapper(ExtendedDatasetRowMapper.class)
+@RegisterRowMapper(DatasetRowMapper.class)
```
wslulciuc, line 37: 👍

src/main/java/marquez/db/DatasetDao.java
```diff
@@ -144,9 +143,44 @@ void updateLastModifedAt(
...
+          + ":type, "
+          + ":now, "
+          + ":now, "
+          + ":namespaceUuid, "
+          + ":sourceUuid, "
+          + ":name, "
+          + ":name, "
+          + ":description) "
+          + "ON CONFLICT (namespace_uuid, name) "
+          // todo: what is the unique identifier here?
```
wslulciuc, line 171: A dataset is unique on `namespace_uuid`, `source_uuid`, `name` and `physical_name` (see [initial schema](https://github.com/MarquezProject/marquez/blob/main/src/main/resources/marquez/db/migration/V1__initial_schema.sql#L47))

src/main/java/marquez/api/model/LineageEvent.java
```diff
@@ -0,0 +1,219 @@
...
+
+  @Valid
+  @ToString
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  public static class DataSourceFacet extends BaseFacet {
+
+    public String name;
+    public String uri;
+  }
+}
```
wslulciuc, line 219: _Minor_: I think the `Lineage` prefix may be redundant for the inner classes in the event. For example, `LineageEvent.LineageDataset` and `LineageEvent.Dataset` both convey that the a dataset is part of the event.

src/main/java/marquez/api/model/LineageEvent.java
```diff
@@ -0,0 +1,219 @@
+package marquez.api.model;
```
wslulciuc, line 1: Mind moving the `LineageEvent` class to the [`service.models`](https://github.com/MarquezProject/marquez/tree/main/src/main/java/marquez/service/models) pkg to keep things consistent with how the resource layer currently interacts with the service layer?

src/test/java/marquez/AbstractIntegrationTest.java
```diff
@@ -0,0 +1,38 @@
...
+import io.dropwizard.testing.ConfigOverride;
+import io.dropwizard.testing.ResourceHelpers;
+import io.dropwizard.testing.junit.DropwizardAppRule;
+import java.net.URL;
+import marquez.client.MarquezClient;
+import marquez.client.Utils;
+import org.jdbi.v3.testing.JdbiRule;
+import org.junit.ClassRule;
+
+public abstract class AbstractIntegrationTest {
```
wslulciuc, line 12: Nice! Great idea to move common test setup to a base class other test suites can extend. Mind renaming the class to `BaseIntegrationTest` to best convey it's usage?

src/main/resources/marquez/db/migration/V17__open_lineage.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE lineage_event (
```
wslulciuc, line 1: Mind changes the table plural (`lineage_event` ) to be consistent with the other tables in the schema?

src/main/java/marquez/db/OpenLineageDao.java
henneberger: What about the other facets for the job context?

src/main/java/marquez/common/Utils.java
```diff
@@ -54,7 +56,10 @@ private Utils() {}
   public static ObjectMapper newObjectMapper() {
     final ObjectMapper mapper = Jackson.newObjectMapper();
     mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-    mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
+    mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+    mapper.disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE);
+    mapper.registerModule(new JavaTimeModule());
```
wslulciuc, line 61: We can omit registering `JavaTimeModule` as the `ObjectMapper` instance returned by `Jackson.newObjectMapper()`  configures it for us (see https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jackson/src/main/java/io/dropwizard/jackson/Jackson.java#L70).

src/main/java/marquez/db/mappers/ExtendedRunRowMapper.java
```diff
@@ -34,20 +37,26 @@
...
         timestampOrNull(results, Columns.NOMINAL_START_TIME),
         timestampOrNull(results, Columns.NOMINAL_END_TIME),
         stringOrNull(results, Columns.CURRENT_RUN_STATE),
-        timestampOrNull(results, Columns.STARTED_AT),
+        columnNames.contains(Columns.STARTED_AT)
+            ? timestampOrNull(results, Columns.STARTED_AT)
+            : null,
         uuidOrNull(results, Columns.START_RUN_STATE_UUID),
-        timestampOrNull(results, Columns.ENDED_AT),
+        columnNames.contains(Columns.ENDED_AT) ? timestampOrNull(results, Columns.ENDED_AT) : null,
```
wslulciuc, line 58: _Minor_: Wouldn't` timestampOrNull()` return `null` if not present?  Wondering if we need the extra check.

henneberger: timestampOrNull will throw an exception if the column doesn't exist

henneberger: A larger refactor would be to pass in the column names to every function, which i did in https://github.com/henneberger/marquez/blob/feature/openlineage/src/main/java/marquez/db/mappers/RunMapper.java 

We can bring that paradigm over

wslulciuc: Nice. That's definitely cleaner 👍

henneberger: Lets backlog it and we can migrate it in a different commit

wslulciuc:
@henneberger: This looks great! Looks like the build is failing because of formatting, but once fixed, we can merge 👍

wslulciuc (approved):
LGTM 👍 💯 🥇

src/main/java/marquez/db/OpenLineageDao.java
julienledem: I think we should add a job_facets column in the JobVersion or possibly a JobFacets table linked to the JobVersion.

wslulciuc: Might make sense to add a `job_facets` table as facets may not change from version to version (similar in approach to how we handle _job.contexts_)

PR #883, opened 2021-01-05
Fix in the bigquery input_tables missing

nizardeen:
Handled the BigQueryExtractor without returning it from None

### DIFF
--- a/integrations/airflow/marquez_airflow/extractors/bigquery_extractor.py
+++ b/integrations/airflow/marquez_airflow/extractors/bigquery_extractor.py
@@ -115,16 +115,29 @@ class BigQueryExtractor(BaseExtractor):
             context=context
         )]
 
+    def get_parsed_table_inputs(self,data):
+        data1 = re.sub(r'[?|`]',r'',data)
+        datas = re.split('[.]',str(data1))
+        table_dict = {"projectId":datas[0],"datasetId":datas[1],"tableId":datas[2]}
+        return table_dict
+
     def _get_input_from_bq(self, job, context, source, client):
-        if not job._properties.get('statistics')\
+        
+        if job._properties.get('statistics').get('query').get('referencedTables') is not None:
+            bq_input_tables = job._properties.get('statistics')\
+            .get('query')\
+            .get('referencedTables')
+
+        elif context['bigquery.sql.parsed.inputs'] is not None:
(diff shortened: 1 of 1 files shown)


julienledem:
Hi @nizardeen, can you share an example of the airflow task you are running to reproduce this?

julienledem:
FYI, when BigQuery caches the result, the input tables are not available in the API call.
The query cache is used only if you don't save the output of your query. (which happens if you are just testing things out and not saving the result)
See: https://cloud.google.com/bigquery/docs/cached-results#cache-exceptions
This can happen when testing things out but will not happen in production.

integrations/airflow/marquez_airflow/extractors/bigquery_extractor.py
```diff
@@ -115,16 +115,29 @@ def extract_on_complete(self, task_instance) -> [StepMetadata]:
...
+
     def _get_input_from_bq(self, job, context, source, client):
-        if not job._properties.get('statistics')\
+        
+        if job._properties.get('statistics').get('query').get('referencedTables') is not None:
+            bq_input_tables = job._properties.get('statistics')\
+            .get('query')\
+            .get('referencedTables')
+
+        elif context['bigquery.sql.parsed.inputs'] is not None:
```
wslulciuc, line 131: The main usage of the context key `bigquery.sql.parsed.inputs` is to compare the input and output tables returned via the `BigQuery` client call. For `BigQueryOperator`, we use the job details object (see example [here](https://github.com/MarquezProject/marquez/blob/main/integrations/airflow/tests/extractors/job_details.json)) returned by the client to set the in / out tables, so we wouldn't want to default to using the context here (though that might be an option in the future).

Note that the context key also helps to improve the [SQL parser](https://github.com/MarquezProject/marquez/blob/main/integrations/airflow/marquez_airflow/extractors/sql/experimental/parser.py) shipped with the `marquez-airflow` lib if they differ from wants returned by `BigQuery`

julienledem:
@nizardeen Is this still needed? The implementation would need to evolve.

wslulciuc:
Thanks for opening this PR, @nizardeen! For now, I'm going to close this PR as it's a bit out of date and looks like some clarity on the issue is still needed before we can merge. Feel free to reopen and/our reach out to us so we can iterate on your implementation.

Issue #887, opened 2021-01-07
Add endpoint to get a dataset versions


wslulciuc:
Add an endpoint that allows the caller to get a version for a dataset. For example:

## `GET` /namespaces/`{namespace}`/datasets/`{dataset}`/versions/`{version}`

```
{
  "id": {
    "namespace": STRING
    "name": STRING
  },
  "type": ENUM,
  "name": STRING,
  "physicalName": STRING,
  "createdAt": STRING,
  "version": STRING,
  "namespace": STRING,
  "sourceName": STRING,
  "fields": ARRAY,
  "tags": ARRAY
}
```

julienledem:
This was resolved by #888 and released under 0.12.0

PR #888, opened 2021-01-07
Add API to list versions for a dataset

### LINKED ISSUE #884: Add endpoint to list Dataset Versions
Add an endpoint that allows the caller to list version metadata for a given Dataset. For example:

## `GET` /namespaces/`{namespace}`/datasets/`{dataset}`/versions

```
{
  "versions": [
    {
      ... Dataset object as of that version
    }
  ]
}
```

wslulciuc:
This PR adds the following endpoints to the [DatasetAPI](https://marquezproject.github.io/marquez/openapi.html#tag/Datasets):

* **`GET`** /namespaces/`{namespace}`/datasets/`{dataset}`/versions/`{version}`
* **`GET`** /namespaces/`{namespace}`/datasets/`{dataset}`/versions

Fixes #884 #887


src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -97,7 +97,7 @@ default void updateOutputs(UUID versionUuid, UUID outputUuid) {
           + "WHERE uuid = :rowUuid")
   void updateLatestRun(UUID rowUuid, Instant updatedAt, UUID latestRunUuid);
 
-  static final String EXTENDED_SELECT =
+  String EXTENDED_SELECT =
```
julienledem, line 100: this should stay final.

src/main/java/marquez/service/models/DatasetVersion.java
```diff
@@ -0,0 +1,86 @@
...
+@ToString
+@JsonTypeInfo(
+    use = JsonTypeInfo.Id.NAME,
+    include = JsonTypeInfo.As.EXISTING_PROPERTY,
+    property = "type")
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = DbTable.class, name = "DB_TABLE"),
+  @JsonSubTypes.Type(value = Stream.class, name = "STREAM")
+})
+public abstract class DatasetVersion {
```
julienledem, line 47: should we use the lombok @Delegate annotation? DatasetVersion is Dataset + a version UUID

julienledem (commented):
I think we just need to update the dataset test case and we are good to go.

src/main/java/marquez/db/JobVersionDao.java
wslulciuc: Ahh, nice catch 👍

api/src/main/java/marquez/service/DatasetService.java
```diff
@@ -265,6 +277,26 @@ public boolean fieldExists(
     }
   }
 
+  public Optional<DatasetVersion> getVersion(@NonNull Version version) {
+    return datasetVersionDao.findBy(version.getValue()).map(this::toDatasetVersion);
+  }
+
+  public ImmutableList<DatasetVersion> getVersionsFor(
```
henneberger, line 284: I think an easier semantic to maintain would be that everything is assumed to be immutable unless explicitly labeled as mutable for semantic reasons (e.g. getMutableVersionsFor..). Tying to Guava ImmutableList is a bit awkward since Java also has immutable lists.

api/src/main/java/marquez/service/DatasetService.java
```diff
@@ -292,20 +324,67 @@ private Dataset toDataset(
...
+            .collect(toImmutableList());
+    final Run createdByRun =
+        datasetVersionRow
+            .getRunUuid()
+            .map(runUuid -> runService.getRun(RunId.of(runUuid)).orElse(null))
+            .orElse(null);
+    return Mapper.toDatasetVersion(datasetRow, tags, datasetVersionRow, fields, null, createdByRun);
+  }
+
+  private DatasetVersion toDatasetVersion(@NonNull ExtendedDatasetVersionRow datasetVersionRow) {
```
henneberger, line 361: This should be in the dao so this can be handled in one connection to put less contention of the connection pool, otherwise it will create n * ~4 connections per query. Since the limit is 100, it'd completely fill the connection pool if the full object is returned.

henneberger (approved):

api/src/main/java/marquez/service/DatasetService.java
wslulciuc: Agreed. Let's move towards having our services, models, etc use interfaces (in this case a `Map`) and not have a type defined in the guava lib as a return type.

api/src/main/java/marquez/service/DatasetService.java
wslulciuc: Yeah, working with the current `dao` APIs, this was the "easiest" approach (definitely not ideal though). We'll want to start pushing more and more of this logic down to the database which is outlined in your [Object Model Rearchitecture Proposal ](https://github.com/MarquezProject/marquez/issues/912). We'll want to start moving towards a model outlined in your proposal.

src/main/java/marquez/service/models/DatasetVersion.java
wslulciuc: The object does differ enough where I think a new model make sense. For example, a version has a `createdByRun` that will return metadata about the run that created the version, versions are also immutable so an `updatedAt` field won't be part of the response model.

That said, using `@Delegate` would be a really nice approach if you'd be able to filter with delegate methods get generate.

PR #891, opened 2021-01-13
Use api/src/ in Dockerfile

wslulciuc:
This PR fixes the docker build by using the path `api/src/` in `Dockerfile`.


henneberger (approved):
may be better to move to the api folder

julienledem (approved):

wslulciuc:
> may be better to move to the api folder

@henneberger: I agree, but we'll also have to move the gradle files to `api/src` as well to properly build the docker image with docker compose. It's something we can look into as follow up work.

PR #894, opened 2021-01-19
Add facets to openapi marquez spec

henneberger:
This proposal adds open lineage facets to the marquez API for jobs, datasets, and runs. Additionally, this deprecates the context parameter and convert contexts to a special context facet. This will allow cross compatibility between the existing marquez apis and open lineage.

See: https://github.com/MarquezProject/marquez/issues/895


spec/openapi.yml
```diff
@@ -1069,8 +1112,16 @@ components:
         args:
           description: The arguments of the run.
           type: object
+        facets:
+          $ref: '#/components/schemas/runFacets'
       example:
         args: {'email': 'me@example.com', 'emailOnFailure': false, 'emailOnRetry': true, 'retries': 1}
+        facets: {
+          'context': {
```
wslulciuc, line 1120: I would favor deprecating the job `context` field in a future release of Marquez. The context field had a similar goal as facets: _to allow for extra metadata to be sent along with job metadata not part of the core model_. I've opened an issue to start communicating the transition: https://github.com/MarquezProject/marquez/issues/910

spec/openapi.yml
```diff
@@ -1069,8 +1112,16 @@ components:
...
+            '_producer': 'https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client',
+            '_schemaURL': 'https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet'
```
wslulciuc, line 1122: Would a job have a `_schemaURL`?

spec/openapi.yml
```diff
@@ -1124,6 +1177,12 @@ components:
         endedAt: null
         durationMs: null
         args: {'email': 'me@example.com', 'emailOnFailure': 'false', 'emailOnRetry': 'true', 'retries': '1'}
+        facets: {
+          'context': {
```
wslulciuc, line 1181: Currently, a run doesn't contain the `context` field (though that was on the roadmap to add). We can omit the `context` in `facets`. (Similar comment for other `runAsRunning `,  `runAsCompleted`, etc)

spec/openapi.yml
```diff
@@ -1337,7 +1421,12 @@ components:
         endedAt: 2019-05-09T20:05:46.815920Z
         durationMs: 4250894125
         args: {'email': 'me@example.com', 'emailOnFailure': 'false', 'emailOnRetry': 'true', 'retries': '1'}
-
+        facets: {
+          'context': {
+            '_producer': 'https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client',
+            '_schemaURL': 'https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet'
```
wslulciuc, line 1427: Looking over the OpenLineage spec, should these facets b part of a run?

wslulciuc:
@henneberger : Should we also add facets to the _source_ and _namespace_ models? I think it'd make sense to have the core models of marquez contain the _facet_ field for extensibility.

spec/openapi.yml
julienledem: this is an attribute of the facet. All facets have pointer to their schema

spec/openapi.yml
julienledem: As this change is deprecating the context field for the future release of Marquez, it sounds like this is aligned?
By defining a `context` facet, it has to be nested under `facets`.

spec/openapi.yml
julienledem: But it does make sense to add it, right?
We do want to be able to add this information here. Are you suggesting an alternative?

spec/openapi.yml
julienledem: do you mean, adding a level of nesting?
To me, it makes sense to add the run facets as part of the run event here.

julienledem (approved):
This makes sense to me. I like that we live all facets be generic in this spec. Marquez treats all facets the same whether they are standard or custom.
I would advise against adding facets to source and namespace in that spec.
In openlineage, the namespace is just a mechanism to make unique name, the metadata is around jobs, datasets and runs.

spec/openapi.yml
wslulciuc: Why not just deprecate the `context`? We can communicate the release when the context support will be dropped. Having both `context` and `facets` to extend the metadata for the core models is confusing.

spec/openapi.yml
wslulciuc: Is this the schema for the job? I'm not understanding the  usage of `_schemaURL` here.

spec/openapi.yml
wslulciuc: Im not sure. Is `context` a facet?

spec/openapi.yml
wslulciuc: I see `_producer ` and `_schemaURL` are base facets within the spec.

spec/openapi.yml
wslulciuc: I see, I miss understood. We'll be defining the `context` as a _facet_ and deprecating the usage of `Job.context`. This makes sense 👍. I think some keys within the context can become a facet, for example the _sql_ for the job.

spec/openapi.yml
wslulciuc: The field `_schemaURL` points to the definition of the facet! Sorry, had to give the openlineage spec a quick review for clarity :)

spec/openapi.yml
wslulciuc: Based on how the `context` is defined for a job, this is considered a custom facet. Make sense to add it to a run as well 👍

spec/openapi.yml
wslulciuc: It wasn't clear on how `_producer` and `_schemaURL` were used within the facet (which I know have clarity on). Yep, adding `facets` to a run definitely makes sense here 👍

wslulciuc (approved):
Really excited to begin adding support for _facets_ to the marquez model 💯

PR #898, opened 2021-01-20
Add optional token to spark args

henneberger:
Airflow Dataproc operator doesn't have access to env variables so this removes them completely. Token is pass as a pseudo uri in the form of 'http://localhost:5000/api/v1/namespaces/ns_name/jobs/job_name/runs/ea445b5c-22eb-457a-8007-01c7c52b6e54?token=abc`.

The spark docs can be updated after that PR lands.

### DIFF
--- a/experimental/integrations/prototype-spark/marquez-spark-agent/src/main/java/marquez/spark/agent/ArgumentParser.java
+++ b/experimental/integrations/prototype-spark/marquez-spark-agent/src/main/java/marquez/spark/agent/ArgumentParser.java
@@ -20,7 +20,7 @@ public class ArgumentParser {
   private final String namespace;
   private final String jobName;
   private final String runId;
-  private final Optional<String> token;
+  private final Optional<String> apiKey;
 
   public static ArgumentParser parse(String agentArgs) {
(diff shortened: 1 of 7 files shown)


experimental/integrations/prototype-spark/marquez-spark-agent/src/main/java/marquez/spark/agent/ArgumentParser.java
```diff
@@ -0,0 +1,71 @@
...
+
+    log.info(
+        String.format("/api/%s/namespaces/%s/jobs/%s/runs/%s", version, namespace, jobName, runId));
+
+    return new ArgumentParser(host, version, namespace, jobName, runId, Optional.ofNullable(token));
+  }
+
+  private static String getToken(List<NameValuePair> nameValuePairList) {
+    String token;
+    if ((token = getNamedParameter(nameValuePairList, "token")) != null) {
```
wslulciuc, line 47: We'll want to use `key` as the query parameter as authentication with Marquez is via an API key

wslulciuc (approved):
LGTM 👍

PR #900, opened 2021-01-20
Expand spark facet definitions, make spark less risky to execute, integration tests.

henneberger:


api/src/test/java/marquez/BaseIntegrationTest.java
```diff
@@ -13,6 +21,7 @@
   protected static final String CONFIG_FILE = "config.test.yml";
   protected static final String CONFIG_FILE_PATH = ResourceHelpers.resourceFilePath(CONFIG_FILE);
   protected static final PostgresContainer POSTGRES = createMarquezPostgres();
+  HttpClient http2 = HttpClient.newBuilder().version(Version.HTTP_2).build();
```
wslulciuc, line 24: _Minor_: Would we also want to mark  `http2` as `static final`?

api/src/test/java/marquez/BaseIntegrationTest.java
```diff
@@ -35,4 +44,15 @@
   protected static PostgresContainer createMarquezPostgres() {
     return PostgresContainer.create("marquez");
   }
+
+  protected CompletableFuture<HttpResponse<String>> sendLineage(String body) {
```
wslulciuc, line 48: _Minor_: We'll eventually want to have the open lineage call in the `marquez-java` client (using the `openlineage-java` client), but make sense to wrap the lineage call in a method for now 👍 . Another option would be to add the call to a `HttpTestUtils` class. For example: `HttpTestUtils.post(endpoint, body)`

api/src/test/java/marquez/SparkAgentIntegrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+@Category(IntegrationTests.class)
+public class SparkAgentIntegrationTest extends BaseIntegrationTest {
```
wslulciuc, line 23: _Minor_: I feel the Marquez integration tests for spark would be more appropriate to have in the spark agent test suite which would allow us to 1) run a spark job in local mode 2) collect an open lineage event 3) query the Marquez to validate the metadata collected for the spark job (similar to our [approach with Airflow](https://github.com/MarquezProject/marquez/blob/main/integrations/airflow/tests/integration/integration.py)). But, make sense to have the integration test be part of the Marquez test suite for now 👍

api/src/test/java/marquez/SparkAgentIntegrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+        this.sendLineage(body)
+            .thenApply(HttpResponse::statusCode)
+            .whenComplete(
+                (val, error) -> {
+                  if (error != null) {
+                    fail("Could not complete request");
+                  }
+                });
+
+    assertEquals((Integer) 201, resp.join());
```
wslulciuc, line 49: Should we also assert the spark job metadata collected?

wslulciuc (approved):
Minor comments, but overall great work @henneberger! 💯

wslulciuc:
@henneberger : Looks like you'll need to address a few merge conflicts before we can merge :)

api/src/test/java/marquez/BaseIntegrationTest.java
henneberger: imo, package private is okay for test classes since there wouldn't really be an instance where this could be abused. I'll migrate it to the others modifiers for consistency.

api/src/test/java/marquez/BaseIntegrationTest.java
henneberger: Noted. We can resolve in the future when we want to migrate to http2-compatible clients to minimize duplicate code.

api/src/test/java/marquez/SparkAgentIntegrationTest.java
henneberger: I'm not sure if its possible to test using source code rather than rebuilt agent/marquez code using any other method than this due to dependencies on the build process, jvm versions (not java versions), and other library dependency conflicts.

wslulciuc: Yeah, you might be right here. When we promote the spark integration status from prototype we can revisit.

PR #901, opened 2021-01-21
Add graphql endpoint

henneberger:

### DIFF
--- a/api/src/main/java/marquez/MarquezApp.java
+++ b/api/src/main/java/marquez/MarquezApp.java
@@ -17,6 +17,7 @@ package marquez;
 import com.codahale.metrics.jdbi3.InstrumentedSqlLogger;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import io.dropwizard.Application;
+import io.dropwizard.assets.AssetsBundle;
 import io.dropwizard.configuration.EnvironmentVariableSubstitutor;
 import io.dropwizard.configuration.SubstitutingSourceProvider;
 import io.dropwizard.db.DataSourceFactory;
@@ -72,6 +73,8 @@ public final class MarquezApp extends Application<MarquezConfig> {
             new EnvironmentVariableSubstitutor(ERROR_ON_UNDEFINED)));
 
     bootstrap.getObjectMapper().disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+
+    bootstrap.addBundle(new AssetsBundle("/assets", "/api/v1/graphql-playground", "index.htm", "graphql-playground"));
   }
 
   @Override
@@ -91,6 +94,7 @@ public final class MarquezApp extends Application<MarquezConfig> {
       }
     }
     registerResources(config, env, source);
+
     registerServlets(env);
   }
 
@@ -106,6 +110,11 @@ public final class MarquezApp extends Application<MarquezConfig> {
 
     final MarquezContext context =
         MarquezContext.builder().jdbi(jdbi).tags(config.getTags()).build();
+
+    if (config.getGraphql().isEnabled()) {
+      env.servlets().addServlet("api/v1/graphql", context.getGraphqlServlet())
+          .addMapping("/api/v1/graphql", "/api/v1/schema.json");
+    }
 
     log.debug("Registering resources...");
     for (final Object resource : context.getResources()) {
--- a/api/src/main/java/marquez/MarquezConfig.java
+++ b/api/src/main/java/marquez/MarquezConfig.java
@@ -27,6 +27,7 @@ import marquez.service.models.Tag;
 public final class MarquezConfig extends Configuration {
   private static final boolean DEFAULT_MIGRATE_ON_STARTUP = true;
   private static final ImmutableSet<Tag> DEFAULT_TAGS = ImmutableSet.of();
+  @Getter private final GraphqlConfig graphql = new GraphqlConfig();
 
   @Getter private final boolean migrateOnStartup = DEFAULT_MIGRATE_ON_STARTUP;
   @Getter private final ImmutableSet<Tag> tags = DEFAULT_TAGS;
@@ -38,4 +39,8 @@ public final class MarquezConfig extends Configuration {
   @Getter
   @JsonProperty("flyway")
   private final FlywayFactory flywayFactory = new FlywayFactory();
+
+  public static class GraphqlConfig {
+    @Getter private final boolean enabled = true;
+  }
 }
--- a/api/src/main/java/marquez/graphql/Daos.java
+++ b/api/src/main/java/marquez/graphql/Daos.java
@@ -1,0 +1,126 @@
+package marquez.graphql;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import marquez.db.JobVersionDao.IoType;
+import marquez.graphql.mapper.ObjectMapMapper;
+import marquez.graphql.mapper.RowMap;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+@RegisterRowMapper(ObjectMapMapper.class)
+public interface Daos {
+  /*
+   * Note: Use must use a non-map type for returning single entries because a type of Map is already
+   * registered to jdbi.
+   */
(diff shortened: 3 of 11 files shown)


julienledem (commented):
Looks great!
I would add a test along the lines of [MarquezAppIntegrationTest](https://github.com/MarquezProject/marquez/blob/main/api/src/test/java/marquez/MarquezAppIntegrationTest.java)

api/src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,129 @@
...
+}
+type RunStateRecord {
+  transitionedAt: DateTime
+  state: String
+  run: Run
+}
+type RunArgs {
+  createdAt: DateTime
+  args: Json
+  checksum: String
```
julienledem, line 36: should we expose the checksum field?

api/src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,129 @@
...
+  run: [Run]
```
julienledem, line 37: I think we don't need to expose in the model that RunArgs is a separate entity. This is an implementation details.
It could be just a json field in Run

api/src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,129 @@
...
+  currentOwner: Owner
+  jobs: [Job]
+  datasets: [Dataset]
+}
+type JobVersion {
+  createdAt: DateTime
+  updatedAt: DateTime
+  location: String
+  version: UUID
+  jobContext: JobContext
```
julienledem, line 74: Same comment as for runargs.
I would embed the context here as string or json and simplify the schema.

api/src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,129 @@
...
+  versions: [JobVersion]
+  namespace: Namespace
+  currentVersion: JobVersion
+}
+type DatasetVersion {
+  createdAt: DateTime
+  dataset: Dataset
+  version: UUID
+  fields: [DatasetField]
+  run: Run
```
julienledem, line 101: this is the run that produced this version, maybe change to a more specific field name? "producedByRun"

api/src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,129 @@
...
+}
+type DatasetField {
+  type: String
+  createdAt: DateTime
+  updatedAt: DateTime
+  name: String
+  description: String
+  dataset: Dataset
```
julienledem, line 109: Do we need to be a ble to navigate back to a Dataset from a field? Should the field just be embeded in dataset?

api/src/main/java/marquez/MarquezApp.java
```diff
@@ -107,6 +113,12 @@ public void registerResources(
     final MarquezContext context =
         MarquezContext.builder().jdbi(jdbi).tags(config.getTags()).build();
 
+    if (config.getGraphql().isEnabled()) {
+      env.servlets()
+          .addServlet("api/v1/graphql", context.getGraphqlServlet())
```
wslulciuc, line 118: Mind using `/graphql` as the root? Looks like [_GraphQL takes a strong opinion on avoiding versioning_](https://graphql.org/learn/best-practices/#versioning).

build.gradle
```diff
@@ -78,6 +78,8 @@ dependencies {
   testCompile 'org.assertj:assertj-core:3.11.1'
   testCompile 'org.mockito:mockito-core:3.3.3'
   testCompile 'org.testcontainers:postgresql:1.13.0'
+
+  compile 'com.smoketurner.dropwizard:graphql-core:2.0.7-1'
```
wslulciuc, line 82: _Minor_: I would keep `compile`-level and `testCompile`-level grouped together in the build file.

api/src/main/java/marquez/MarquezApp.java
```diff
@@ -72,6 +73,10 @@ public void initialize(@NonNull Bootstrap<MarquezConfig> bootstrap) {
             new EnvironmentVariableSubstitutor(ERROR_ON_UNDEFINED)));
 
     bootstrap.getObjectMapper().disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+
+    bootstrap.addBundle(
+        new AssetsBundle(
+            "/assets", "/api/v1/graphql-playground", "index.htm", "graphql-playground"));
```
wslulciuc, line 79: Similar comment as below, but I'd suggest using the mapping `/graphql-playground` at the root-level for the graphql playground.

api/src/main/java/marquez/MarquezConfig.java
```diff
@@ -38,4 +39,8 @@
   @Getter
   @JsonProperty("flyway")
   private final FlywayFactory flywayFactory = new FlywayFactory();
+
+  public static class GraphqlConfig {
```
wslulciuc, line 43: Moving the graphql configuration to the `graphql` package will help group all graphql related code.

api/src/main/java/marquez/MarquezConfig.java
```diff
@@ -38,4 +39,8 @@
...
+    @Getter private final boolean enabled = true;
```
wslulciuc, line 44: Should we disable graphql by default?

api/src/main/java/marquez/graphql/Daos.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+import java.util.UUID;
+import marquez.db.JobVersionDao.IoType;
+import marquez.graphql.mapper.ObjectMapMapper;
+import marquez.graphql.mapper.RowMap;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+@RegisterRowMapper(ObjectMapMapper.class)
+public interface Daos {
```
wslulciuc, line 12: _Minor_: I'm wondering if `GraphqlDao` would be a better name to best express the underlying quries?

api/src/main/java/marquez/graphql/MarquezGraphqlServletBuilder.java
```diff
@@ -0,0 +1,226 @@
...
+import graphql.schema.idl.SchemaParser;
+import graphql.schema.idl.TypeDefinitionRegistry;
+import java.net.URL;
+import java.time.ZonedDateTime;
+import java.util.Map;
+import java.util.UUID;
+import lombok.SneakyThrows;
+import org.jdbi.v3.core.Jdbi;
+
+public class MarquezGraphqlServletBuilder {
```
wslulciuc, line 31: _Minor_: We'll want to be consistent with lowercase vs uppercase in graphql class names. I'd favor all lowercase :)

api/src/main/java/marquez/graphql/MarquezGraphqlServletBuilder.java
```diff
@@ -0,0 +1,226 @@
...
+
+  public GraphQLHttpServlet getServlet(final Jdbi jdbi) {
+    final GraphQLFactory factory = getGraphQLFactory(jdbi);
+
+    final PreparsedDocumentProvider provider =
+        new CachingPreparsedDocumentProvider(factory.getQueryCache(), new MetricRegistry());
```
wslulciuc, line 37: Does this get registered with the global metric registry? You may have to pass in `Environment` and access the metric registry via `env.metrics()` see, https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/MarquezApp.java#L80

api/src/main/java/marquez/MarquezConfig.java
wslulciuc: I would add support to enable / disable graphql in [`marquez.yml`](https://github.com/MarquezProject/marquez/blob/main/marquez.example.yml). So, something like:

```yaml
graphql:
  enabled: ${GRAPHQL_ENABLED:-true}
```

wslulciuc (approved):
I'd make sure users can enable / disable the graphql endpoint via config, once supported, we can merge 👍

api/src/main/java/marquez/MarquezApp.java
wslulciuc: Look over the docs it wasn't made obviously clear on the best approach to version graphql endpoints (if at all). Thanks for insight 👍

PR #911, opened 2021-01-30
Assure safety for open lineage spark calls, fix open lineage mapping to marquez

henneberger:


experimental/integrations/marquez-spark-agent/src/main/java/marquez/spark/agent/ArgumentParser.java
```diff
@@ -37,7 +39,8 @@ public static ArgumentParser parse(String agentArgs) {
     String apiKey = getApiKey(nameValuePairList);
 
     log.info(
-        String.format("/api/%s/namespaces/%s/jobs/%s/runs/%s", version, namespace, jobName, runId));
+        String.format(
+            "%s/api/%s/namespaces/%s/jobs/%s/runs/%s", host, version, namespace, jobName, runId));
```
julienledem, line 43: the `argument` may not provide the runId in the current state?

julienledem (approved):
LGTM

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -101,7 +114,7 @@ void createLineageEvent(
       String producer);
 
   @Transaction
-  default void updateMarquezModel(LineageEvent event) {
+  default UpdateLineageResponse updateMarquezModel(LineageEvent event) {
```
wslulciuc, line 117: Do we want a service to return a db model? Meaning, shouldn't the service define its own model to return and not depend on a model defined by a layer it encapsulates?

experimental/integrations/marquez-spark-agent/integrations/sparkrdd/1.json
```diff
@@ -10,8 +10,8 @@
     "name" : "job_name"
   },
   "inputs" : [ {
-    "namespace" : "ns_name",
-    "name" : "data.txt"
+    "namespace" : "gs.bucket",
+    "name" : "gs://bucket/data.txt"
```
wslulciuc, line 14: A dataset also has a `type` we'll need to set, will that be parsed from the `namespace` or `name`?

henneberger: The type is the scheme part of the URI

api/src/main/java/marquez/db/OpenLineageDao.java
henneberger: I'll adjust it to be the bag of objects like we discussed before and just do the unboxing in the returning function call.

api/src/main/resources/marquez/db/migration/V19__alter_run_to_add_original_name.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE runs ADD original_name VARCHAR(255);
```
wslulciuc, line 1: Is there a description on the usage of `original_name`?

henneberger: OpenLineage uses strings as its run id. This is just for cataloging to retain the original name after we upconverted it to uuid.

The openlineage spec treats the run id as locally unique to the job which also raises some issues with backporting to the marquez API.

wslulciuc: Make sense. I would use `external_id` as the column name since it's an external run ID (_string_, _int_, etc) collected from another system.

wslulciuc (approved):
Great work, @henneberger! Minor comment on updating the column name for the external run ID added to the `runs` table. Good to merge once updated 👍

wslulciuc:
@henneberger: One thing to note is that with the progress of OpenLineage implementation, we are seeing duplication in some logic already part of the  core services of marquez. We'll want to begin scoping out ways to make sharing functionality between the two. All expand a bit more and comment in your [Object Model Rearchitecture Proposal](https://github.com/MarquezProject/marquez/issues/912).

api/src/main/resources/marquez/db/migration/V19__alter_run_to_add_original_name.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE runs ADD external_id VARCHAR(255);
```
wslulciuc, line 1: _Minor_: Looks like the file name also needs an update :)

PR #916, opened 2021-02-03
Make conn url optional and remove creds is present

wslulciuc:
This PR makes the following changes to the [SourceAPI](https://marquezproject.github.io/marquez/openapi.html#tag/Sources) to simplify how metadata for `sources` are collected:

* Make `Source.type` a string
* Make `Source.connectionUrl` _optional_. Not all sources have a connection URL (ex: `BigQuery`)
* Remove credentials from `Source.connectionUrl` if present. For example `postgresql://me:1234@localhost:5431/mydb` has the username / password set to **`me`** / **`234`**. In Marquez, the URL will now be stored as: `postgresql://localhost:5431/mydb`
* Parse source type from `Source.connectionUrl`. For example, `postgresql://localhost:5431/mydb` will have a source type `POSTGRESQL`, `mysql://localhost:5431/mydb` will have a source type `MSQL`, etc

This PR also adds support to update source metadata!

### Example API Call

In the example below, the `Source.type` is **not** required since we have set `Source.connectionUrl` which will be used instead to get the type.

```
# Create a source
curl -X "PUT" "http://localhost:8080/api/v1/sources/my-source" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{
  "connectionUrl": "postgresql://me:1234@localhost:5431/mydb",
  "description": "My first source!"
}'
```
```
{
  "type": "POSTGRESQL",
  "name": "my-source",
  "createdAt": "2021-02-03T07:47:34.679121Z",
  "updatedAt": "2021-02-03T09:16:30.241900Z",
  "connectionUrl": "postgresql://localhost:5431/mydb",
  "description": "My first source!"
}
```


api/src/main/java/marquez/common/Utils.java
```diff
@@ -112,7 +117,46 @@ public static UUID toUuid(@NonNull final String uuidString) {
...
+  static boolean isJdbcUrl(@NonNull final URI connectionUrl) {
+    final String connectionUrlString = connectionUrl.toString();
+    return connectionUrlString.startsWith(JDBC);
+  }
+
+  static URI toUrlWithNoJdbcPrefix(@NonNull final URI connectionUrl) {
+    return isJdbcUrl(connectionUrl)
+        ? URI.create(dropJdbcPrefix(connectionUrl.toString()))
+        : connectionUrl;
+  }
```
henneberger, line 157: This would fail with `jdbc:microsoft:sqlserver://` as a prefix. This is also an issue with the openlineage code location code so we may benefit from our own parser class.

api/src/main/java/marquez/common/Utils.java
```diff
@@ -112,7 +117,46 @@ public static UUID toUuid(@NonNull final String uuidString) {
...
-  public static Instant toInstant(@Nullable final String asIso) {
+  public static Instant toInstantOrNull(@Nullable final String asIso) {
     return (asIso == null) ? null : Instant.from(ISO_INSTANT.parse(asIso));
   }
+
+  public static SourceType sourceTypeFor(@NonNull final URI connectionUrl) {
+    return SourceType.of(toUrlWithNoJdbcPrefix(connectionUrl).getScheme());
+  }
+
+  public static URI urlWithNoCredentials(@NonNull final URI connectionUrl) {
```
henneberger, line 128: the name of the function doesn't match its behavior, I wouldn't expect it to modify jdbc specific urls.

api/src/main/java/marquez/common/models/SourceType.java
```diff
@@ -14,13 +14,42 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+
+@EqualsAndHashCode
+@ToString
+@JsonDeserialize(converter = SourceType.FromValue.class)
+@JsonSerialize(converter = SourceType.ToValue.class)
+public final class SourceType {
```
henneberger, line 31: What is the convention for using classes like this over conventional strings that are associated with typed objects?

api/src/main/java/marquez/common/models/SourceType.java
```diff
@@ -14,13 +14,42 @@
...
+  @Getter private final String value;
+
+  public SourceType(@NonNull final String value) {
+    this.value = checkNotBlank(value, "value must not be blank").toUpperCase().trim();
```
henneberger, line 35: I don't see the benefit from uppercasing their data. Why would we modify the users data? It can be frustrating to work with systems that change your data.

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -283,12 +283,22 @@ default void upsertLineageDataset(
...
           sourceDao.upsert(
               UUID.randomUUID(),
-              getSourceType(ds),
+              getSourceType(ds).getValue(),
+              now,
               now,
               ds.getFacets().getDataSource().getName(),
-              ds.getFacets().getDataSource().getUri());
+              ds.getFacets().getDataSource().getUri(),
+              ds.getFacets().getDescription());
```
henneberger, line 291: This is a bug in the open lineage implementation! There shouldn't be a description field here. That must have accidentally been carried over from porting the openapi spec.

api/src/main/java/marquez/db/SourceDao.java
```diff
@@ -69,19 +47,29 @@
...
-          + "connection_url "
+          + "connection_url, "
+          + "description "
           + ") VALUES ("
           + ":uuid, "
           + ":type, "
-          + ":now, "
-          + ":now, "
+          + ":createdAt, "
+          + ":updatedAt, "
```
henneberger, line 56: Why add a createdAt field? The only instance I can think of is if you want to create a record and set an updated time to be different than when you created it. That doesn't seem like a valid use case to me.

api/src/main/java/marquez/service/SourceService.java
```diff
@@ -17,96 +17,70 @@
...
 
   static Source toSource(@NonNull final SourceRow row) {
     return new Source(
-        SourceType.valueOf(row.getType()),
+        SourceType.of(row.getType()),
         SourceName.of(row.getName()),
         row.getCreatedAt(),
         row.getUpdatedAt(),
-        URI.create(row.getConnectionUrl()),
+        row.getConnectionUrl().map(URI::create).orElse(null),
```
henneberger, line 83: why unbox optionals?

henneberger: This can lead to sneaky exceptions on GET operations which can be an issue if someone wanted to integrate with marquez directly

api/src/main/java/marquez/service/models/SourceMeta.java
```diff
@@ -14,18 +14,50 @@
...
 import marquez.common.models.SourceType;
 
-@Value
-public class SourceMeta {
-  @NonNull SourceType type;
-  @NonNull URI connectionUrl;
-  @Nullable String description;
+@EqualsAndHashCode
+@ToString
+public final class SourceMeta {
```
henneberger, line 32: Why is this class final? It seems like we would want people to be able to extend the meta class so it can meet their usecase.

api/src/main/java/marquez/service/models/SourceMeta.java
```diff
@@ -14,18 +14,50 @@
...
+  }
+
+  @JsonCreator
+  public SourceMeta(
+      @Nullable final SourceType type,
+      @Nullable final URI connectionUrl,
+      @Nullable final String description) {
+    this.type =
+        (connectionUrl == null)
+            ? checkNotNull(type, "connection url must be provided when type == null")
```
henneberger, line 52: The intended semantics isn't really that obvious. What happens if the provided connectionurl isn't parsable or can't be mapped to a type?

api/src/test/java/marquez/common/UtilsTest.java
```diff
@@ -98,6 +104,77 @@ public void testChecksumFor_notEqual() {
...
+    assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> Utils.toUuid(""));
+    assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> Utils.toUuid(" "));
+  }
+
+  @Test
+  public void testToUuid_throwsOnLengthNotValid() {
+    assertThatExceptionOfType(IllegalArgumentException.class)
+        .isThrownBy(() -> Utils.toUuid("53678291-e1fa-4f1b-bb29-939b-53a855c48817"));
+    assertThatExceptionOfType(IllegalArgumentException.class)
+        .isThrownBy(() -> Utils.toUuid("53678291-e1fa-4f1b-53a855c48817"));
```
henneberger, line 125: Why create our own function to parse uuids when the UUID class has a parser?

api/src/test/java/marquez/common/UtilsTest.java
```diff
@@ -98,6 +104,77 @@ public void testChecksumFor_notEqual() {
...
+  }
+
+  @Test
+  public void testToInstantOrNull() {
+    final Instant nullInstant = Utils.toInstantOrNull(null);
+    assertThat(nullInstant).isNull();
+
+    final String asIso = "2021-01-09T19:49:24.201Z";
+    final Instant instant = Utils.toInstantOrNull(asIso);
```
henneberger, line 134: Jackson should be handling Instant behavior. Invalid requests should fail at the resource layer so I don't see the need for this.

api/src/test/java/marquez/common/models/ModelGenerator.java
```diff
@@ -38,50 +38,29 @@ public static OwnerName newOwnerName() {
...
-        break;
-      default:
-        throw new IllegalArgumentException();
+    if ("POSTGRESQL".equals(type.getValue())) {
+      return URI.create("jdbc:postgresql://localhost:5432/test" + newId());
+    } else if ("KAFKA".equals(type.getValue())) {
+      return URI.create("kafka://localhost:9092");
     }
-    return URI.create(connectionUrlString);
+    throw new IllegalArgumentException();
```
henneberger, line 63: Should return a uri with the type as the scheme by default to reduce test burden

api/src/test/java/marquez/db/SourceDaoTest.java
```diff
@@ -35,73 +41,133 @@
 
 @Category({DataAccessTests.class, IntegrationTests.class})
 public class SourceDaoTest {
+  private static final UUID ROW_UUID = UUID.randomUUID();
+  private static final Instant NOW = Instant.now();
+  private static final SourceType TYPE = newDbSourceType();
+  private static final SourceName NAME = newSourceName();
+  private static final URI CONNECTION_URL = newConnectionUrlFor(TYPE);
+  private static final String DESCRIPTION = newDescription();
```
henneberger, line 49: This can all be stored in a SourceRow so its typed and easier to identify usage of Source models

api/src/test/java/marquez/db/SourceDaoTest.java
```diff
@@ -35,73 +41,133 @@
...
   @Test
   public void testFindAll() {
-    final List<SourceRow> newRows = newSourceRows(4);
-    newRows.forEach(newRow -> sourceDao.insert(newRow));
-
-    final List<SourceRow> rows = sourceDao.findAll(4, 0);
+    final SourceName name0 = newSourceName();
+    final SourceName name1 = newSourceName();
+    final SourceName name2 = newSourceName();
+    final SourceName name3 = newSourceName();
```
henneberger, line 154: Should one of these names have a duplicate name to test the unique semantic?

api/src/main/java/marquez/common/models/SourceType.java
wslulciuc: Mainly, it's an alternative to working with just the literal _string_ and provides the added context to method calls that can be self-documenting. For example:

```
// Using  literal strings
Source(String, String ,...)

// Using defined types
Source(SourceType, SourceName, ...)
```

For now, we have a simple check that the _value_ for a `SourceType` can not be _null_, or _blank_ but this can be expanded to check for allowed length, characters, etc (similar to [NamespaceName](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/common/models/NamespaceName.java)). You also get that extra assurance that you are working with a valid value when calling `SourceType.getValue()`.

Of course, not everything has to be a type (ex: the description field for most of our models) and it's more of a case-by-case basis.

api/src/main/java/marquez/common/models/SourceType.java
wslulciuc: My initial thought was to keep types upper case to be consistent with types in other models (`Dataset`, `Job`), but given that this is a user provided field, and not a field that is set / owned by Marquez, makes sense to not modify it.

api/src/test/java/marquez/common/models/ModelGenerator.java
wslulciuc: Sure, that sounds like a reasonable default.

wslulciuc:
I think being able to parse the connection URL correctly is important before this gets merged. I moved this branch to _in progress_ to have more time to work on the logic and not block our `0.12.0` release. I've also moved the dropping of the `Source.type` enum to #919

api/src/main/java/marquez/common/models/SourceType.java
henneberger: I'm fine with this pattern if we can move to a common object model and use visitor patterns to transform our objects between contexts. Currently its a bit onerous to have to always box and unbox String values to call functions.

julienledem:
Is this PR now outdated?

wslulciuc:
> Is this PR now outdated?

Yeah, though some of the features we'd still want to make. For example, removing the username / password from the connection URL. I'll close for now, though @henneberger's feedback is still very much relevant and can be referenced when revisiting this.

PR #921, opened 2021-02-04
Use user provided dataset names for open lineage

henneberger:
This removes the pretty function to convert dataset URI names to a dot separated names. Since URIs have mixed specs, e.g.`jdbc:oracle:thin:@localhost:1521:orclpdb1`, the implementation for prettifying names is incomplete. Instead of needing to backfill the dataset names in the future, which would be a difficult endeavor, we should just accept names as-is and introduce a display name for UI related rendering.


wslulciuc (approved):
LGTM 👍

PR #934, opened 2021-02-09
Denormalize namespace name to job table

henneberger:
The namespace uuid column and constraint can be dropped, and a new constraint on namespace name can be added in the next release


julienledem (approved):

api/src/main/resources/marquez/db/migration/V20__alter_jobs_to_add_namespace_name.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE jobs ADD namespace_name VARCHAR(255);
```
wslulciuc, line 1: With []() now merged, mind updating the migration version to `V22`?

wslulciuc (changes requested):
Looks good 👍 But, we'll need to update the migration version to `V22` to maintain ordering.

api/src/main/resources/marquez/db/migration/V20__alter_jobs_to_add_namespace_name.sql
henneberger: pr #935 was branched off of this commit so this pr can be closed

wslulciuc: 👍

PR #1007, opened 2021-02-24
Create job version record only when inputs/outputs are fully known

henneberger:


api/src/main/java/marquez/db/RunStateDao.java
```diff
@@ -15,71 +15,40 @@
...
+    RunStateRow runStateRow = upsert(UUID.randomUUID(), transitionedAt, runUuid, runState);
+    runDao.updateRunState(runUuid, transitionedAt, runState);
+    if (runState.isDone()) {
+      runDao.updateEndState(runUuid, transitionedAt, runStateRow.getUuid());
+    } else if (runState.isStarting()) {
+      runDao.updateStartState(runUuid, transitionedAt, runStateRow.getUuid());
+    }
+    //    if (complete && outputVersionUuids != null && outputVersionUuids.size() > 0) {
+    //      createDatasetDao().updateLastModifedAt(outputVersionUuids, updateAt);
+    //    }
```
henneberger, line 52: what is the logic supposed to be for last modified at?

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -178,11 +115,15 @@ default void updateJobVersionUuid(UUID rowUuid, Instant updatedAt, UUID jobVersi
           + ":runArgsUuid, "
           + ":nominalStartTime, "
           + ":nominalEndTime, "
-          + ":runStateType "
+          + ":runStateType,"
+          + ":runStateTime, "
+          + ":namespaceName, "
+          + ":jobName "
```
henneberger, line 121: todo: add location

api/src/main/java/marquez/api/DatasetResource.java
```diff
@@ -238,6 +245,12 @@ void throwIfNotExists(@NonNull NamespaceName namespaceName, @NonNull DatasetName
     }
   }
 
+  void throwIfSourceNotExists(SourceName sourceName) throws MarquezServiceException {
+    if (!sourceDao.exists(sourceName.getValue())) {
```
wslulciuc, line 249: We'll want to use [`SourceService.exists()`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/service/SourceService.java#L69) that way the resource layer only depends on the service layer.

api/src/main/java/marquez/api/JobResource.java
```diff
@@ -84,7 +92,12 @@ public Response createOrUpdate(
       @Valid JobMeta jobMeta)
       throws MarquezServiceException {
     throwIfNotExists(namespaceName);
-    jobMeta.getRunId().ifPresent(this::throwIfNotExists);
+    if (jobMeta.getRunId().isPresent()) {
```
wslulciuc, line 95: 👍

api/src/main/java/marquez/api/JobResource.java
```diff
@@ -213,6 +226,28 @@ void throwIfNotExists(@NonNull RunId runId) throws MarquezServiceException {
     }
   }
 
+  private void throwIfJobDoesNotMatchRun(RunId runId, String namespaceName, String jobName) {
+    Optional<Run> runRow = runService.getRun(runId);
```
wslulciuc, line 230: _Minor_: Here we are interactive with a `Run` object instead of a row and can adjust the naming to reflect that.

api/src/main/java/marquez/api/JobResource.java
```diff
@@ -213,6 +226,28 @@ void throwIfNotExists(@NonNull RunId runId) throws MarquezServiceException {
...
+      throw new RunNotFoundException(runId);
+    }
+  }
+
+  private void throwIfDatasetsNotExist(ImmutableSet<DatasetId> datasets) {
+    if (datasets == null) {
+      return;
+    }
+    for (DatasetId datasetId : datasets) {
+      if (!datasetDao.exists(datasetId.getNamespace().getValue(), datasetId.getName().getValue())) {
```
wslulciuc, line 245: I would add the method `DatasetService.exists(DatasetId)` to ensure we properly encapsulate the dao layer behind service calls.

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -228,10 +183,87 @@ ExtendedRunRow upsert(
...
           + "VALUES (:runUuid, :datasetVersionUuid) ON CONFLICT DO NOTHING")
   void updateInputMapping(UUID runUuid, UUID datasetVersionUuid);
+
+  @Transaction
+  default void notifyJobChange(UUID runUuid, JobRow jobRow, JobMeta jobMeta) {
+    DatasetDao datasetDao = createDatasetDao();
+
+    upsertRun(runUuid, jobRow.getName(), jobRow.getNamespaceName());
+
+    if (jobMeta.getInputs() != null) {
```
julienledem, line 203: does this need to deal with outputs too?

api/src/main/java/marquez/db/RunStateDao.java
julienledem: I would think that it matches the end time of the last run the had this dataset as an output?

julienledem: @wslulciuc ?

julienledem (approved):
This looks great.
Just a few comments/questions below

api/src/main/java/marquez/db/JobDao.java
```diff
@@ -91,6 +78,53 @@
...
+  }
+
+  default PGobject toJson(Set<DatasetId> dataset, ObjectMapper mapper) {
+    try {
+      PGobject jsonObject = new PGobject();
+      jsonObject.setType("json");
+      jsonObject.setValue(mapper.writeValueAsString(dataset));
+      return jsonObject;
+    } catch (Exception e) {
+      return null;
```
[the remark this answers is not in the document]
henneberger: This just tracks what job inputs/outputs the client expected to have for their job but we don't actually use this data for anything. I'd rather just suppress any exceptions or not include it all rather than fail.

api/src/main/java/marquez/db/JobDao.java
```diff
@@ -123,5 +169,9 @@ JobRow upsert(
       UUID namespaceUuid,
       String namespaceName,
       String name,
-      String description);
+      String description,
+      UUID jobContextUuid,
+      String location,
+      Object inputs,
+      Object outputs);
```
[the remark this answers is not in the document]
henneberger: I'll migrate to PGobject

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -174,26 +175,32 @@ default UpdateLineageRow updateMarquezModel(LineageEvent event) {
               runUuid,
               event.getRun().getRunId(),
               now,
-              jobVersion.getUuid(),
+              null,
```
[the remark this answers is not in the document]
henneberger: yeah, the logic moved to `updateMarquezOnComplete`

api/src/main/java/marquez/db/RunDao.java
henneberger: No, there is a comment on the function above it that communicates why. There is an alternative route to registering /output/ datasets in the dataset api. The run service only registers inputs. I can try to spruce this up so its not so confusing.

api/src/main/java/marquez/db/JobDao.java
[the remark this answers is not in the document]
wslulciuc: It might be helpful to log a warning?

api/src/main/java/marquez/db/RunStateDao.java
wslulciuc: @henneberger: When a job reaches a `COMPLETE` state, there's a reasonable assumption that the output datasets have been modified. We can set `modifedAt` equal to `transitionedAt`

api/src/main/resources/marquez/db/migration/V24__alter_jobs.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE jobs ADD job_context_template_uuid UUID;
+ALTER TABLE jobs ADD location_template varchar;
+ALTER TABLE jobs ADD inputs_template JSONB;
+ALTER TABLE jobs ADD outputs_template JSONB;
```
wslulciuc, line 4: It think want to prefix the columns with `current_*`. For example:

```
current_job_context_uuid
current_ location
etc
```

wslulciuc (approved):
Left some comments (mainly just the columns names added to the `jobs` table), but we're good to merge! Nice work on reduce code complexity 💯

api/src/main/resources/marquez/db/migration/V24__alter_jobs.sql
wslulciuc: Agreed. Maybe for column names / when we modify the data model we can open an issue to quickly decide on naming beforehand?

wslulciuc: Also, my fault for providing late feedback 😬

api/src/main/java/marquez/db/JobDao.java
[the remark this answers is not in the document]
henneberger: Dao classes can't attach an slf4j logger

PR #1013, opened 2021-02-24
helm chart: fix configmap name mismatch

ecerulm:
Current helm chart will only deploy correctly if the release name is exactly `marquez`. 

For example, if release is `marquez-test` the `ConfigMap` will be named `marquez-test-config`  and the `Deployment` still looks for `marquez-config`.


wslulciuc:
Superseded by https://github.com/MarquezProject/marquez/pull/1019

PR #1020, opened 2021-02-26
Import Spark DatasetMetricFacet facet

collado-mike:
Imported the proposed DatasetMetricFacet definition and added the Spark metric collection implementation. This implementation only captures row count from the spark plan, but the foundation is here to collect more metrics in the future.


henneberger (approved):

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -142,14 +171,20 @@ protected String getEventType(JobResult jobResult) {
...
+          DatasetMetricFacet metricFacet) {
     Map<String, Object> additionalFacets = new HashMap<>();
     if (logicalPlanFacet != null) {
       additionalFacets.put("spark.logicalPlan", logicalPlanFacet);
     }
     if (jobError != null) {
       additionalFacets.put("spark.exception", jobError);
     }
+    if (metricFacet != null) {
+      additionalFacets.put("dataset.metrics", metricFacet);
```
julienledem, line 186: This facet should be an outputFacet on the output dataset rather than on the run.

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -95,6 +103,26 @@ private ParentRunFacet buildParentFacet() {
...
+                        }
+                      });
+      return DatasetMetricFacet.builder()
+              .metricValues(
+                      ImmutableMap.of("rows", DatasetMetricFacet.Metric.count(count.longValue())))
+              .build();
+
+    } else {
+      return null;
+    }
```
julienledem, line 123: I think we should make a facet that is specific to this particular data. 
A generic data quality facet is also useful but used in a different context in complement of this.
How about: 
```
OutputSizeFacet: {
   rowCount: 123,
   ...
}                   
```
Also the logicalPlan probably is an estimate for the planner when the physical plan after execution would probably have the actual count.

julienledem (approved):

PR #1021, opened 2021-02-26
Don't set run state update time to state transition time

wslulciuc:
This PR fixes the timestamp used to set the `updatedAt` column for the table `runs`. It now uses the current time. Marquez provides the option to override the state transition time for a run by the caller. This transition time is also used to also set `runs.updatedAt`, which is incorrect.

### DIFF
--- a/api/src/main/java/marquez/db/RunStateDao.java
+++ b/api/src/main/java/marquez/db/RunStateDao.java
@@ -52,7 +52,7 @@ public interface RunStateDao extends SqlObject {
                 .execute());
     // State transition
     final Instant updateAt = row.getTransitionedAt();
-    createRunDao().updateRunState(row.getRunUuid(), updateAt, row.getState());
+    createRunDao().updateRunState(row.getRunUuid(), Instant.now(), row.getState());
     if (starting) {
       createRunDao().updateStartState(row.getRunUuid(), updateAt, row.getUuid());
     }
--- a/api/src/main/java/marquez/db/RunStateDao.java
+++ b/api/src/main/java/marquez/db/RunStateDao.java
@@ -52,7 +52,6 @@ public interface RunStateDao extends SqlObject {
                 .execute());
     // State transition
     final Instant updateAt = row.getTransitionedAt();
-    createRunDao().updateRunState(row.getRunUuid(), Instant.now(), row.getState());
     if (starting) {
       createRunDao().updateStartState(row.getRunUuid(), updateAt, row.getUuid());
     }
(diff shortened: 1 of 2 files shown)


api/src/main/java/marquez/db/RunStateDao.java
```diff
@@ -52,7 +52,7 @@ default void insert(
                 .execute());
     // State transition
     final Instant updateAt = row.getTransitionedAt();
-    createRunDao().updateRunState(row.getRunUuid(), updateAt, row.getState());
+    createRunDao().updateRunState(row.getRunUuid(), Instant.now(), row.getState());
```
henneberger, line 55: Why ignore the user provided transition time? They are telling us when it happened so isn't that the source of truth?

julienledem:
Can we add a unit test that reproduces the problem? Thanks!

api/src/main/java/marquez/db/RunStateDao.java
wslulciuc: The call `createRunDao().updateRunState()` updates the current run state for the `runs` table. We are still using transitioned time provided by the caller. I'll update the method name to `createRunDao().updateCurrentRunState()` to better reflect what is happening.

henneberger: Yeah, the most recent updated timestamp in the runs table would be the most recent state transition.

wslulciuc: This issue is that the call can provide a transition time in the past

henneberger: then the run state would be out of sync. You could employ the same strategy we used for the start state to keep them in sync without forcing a new timestamp. https://github.com/MarquezProject/marquez/blob/77d14045dacd8a5dcf16085c2b7e4f4c35062c0b/api/src/main/java/marquez/db/RunDao.java#L121

wslulciuc: Do we care that the `updatedAt` and the `transitionAt` time can possibly be in the past, but the `createdAt` can be in the "future"? I've always viewed `updatedAt` as meaning "when the row was updated".

henneberger: If it is just for auditing then we could just drop it and use logs for the audit trail.

wslulciuc: We return `updatedAt` back to the caller to convey that the state was updated for the run and usually only in the case of a state transition, so we still want to keep the timestamp around. I think it make sense to add your condiiton suggestion above:

```
updated_at < :updatedAt
```

henneberger: my PR #1007 adds a transition_at column for the current run state. If updatedAt is for audit logging then we shouldn't be using it in business logic

wslulciuc: Cool. We can close this PR then 👍

wslulciuc:
Will be address in #1007

PR #1036, opened 2021-03-05
Add data lineage query to graphql endpoint (3 of 3)

henneberger:
Resolves a data lineage by Job name and namespace.


api/src/main/java/marquez/db/mappers/DatasetVersionMapper.java
```diff
@@ -0,0 +1,61 @@
...
+import marquez.service.models.DbTableVersion;
+import marquez.service.models.Run;
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.statement.StatementContext;
+
+public final class DatasetVersionMapper implements RowMapper<DatasetVersion> {
+  @Override
+  public DatasetVersion map(@NonNull ResultSet results, @NonNull StatementContext context)
+      throws SQLException {
+    return new DbTableVersion(
```
henneberger, line 43: adding stream stuff here...

api/src/main/java/marquez/db/mappers/JobMapper.java
```diff
@@ -0,0 +1,90 @@
...
+        JobType.valueOf(stringOrThrow(results, Columns.TYPE)),
+        JobName.of(stringOrThrow(results, Columns.NAME)),
+        timestampOrThrow(results, Columns.CREATED_AT),
+        timestampOrThrow(results, Columns.UPDATED_AT),
+        getDatasetFromJsonOrNull(results, "current_inputs"),
+        getDatasetFromJsonOrNull(results, "current_outputs"),
+        urlOrNull(results, "current_location"),
+        toContext(results, Columns.CONTEXT),
+        stringOrNull(results, Columns.DESCRIPTION),
+        null // toRun(results, "run")
```
wslulciuc, line 64: We'll want to return the latest run for a job if it exists.

henneberger: This is done late in two parts in the job dao. I tried writing the more complex query to resolve the run in one go but ran into a sql.timestamp->instant serialization issue. It'd be solveable but i wanted to keep it simple here.

wslulciuc: Sounds good. Maybe add a note on fixing the `sql.timestamp->instant serialization` issue at a later time and/or the current logic?

wslulciuc (approved):
LGTM 👍 💯 🥇

PR #1042, opened 2021-03-08
Order runs by `started_at` date

phixMe:
This changes the API to return runs ordered by the `STARTED_AT` (`transitioned_at` in `run_states` table) time instead of the `created_at` time.


api/src/main/java/marquez/db/RunDao.java
```diff
@@ -151,7 +151,7 @@ default void updateJobVersionUuid(UUID rowUuid, Instant updatedAt, UUID jobVersi
       SELECT_RUN
           + "INNER JOIN job_versions AS jv ON r.job_version_uuid = jv.uuid "
           + "WHERE jv.namespace_name = :namespace and jv.job_name = :jobName "
-          + "ORDER BY r.created_at DESC "
+          + "ORDER BY STARTED_AT DESC NULLS LAST "
```
wslulciuc, line 154: I think this should be:

```
ORDER BY r.started_at DESC NULLS LAST
```

wslulciuc:
@phixMe: Mind also adding / updating an integration test for listing runs to [`MarquezAppIntegrationTest`](https://github.com/MarquezProject/marquez/blob/main/api/src/test/java/marquez/MarquezAppIntegrationTest.java)?

api/src/main/java/marquez/db/RunDao.java
phixMe: The runs table doesn't have a `started_at` time. This is using the alias from the `SELECT_RUN` constant which is the joined in `transitioned_at` column from `run_states`.

wslulciuc: Ahh, I see 👍

api/src/main/resources/marquez/db/migration/V23__index_transitioned_at_on_run_states.sql
```diff
@@ -0,0 +1,3 @@
+create index run_states_transitioned_at_index
+    on run_states (transitioned_at desc);
```
wslulciuc, line 2: Adding this index seems fine with me, @henneberger thoughts?

wslulciuc (approved):
LGTM 🥇

api/src/main/resources/marquez/db/migration/V23__index_transitioned_at_on_run_states.sql
henneberger: Should this on be `runs`, not `run_states`?

henneberger: Oh, that is in the PR that I've been trying to get landed.

wslulciuc: Yeah, currently, the `transitioned_at` column is only on the `run_states` table.

henneberger: This index isn't good since we aren't trying to order the entire run state table by transition time.  Can we wait until that gets landed and have an index that does:
```
create index run_transitioned_at_index
    on run_states (namespace_name, job_name, transitioned_at desc);
    ```

wslulciuc:
FYI, we can still merge the PR though the integration test for Airflow is failing (it's unrelated and currently fails only for forks)

api/src/main/resources/marquez/db/migration/V23__index_transitioned_at_on_run_states.sql
wslulciuc: Sure, so remove the `transitioned_at` index for now? /cc @phixMe

phixMe: @wslulciuc @henneberger Sounds good, removing it.

Issue #906, opened 2021-01-22
Add pre-commit git hook to check code style


wslulciuc:
Failing pipelines due to formatting errors are annoying! Let's add a `pre-commit` git hook to check / apply code formatting using spotless.

wslulciuc:
Fixed: https://github.com/MarquezProject/marquez/pull/1057

PR #1045, opened 2021-03-08
Updated build to automatically switch to JDK 8 if env is available

collado-mike:
This enables building the entire project with a single gradle command, making it easier to run all tests before committing


integrations/spark/src/test/java/marquez/spark/agent/lifecycle/LibraryTest.java
```diff
@@ -1,11 +1,7 @@
 package marquez.spark.agent.lifecycle;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.fail;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.reset;
-import static org.mockito.Mockito.when;
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
```
wslulciuc, line 4: _Minor:_ As part of our style guide, we avoid wildcards `*`. See, https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

wslulciuc (approved):
Nice work, @collado-mike! Left a minor comment, otherwise 👍

wslulciuc:
@collado-mike: Looks like you'll also need to sign your commits :)

integrations/spark/src/test/java/marquez/spark/agent/lifecycle/LibraryTest.java
collado-mike: Hmm, thanks. Intellij must have done that automatically. 

If you don't mind, can we just codify these kinds of style preferences into the style checker? I don't like wasting PR comments on things that can/should be caught by a linter. It's also a waste of build cycles, as I'll now end up modifying the commit, pushing to origin, waiting for circle,... I think spotless probably has a rule to disallow * imports, so when I run `gradle build` (for the whole project 😁), it'll catch these issues before I push. Sound ok? I can just update the style checker as part of this PR (unless you prefer a separate one).

wslulciuc: Make sense. It's cool to merge without the style change. I've been meaning to add a pre-commit hook that checks the code for styling (see https://github.com/MarquezProject/marquez/issues/906) and will get to it this some time this week.

PR #1055, opened 2021-03-09
Removed checkstyle and spotbugs configuration from build

collado-mike:
Checkstyle and spotbugs are adding a ton of noise to the build. Since neither plugin is set to fail the build and numerous violations of each is found in the codebase, every build logs a bunch of exceptions and style warnings that are ignored. This PR removes both plugins from the build so only errors that fail the build are in the build log


wslulciuc (approved):
LGTM 👍

PR #1056, opened 2021-03-10
Spotless fail wildcard imports

collado-mike:
This configures `gradle build` to fail if wildcard imports are found. Even though the spotless plugin applies the `googleJavaFormat()` rules, and the google formatter disallows wildcard imports, the build doesn't fail if wildcard imports are found.

It would be nice to automatically fix wildcard imports, but that's a _lot_ more work, so...


wslulciuc (approved):
Sweet! 🥇

PR #1057, opened 2021-03-10
Add .pre-commit-config.yaml

wslulciuc:
This PR adds support for managing git hooks using [`pre-commit`](https://pre-commit.com/index.html):

Screen Shot 2021-03-09 at 4 45 22 PM

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+repos:
+  -   repo: https://github.com/pre-commit/pre-commit-hooks
+      rev: v3.2.0
+      hooks:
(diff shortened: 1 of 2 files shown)


collado-mike (approved):
LGTM.

Why is it that `spotlessCheck` will fail on the CI server, but doesn't fail when running `gradle build` locally?

wslulciuc:
> Why is it that spotlessCheck will fail on the CI server, but doesn't fail when running gradle build locally?

@collado-mike It's possible you have binaries cached? Locally, if I use `clean`, the build will fail:

```
$ ./gradlew clean :api:build
```

PR #1058, opened 2021-03-10
Address sporadic spark integration test failures

collado-mike:
Upgraded to JUnit 5 in order to use the `@RepeatedTest` annotation to run the `LibraryTest` many times. This got the build to reliably fail (though I had to use a big number to get it fail at least once every build). I also had to configure the build to print out the test result text files (and configured circle to upload the test results). I found the test was due to a concurrency problem where the queryExecution was sometimes removed from a map before we had a chance to access it. Changing the code to keep a reference to the queryExecution means we're not subject to concurrency problems anymore. Build now reliably passes.


henneberger (approved):
Nice!

PR #1065, opened 2021-03-12
Refactor spark plan traversal to find input/output datasets from Datasources

collado-mike:
Refactor the logical plan traversal to use a composition rather than inheritance approach. This aligns dataset/datasource naming and namespaces with previous discussions re: URI format. Also, I included classpath checking for BigQuery relation matching to avoid runtime errors for consumers that aren't running on Dataproc and don't have the BQ library on the classpath.


integrations/spark/integrations/sparksql/1.json
```diff
@@ -1,65 +1,176 @@
...
   },
-  "inputs" : [ {
-    "namespace" : "gs.bucket",
-    "name" : "gs://bucket/data.txt"
-  } ],
-  "outputs" : [ ],
-  "producer" : "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client"
+  "inputs": [
+    {
+      "namespace": "file://",
```
henneberger, line 153: Can we drop the trailing `://` part? It seems implied.

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/HadoopFsRelationVisitor.java
```diff
@@ -0,0 +1,88 @@
...
+                  .namespace(namespace)
+                  .name(uri.getPath())
+                  .facets(
+                      DatasetFacet.builder()
+                          .schema(PlanUtils.schemaFacet(relation.schema()))
+                          .dataSource(
+                              DatasourceDatasetFacet.builder()
+                                  ._producer(
+                                      URI.create(
+                                          "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client"))
```
julienledem, line 76: this should identify this code. I would use: "https://github.com/MarquezProject/blob/1.13.0/marquez/integrations/spark"
(if you can look up the current version)

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/HadoopFsRelationVisitor.java
```diff
@@ -0,0 +1,88 @@
...
+                                  ._schemaURL(
+                                      URI.create(
+                                          "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#DatasourceDatasetFacet"))
```
julienledem, line 79: v1-0-0 doesn't exist yet. Maybe use main in the meantime?

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/InputDatasetVisitors.java
```diff
@@ -0,0 +1,34 @@
...
+  public InputDatasetVisitors(SQLContext sqlContext) {
+    this.sqlContext = sqlContext;
+  }
+
+  @Override
+  public List<PartialFunction<LogicalPlan, List<Dataset>>> get() {
+    List<PartialFunction<LogicalPlan, List<Dataset>>> list = new ArrayList<>();
+    list.add(new HadoopFsRelationVisitor(sqlContext.sparkContext()));
+    list.add(new JDBCRelationVisitor(sqlContext));
+    if (BigQueryNodeVisitor.hasBigQueryClasses()) {
```
julienledem, line 29: Can we call the static method if the class itself depends on classes missing in the classpath? Or would an exception occured only when calling the method actually depending on it?

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/JDBCRelationVisitor.java
```diff
@@ -0,0 +1,108 @@
...
+                          DatasetFacet.builder()
+                              .schema(PlanUtils.schemaFacet(relation.schema()))
+                              .dataSource(
+                                  DatasourceDatasetFacet.builder()
+                                      ._producer(
+                                          URI.create(
+                                              "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client"))
+                                      ._schemaURL(
+                                          URI.create(
+                                              "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#DatasourceDatasetFacet"))
```
julienledem, line 100: same comment and let's factor this logic out

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/InputDatasetVisitors.java
julienledem: is there a way to test this in the build?

integrations/spark/src/main/scala/marquez/spark/agent/plan/DatasetCollector.scala
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.spark.sql.execution.QueryExecution
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, InsertIntoDataSourceCommand, InsertIntoHadoopFsRelationCommand, LogicalRelation}
+import org.apache.spark.sql.execution.metric.SQLMetrics
+import org.apache.spark.sql.marquez.JDBCOperation
+import org.apache.spark.sql.types.StructType
+
+import java.net.URI
+import scala.collection.JavaConversions._
+
+class DatasetCollector {
```
julienledem, line 18: Is the Scala code used?

julienledem (approved):
this looks great.
I made a few comments but otherwise this looks good to go.

wslulciuc (approved):
Great work, @collado-mike! I have a similar question on the scala code usage as Julien, otherwise 💯 🥇

integrations/spark/integrations/sparksql/1.json
collado-mike: This is derived from the dataset URI. Running in a spark cluster in the cloud, it would be read from a URI like `gs://bucket/path/to/file`. Running locally, it's `file:///path/to/file`. In those other cases, the namespace really is `gs://bucket`- it's that in the `file://` URI, the authority is blank.

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/InputDatasetVisitors.java
collado-mike: The class itself doesn't have a direct dependency on the BigQueryRelation class- only if the class had a static initializer (static property or static initialization block) that referenced the BigQueryRelation class would it cause a problem.

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/JDBCRelationVisitor.java
collado-mike: Sure, I copied these URLs from elsewhere in the code. I'll refactor all the references to a utility class.

integrations/spark/src/main/scala/marquez/spark/agent/plan/DatasetCollector.scala
collado-mike: Oops! That's what I get for not checking my own PRs 🤦🏽‍♂️

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/InputDatasetVisitors.java
julienledem: 👍

julienledem (approved):

PR #1092, opened 2021-03-16
Bump airflow integration's required python version to 3.6

mobuchowski:
This change is more about acknowledging fact than dropping support, as python >3.5 features are used in codebase, like f-strings.


wslulciuc (approved):
LGTM 👍

PR #1098, opened 2021-03-17
Add job context to run, fix lineage listener dataset version id and create job version when done

henneberger:

### DIFF
--- a/api/src/main/java/marquez/db/RunDao.java
+++ b/api/src/main/java/marquez/db/RunDao.java
@@ -75,9 +75,10 @@ public interface RunDao extends BaseDao {
   void updateEndState(UUID rowUuid, Instant transitionedAt, UUID endRunStateUuid);
 
   String SELECT_RUN =
-      "SELECT r.*, ra.args, r.started_at, r.ended_at, ra.args "
+      "SELECT r.*, ra.args, ra.args, ctx.context "
           + "FROM runs AS r "
-          + "LEFT OUTER JOIN run_args AS ra ON (ra.uuid = r.run_args_uuid) ";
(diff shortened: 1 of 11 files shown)


api/src/main/java/marquez/db/RunDao.java
```diff
@@ -274,6 +280,6 @@ default RunRow upsertFromRun(
   @SqlQuery(
       SELECT_RUN
           + "where r.job_name = :jobName and r.namespace_name = :namespaceName "
-          + "order by updated_at desc limit 1")
```
phixMe, line 277: Thanks @henneberger for the update

api/src/main/java/marquez/service/RunService.java
```diff
@@ -69,37 +69,36 @@ public void markRunAs(
     ExtendedRunRow runRow = findByRow(runId.getValue()).get();
     runStateDao.updateRunState(runId.getValue(), runState, transitionedAt);
 
-    if (runState == COMPLETED) {
+    if (runState.isDone()) {
```
wslulciuc, line 72: Do we want to create job versions even though the job run isn't successful? Only successful runs have output datasets.

wslulciuc: Hmm, this logic might make sense. A failed or aborted job will simply produce a job version without any outputs.

julienledem: Yes, there should be a version for failed job. That's how you'll be able to figure out the job has failed because a bug was introduced in the latest version for example.

wslulciuc: Yeah, that make sense and looks like we still maintain the logic to only notify the `RunListener` of input / output changes on job completion.

wslulciuc (approved):
LGTM 👍

PR #1102, opened 2021-03-18
draft: Airflow openlineage integration

mobuchowski:


integrations/airflow/marquez_airflow/dag.py
```diff
@@ -55,19 +53,19 @@
...
         # run Airflow's create_dagrun() first
         dagrun = super(DAG, self).create_dagrun(*args, **kwargs)
 
         create_dag_start_ms = self._now_ms()
         try:
-            _MARQUEZ.create_namespace()
             self._register_dagrun(
                 dagrun,
-                DagUtils.get_execution_date(**kwargs),
-                DagUtils.get_run_args(**kwargs)
```
julienledem, line 70: is this something that needs to be added as a facet?

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -77,44 +75,47 @@ def create_dagrun(self, *args, **kwargs):
...
-                    self.new_run_id(),
-                    step,
-                    run_args,
+                step = self._extract_metadata(dagrun, task)
+
+                job_name = self._marquez_job_name(self.dag_id, task.task_id)
+                task_run_id = _MARQUEZ.start_task(
+                    job_name,
+                    self.description,
+                    DagUtils.to_iso_8601(self._now_ms()),
```
julienledem, line 93: ideally this is the official start time per airflow.

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -125,60 +126,60 @@ def handle_callback(self, *args, **kwargs):
...
+                self._get_location(task),
+                DagUtils.to_iso_8601(task_instance.start_date),
+                DagUtils.to_iso_8601(task_instance.end_date),
                 step,
-                run_args,
-                DagUtils.to_iso_8601(ti.start_date),
-                DagUtils.to_iso_8601(ti.end_date)
-            ) for step in steps]
-            if not run_ids:
+            )
```
julienledem, line 160: this looks duplicated (and slightly different) from line 90

integrations/airflow/marquez_airflow/extractors/__init__.py
```diff
@@ -124,24 +129,29 @@ def __repr__(self):
...
-        if context:
-            self.context = context
+        self.inputs = inputs
+        self.outputs = outputs
+        self.context = context
+
+        if not inputs:
+            self.inputs = []
+        if not outputs:
+            self.outputs = []
```
julienledem, line 152: I think we should be able to tell the difference between "we have no inputs" ( [] ) and "I don't know if we have inputs" ( null )

integrations/airflow/marquez_airflow/extractors/__init__.py
```diff
@@ -166,12 +176,11 @@ def validate(self):
         assert (self.operator_class is not None and
                 self.operator.__class__ == self.operator_class)
 
-    # TODO: Only return a single StepMetadata object.
     @abstractmethod
-    def extract(self) -> [StepMetadata]:
+    def extract(self) -> StepMetadata:
```
julienledem, line 180: the only thing is that we might break compatibility for existing custom extractors.

julienledem: (but otherwise this is a good change)

integrations/airflow/openlineage/facet.py
```diff
@@ -0,0 +1,119 @@
...
+
+from openlineage.constants import DEFAULT_PRODUCER
+
+
+PRODUCER = os.getenv("MARQUEZ_PRODUCER", DEFAULT_PRODUCER)
+SCHEMA_URI = "https://raw.githubusercontent.com/OpenLineage/OpenLineage/main/spec/OpenLineage.json"
+
+
+@attr.s
+class BaseFacet:
```
julienledem, line 14: nice!

julienledem: eventually this should go in OpenLineage itself

integrations/airflow/tests/integration/integration.py
```diff
@@ -157,7 +159,8 @@ def check_datasets_meta():
     assert out_table['sourceName'] == SOURCE_NAME
     assert out_table['fields'] == OUT_TABLE_FIELDS
     assert len(out_table['tags']) == 0
-    assert out_table['lastModifiedAt'] is not None
+    # TODO: marquez does not update this on openlineage event
+    # assert out_table['lastModifiedAt'] is not None
```
julienledem, line 163: @henneberger should it?

integrations/airflow/tests/integration/integration.py
```diff
@@ -175,19 +179,23 @@ def check_jobs_meta():
...
-           'airflow.operators.postgres_operator.PostgresOperator'
-    assert if_not_exists_job['context']['airflow.task_info'] is not None
-    assert if_not_exists_job['context']['airflow.version'] == AIRFLOW_VERSION
-    assert if_not_exists_job['context']['marquez_airflow.version'] == MARQUEZ_AIRFLOW_VERSION
+    # TODO: no airflow context data yet
+    # assert if_not_exists_job['context']['airflow.operator'] == \
+    #        'airflow.operators.postgres_operator.PostgresOperator'
+    # assert if_not_exists_job['context']['airflow.task_info'] is not None
+    # assert if_not_exists_job['context']['airflow.version'] == AIRFLOW_VERSION
+    # assert if_not_exists_job['context']['marquez_airflow.version'] == MARQUEZ_AIRFLOW_VERSION
```
julienledem, line 190: let's add those to a custom facet.

julienledem (commented):
This is great progress.
I left some comments inline.
I'm looking forward to seeing this completed.

integrations/airflow/tests/integration/integration.py
mobuchowski: Now probably does, will check with this fix:
https://github.com/MarquezProject/marquez/pull/1107

integrations/airflow/marquez_airflow/dag.py
mobuchowski: Good catch, we can use task instance's start time here.

integrations/airflow/marquez_airflow/extractors/__init__.py
mobuchowski: Makes sense, but is this distinction meaningful in OpenLineage events?

integrations/airflow/marquez_airflow/dag.py
mobuchowski: I think it can be - if we think that whether this dag run is externally triggered is a meaningful information. 
As it already is in the repo, I'd add it.

integrations/airflow/marquez_airflow/dag.py
mobuchowski: The original assumption that I kept is that 

    # Note: task_run_id could be missing if it was removed from airflow
    # or the job could not be registered.

Then we emit START event again, as we OpenLineage contract specifies that we need to have 1 start event.
If the assumption is not true, then it could be removed I think.

integrations/airflow/openlineage/facet.py
mobuchowski: Agreed

integrations/airflow/tests/integration/integration.py
mobuchowski: Will add those custom facets in separate PR

integrations/airflow/marquez_airflow/extractors/__init__.py
mobuchowski: That makes sense, but the question is if the distinction is meaningful for OpenLineage events. If not, then non-nulleable data would be prefered IMO.

integrations/airflow/marquez_airflow/extractors/__init__.py
mobuchowski: Added few compatibility changes for this situation

PR #1135, opened 2021-03-22
[blocked] Marquez-airflow openlineage integration

mobuchowski:

PR #1138, opened 2021-03-23
Add logs to setup script for airflow integration test

wslulciuc:


wslulciuc:
Will be fixed in #1135

PR #1140, opened 2021-03-23
Make transitioned_at column a date (2 of 2)

henneberger:
- Denormalize dataset fields
- Add tag to dataset field
- Write new tag with denormalizations
- Remove tag uuids from dataset row
- Fix string concat
- Update db migrate, add run index
- Add indexes
- Drop old index
- Fix old index
- Make transitioned_at column a date


api/src/main/java/marquez/db/RunDao.java
```diff
@@ -237,7 +237,14 @@ default void registerRunOutputDataset(UUID runUuid, JobMeta jobMeta) {
                       toSchemaFields(d.getFields()),
                       runUuid);
               datasetVersionDao.upsert(
-                  UUID.randomUUID(), Instant.now(), dsRow.get().getUuid(), version, runUuid);
+                  UUID.randomUUID(),
+                  Instant.now(),
+                  dsRow.get().getUuid(),
+                  version,
+                  runUuid,
+                  datasetVersionDao.toPgObjectFields(d.getFields()),
```
henneberger, line 245: This change won't propagate tags on fields. A change would be required to keep the current propagation behavior.

wslulciuc: The tags for the dataset fields wouldn't be propagated? Mind opening an issue outlining the change? Thanks!

wslulciuc (approved):
Nice catch! 🚢 it

PR #1143, opened 2021-03-24
Add seed cmd

wslulciuc:
This PR replaces [seed-db.sh](https://github.com/MarquezProject/marquez/blob/main/web/docker/db/seed-db.sh) with `marquez-api.jar seed` that will seed the HTTP API with metadata:

```
$ java -jar marquez-api.jar seed --help
       seed [--host HOST] [--port PORT] [-h] [file]

seeds the HTTP API with metadata

positional arguments:
  file                   application configuration file

named arguments:
  --host HOST            the HTTP API server host (default: localhost)
  --port PORT            the HTTP API server port (default: 8080)
```


api/src/main/java/marquez/cli/SeedCommand.java
```diff
@@ -0,0 +1,1419 @@
...
+import marquez.client.models.NamespaceMeta;
+import marquez.client.models.Run;
+import marquez.client.models.RunMeta;
+import marquez.client.models.Source;
+import marquez.client.models.SourceMeta;
+
+@Slf4j
+public final class SeedCommand extends ConfiguredCommand<MarquezConfig> {
+  static final String DEFAULT_DATAKIN_HOST = "localhost";
+  static final int DEFAULT_DATAKIN_PORT = 8080;
```
collado-mike, line 50: `MARQUEZ_HOST` and `MARQUEZ_PORT`

api/src/main/java/marquez/cli/SeedCommand.java
```diff
@@ -0,0 +1,1419 @@
...
+            .description("Food delivery example!")
+            .build();
+    final Namespace newNamespace = client.createNamespace(NAMESPACE_NAME, namespaceMeta);
+    log.info("Created namespace: {}", newNamespace);
+
+    // (2) Create source
+    final SourceMeta sourceMeta =
+        SourceMeta.builder()
+            .type("MYSQL")
+            .connectionUrl("jdbc:mysql://localhost:3306/deliveries")
```
collado-mike, line 109: postgres?

api/src/main/java/marquez/cli/SeedCommand.java
```diff
@@ -0,0 +1,1419 @@
...
+import marquez.client.models.JobMeta;
+import marquez.client.models.Namespace;
+import marquez.client.models.NamespaceMeta;
+import marquez.client.models.Run;
+import marquez.client.models.RunMeta;
+import marquez.client.models.Source;
+import marquez.client.models.SourceMeta;
+
+@Slf4j
+public final class SeedCommand extends ConfiguredCommand<MarquezConfig> {
```
collado-mike, line 48: Doc comments 😁

api/src/main/java/marquez/cli/SeedCommand.java
```diff
@@ -0,0 +1,1419 @@
...
+                  .type(change.getToType())
+                  .tags(field.getTags())
+                  .description(field.getDescription().orElse(null))
+                  .build()
+              : field);
+    }
+    return fieldsWithChange.build();
+  }
+
+  static final ImmutableMap<String, DbTableMeta> DATASET_META =
```
collado-mike, line 255: Seems like something that would ultimately be better as a yaml or json file. Not a blocking comment, but just a thought.

collado-mike (commented):
Awesome. Just a quick typo and then please add a short doc comment describing what the process does

api/src/main/java/marquez/cli/SeedCommand.java
wslulciuc: We originally had the metadata in a `json` file, but became increasingly hard to manage and switch to using the java client instead vs `json` + curl cmds.

api/src/main/java/marquez/cli/SeedCommand.java
wslulciuc: Good point!

api/src/main/java/marquez/cli/SeedCommand.java
wslulciuc: For the example, the database type isn't as important. But, I'll update because postgres is awesome!

api/src/main/java/marquez/cli/SeedCommand.java
collado-mike: Yeah, I can see why the logic makes sense to have in the Java command- it's mostly the data I was thinking to pull into a config file.

wslulciuc: The data can get a bit complex. For example, defining how many runs a job should have, for what run a code change was introduced, associating a run with a transitions etc. I think a yaml file would make a lot of sense when we switch to using OpenLineage events, it was just easier to define the logic in code.

PR #1146, opened 2021-03-26
Spark integration fixes

collado-mike:
These fixes address some failures happening when posting lineage events from Spark to Marquez. Notably, datasets with `long` data types (there are many :) ) fail to get written. This was irritating to track down because the code swallows the error if no type mapping is found. Additionally, the JDBC relation visitor didn't include the table name in the URI, which caused the dataset to have a namespace, but no name. 

Finally, I've seen the `SparkSQLExecutionContext` in the context map return null, causing some NPEs. Added a null check for that.


henneberger (approved):

julienledem (approved):
LGTM!

PR #1148, opened 2021-03-26
[blocked] Airflow bigquery stats

mobuchowski:
Depends on https://github.com/MarquezProject/marquez/pull/1147


integrations/airflow/marquez_airflow/dag.py
```diff
@@ -77,44 +77,49 @@ def create_dagrun(self, *args, **kwargs):
 
         return dagrun
 
-    def _register_dagrun(self, dagrun, execution_date, run_args):
+    # We make the assumption that when a DAG run is created, it's
```
collado-mike, line 80: nit- `its`

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -77,44 +77,49 @@ def create_dagrun(self, *args, **kwargs):
...
+                )
             except Exception as e:
                 self.log.error(
                     f'Failed to record task {task_id}: {e} '
                     f'{self._timed_log_message(t)}',
                     exc_info=True)
 
     def handle_callback(self, *args, **kwargs):
-        self.log.debug(f"handle_callback({args}, {kwargs})")
+        self.log.info(f"handle_callback({args}, {kwargs})")
```
collado-mike, line 117: This seems very much like a debug log line

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -125,60 +130,61 @@ def handle_callback(self, *args, **kwargs):
...
-        task = self.get_task(ti.task_id)
-        run_ids = JobIdMapping.pop(
-            self._marquez_job_name_from_ti(ti), dagrun.run_id, session)
-        steps = self._extract_metadata(dagrun, task, ti)
+    def _report_task_instance(self, task_instance, dagrun, session):
+        task = self.get_task(task_instance.task_id)
 
-        # Note: run_ids could be missing if it was removed from airflow
+        # Note: task_run_id could be missing if it was removed from airflow
         # or the job could not be registered.
```
collado-mike, line 148: The danger here is in some sub-workflow that registered itself as a child of this task losing its connection if the original DAG task never completes, right? Can we consider deriving deterministic task run ids so that if we somehow lose the job id mapping, we can still generate the same id? E.g., I think the DAG run id is already a UUID, right? Is there harm in appending the task name to the DAG run id and generating a unique run id that way?

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -188,12 +194,27 @@ def _extract_metadata(self, dagrun, task, ti=None):
...
+                        )
+                    elif len(step) == 1:
+                        return step[0]
+                    else:
+                        self.log.warning(
+                            f'Extractor {extractor.__name__} {task_info} '
+                            f'returned more then one StepMetadata instance: {step} '
+                            f'will drop steps except for first!'
+                        )
+                        return step[0]
```
collado-mike, line 216: Minor point, but I would structure this as 
```python
if len(step) == 0:
  # return step metadata
elif len(step) > 1: 
  self.log.warning(...)
# return step[0] as long as there is at least one element
return step[0]
```

integrations/airflow/marquez_airflow/extractors/bigquery_extractor.py
```diff
@@ -29,22 +31,66 @@
...
 from marquez_airflow.utils import (
     get_job_name
 )
 
 # BIGQUERY DAGs doesn't use this.
 # Required to pass marquez server validation.
+from openlineage.facet import BaseFacet
+
 _BIGQUERY_CONN_URL = \
     'jdbc:bigquery://https://www.googleapis.com/bigquery/v2:443'
```
collado-mike, line 45: Based on the naming spec at https://github.com/OpenLineage/OpenLineage/pull/31/files , I think the namespace for BigQuery tables is just `bigquery`

integrations/airflow/marquez_airflow/facets.py
```diff
@@ -0,0 +1,28 @@
+import attr
+
+from airflow.version import version as AIRFLOW_VERSION
+from marquez_airflow.version import VERSION as MARQUEZ_AIRFLOW_VERSION
+
+from openlineage.facet import BaseFacet
+
+
+@attr.s
+class AirflowVersionRunFacet(BaseFacet):
```
collado-mike, line 10: Very nice!

integrations/airflow/marquez_airflow/marquez.py
```diff
@@ -15,108 +22,187 @@
 log = logging.getLogger(__name__)
 
 
-class Marquez:
+class MarquezAdapter:
```
collado-mike, line 25: Some class and method level doc comments would be extremely useful here, as I can imagine others wanting to tap in here for their own purposes.

integrations/airflow/openlineage/client.py
```diff
@@ -0,0 +1,61 @@
...
+
+@attr.s
+class OpenLineageClientOptions:
+    timeout: float = attr.ib(default=5.0)
+    verify: bool = attr.ib(default=True)
+    api_key: str = attr.ib(default=None)
+    adapter: HTTPAdapter = attr.ib(default=None)
+
+
+class OpenLineageClient:
```
collado-mike, line 20: Worth moving into a separate module from the airflow integration?

integrations/airflow/openlineage/facet.py
```diff
@@ -0,0 +1,119 @@
...
+
+from openlineage.constants import DEFAULT_PRODUCER
+
+
+PRODUCER = os.getenv("MARQUEZ_PRODUCER", DEFAULT_PRODUCER)
+SCHEMA_URI = "https://raw.githubusercontent.com/OpenLineage/OpenLineage/main/spec/OpenLineage.json"
+
+
+@attr.s
+class BaseFacet:
```
collado-mike, line 14: Same here- all these facets are general and I think they're better off in a separate OpenLineage client module

PR #1151, opened 2021-03-26
Add lineage service

henneberger:


api/src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -136,16 +163,22 @@ default JobVersionBag createJobVersionOnComplete(
 
     createRunDao().updateJobVersion(runUuid, jobVersion.getUuid());
 
-    jobDao.updateVersion(jobRow.getUuid(), transitionedAt, jobVersion.getUuid());
+    // Only set current version if completed
```
wslulciuc, line 166: Do we only want to set the latest version on complete? We also create new versions when a job run is [marked as _done_](bourdains).

api/src/main/java/marquez/api/LineageResource.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import marquez.service.ServiceFactory;
+import marquez.service.models.NodeId;
+
+@Path("/api/v1-beta")
```
wslulciuc, line 22: Good idea to make the distinction that the lineageAPI is in `beta` and something we are still working on 👍

api/src/main/java/marquez/db/JobVersionDao.java
henneberger: Yeah. The latest dataset version is the latest version that was successful. This makes sense in the context of a lineage graph as the graph can still be constructed w/ the latest run returned is the actual latest run which could be in a failed state.

wslulciuc: Sounds good. When we introduce the JobVersion API, we'll want to document the logic for how the current version is set.

wslulciuc (approved):
LGTM 💯 🥇

PR #1162, opened 2021-03-29
Assure latest tags are returned from dataset

henneberger:


wslulciuc (approved):
LGTM 👍

api/src/main/java/marquez/service/models/Dataset.java
```diff
@@ -54,7 +56,7 @@
   @Getter private final Instant updatedAt;
   @Getter private final NamespaceName namespace;
   @Getter private final SourceName sourceName;
-  @Getter private final ImmutableList<Field> fields;
+  @Getter @Setter private List<Field> fields;
```
collado-mike, line 59: We should avoid backward incompatible changes like this. Dataset is a pretty core model, so I can imagine a lot of other code accesses this field. Given that we use Lombok, I think we could change the field to be an ImmutableList and just implement the getter to return a plain List.

PR #1164, opened 2021-03-29
Update OpenLinageDao to stop converting URI structures to contain underscores instead of colons and slashes

collado-mike:


wslulciuc (approved):
LGTM 👍

henneberger (approved):

PR #1180, opened 2021-04-01
Added BigQueryRelationTransformer and deleted BigQueryNodeVisitor

collado-mike:
Since the big query jar isn't loaded by the main classloader in a Spark application, the existing BigQueryNodeVisitor never registered the BigQueryRelation on the classpath. 

This solution uses the existing instrumentation setup to modify the BigQueryRelation class and other associated BQ logical plan nodes to implement a new `DatasetSource` interface. This allows the rewritten classes to be discovered by the node visitor regardless of the classloader that loads the BigQueryRelation class. 

I think this is a good _temporary_ solution, though I think there's a better long-term approach I'm happy to write up in a new issue. 

I also included some test infrastructure, specifically for mocking BigQuery, so we can validate the node visiting, including input/output statistics, dataset/datasource naming, and schema detection. The test infrastructure sets up a mock BQ datasource that returns a supplied DataFrame to mock BQ table reads. There is still more test cases to be covered (e.g., JDBC relations, reads from HDFS, writes to BQ, etc.), but this is a good start and I didn't want to make the PR to big to be reviewed.

Incidentally, the test caught a bug where HDFS writes weren't reporting the correct schema.


wslulciuc:
> I think this is a good temporary solution, though I think there's a better long-term approach I'm happy to write up in a new issue.

@collado-mike Sounds great! Mind writing a proposal on your thoughts for a long term solution? You can add your proposal as an issue and use the label [`proposal`](https://github.com/MarquezProject/marquez/labels/proposal)

integrations/spark/src/test/java/com/google/cloud/spark/bigquery/repackaged/com/google/cloud/bigquery/MockBigQueryRelationProvider.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.spark.sql.sources.TableScan;
+import org.apache.spark.sql.types.StructType;
+import org.mockito.Mockito;
+import scala.Option;
+import scala.collection.JavaConversions;
+import scala.collection.immutable.Map;
+import scala.runtime.AbstractFunction0;
+
+/**
+ * Mock relation provider that uses a {@link Mockito} mock instance of {@link BigQuery} for testing.
```
wslulciuc, line 32: 💯

integrations/spark/src/test/java/marquez/spark/agent/lifecycle/LibraryTest.java
```diff
@@ -34,25 +31,15 @@
...
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 import scala.Tuple2;
 
+@ExtendWith(SparkAgentTestExtension.class)
```
wslulciuc, line 41: 🥇

wslulciuc (commented):
Amazing work, @collado-mike! 🥇

wslulciuc (approved):

PR #1190, opened 2021-04-06
Basic CTE support in marquez-airflow

mobuchowski:
This adds basic CTE support in marquez-airflow. Hovewer, I'm not happy with the solution. Now, it just extends existing parser, which works directly on tokenized statements provided by sqlparse. If we want to add more functionality to existing parser, it would be good to move to working on more logical structure - translate token list to abstract syntax tree, and do the dataset extraction work on the AST itself. It won't be easy task tho, there's lot of complexity there, not to mention different SQL variants.


integrations/airflow/marquez_airflow/extractors/sql/parser.py
```diff
@@ -45,20 +44,35 @@ def _match_on(token, keywords):
...
+        try:
+            # Determine if the table contains the schema
+            # separated by a dot (format: 'schema.table')
+            dot = next(token_list)
+            if dot.match(Punctuation, '.'):
+                table_name += dot.value
+                table_name += next(token_list).value
+        except StopIteration:
+            pass
+        return table_name
```
julienledem, line 60: We should normalize the table name to be `{schema}.{table}`

integrations/airflow/marquez_airflow/extractors/sql/parser.py
```diff
@@ -45,20 +44,35 @@ def _match_on(token, keywords):
     return token.match(T.Keyword, values=keywords)
 
 
-def _get_table(tokens, idx):
+def _get_tables(tokens, idx) -> Tuple[int, List[DbTableName]]:
+    def parse_ident(ident: Identifier) -> str:
```
julienledem, line 48: please add a comment on this function describing what it does

integrations/airflow/marquez_airflow/extractors/sql/parser.py
```diff
@@ -85,19 +104,57 @@ def parse(sql: str) -> SqlMeta:
         # We assume only one statement in SQL
         tokens = TokenList(statements[0].tokens)
```
julienledem, line 105: It looks like we could easily iterate on the statements if there's more than one and return all inputs and outputs

integrations/airflow/marquez_airflow/extractors/sql/parser.py
```diff
@@ -74,8 +88,13 @@ class SqlParser:
...
 
-    @staticmethod
-    def parse(sql: str) -> SqlMeta:
+    def __init__(self):
+        self.ctes = set()
+        self.intables = set()
+        self.outtables = set()
+
+    @classmethod
+    def parse(cls, sql: str) -> SqlMeta:
```
julienledem, line 97: We should pass the default schema as a parameter to enable normalizing table names

integrations/airflow/marquez_airflow/extractors/sql/parser.py
```diff
@@ -85,19 +104,57 @@ def parse(sql: str) -> SqlMeta:
...
+
+        # AS keyword
+        offset, as_keyword = group.token_next(offset, skip_ws=True, skip_cm=True)
+        if not as_keyword.match(T.Keyword, values=['AS']):
+            return [], None
+
+        offset, parens = group.token_next(offset, skip_ws=True, skip_cm=True)
+        if isinstance(parens, Parenthesis) or parens.is_group:
+            return cte_name.value, self.recurse(TokenList(parens.tokens)).in_tables
+        raise RuntimeError(f"Parens {parens} are not Parenthesis")
```
julienledem, line 160: is there a way to add the token index to the error message?

integrations/airflow/marquez_airflow/extractors/sql/parser.py
```diff
@@ -85,19 +104,57 @@ def parse(sql: str) -> SqlMeta:
...
+        _, group = tokens.token_next(idx, skip_ws=True, skip_cm=True)
+
+        if not group.is_group:
+            return [], None
+
+        # get CTE name
+        offset = 1
+        cte_name = group.token_first(skip_ws=True, skip_cm=True)
+        # handle recursive keyword
+        if cte_name.match(T.Keyword, values=['RECURSIVE']):
```
julienledem, line 149: Can you add a test of a `with recursive` querry?

integrations/airflow/marquez_airflow/extractors/sql/parser.py
```diff
@@ -85,19 +104,57 @@ def parse(sql: str) -> SqlMeta:
...
+            offset, cte_name = group.token_next(offset, skip_ws=True, skip_cm=True)
+
+        # AS keyword
+        offset, as_keyword = group.token_next(offset, skip_ws=True, skip_cm=True)
+        if not as_keyword.match(T.Keyword, values=['AS']):
+            return [], None
```
julienledem, line 155: should this be an error case?

julienledem (approved):
This looks great. 
I made a comment on normalizing table names to `{schema}.{table}` but this is independent of this change and should be a follow up change.
Can you add more comments to explain the logic in key places (for example why CTEs aliases should not be considered an input)?
I also added a few comment related to the missing `WITH RECURSIVE` test case. Codecov is pretty useful at pointing those out.

julienledem:
I have opened an issue to treat the table name normalization as a follow up: https://github.com/MarquezProject/marquez/issues/1224

wslulciuc (approved):
LGTM once PR feedback is addressed 💯

integrations/airflow/marquez_airflow/extractors/sql/parser.py
mobuchowski: Done.

integrations/airflow/marquez_airflow/extractors/sql/parser.py
mobuchowski: Maybe in following PR? I think returning those would require more changes in parser and code that uses parser, since I don't want to just union input and output datasets, they should be separate.

integrations/airflow/marquez_airflow/extractors/sql/parser.py
mobuchowski: Added.

integrations/airflow/marquez_airflow/extractors/sql/parser.py
mobuchowski: will do in following PR :+1:

integrations/airflow/marquez_airflow/extractors/sql/parser.py
mobuchowski: Added. Also, fixed, because it was bugged, so good catch :slightly_smiling_face:

integrations/airflow/marquez_airflow/extractors/sql/parser.py
mobuchowski: Yes, fixed and added test for this.

PR #1191, opened 2021-04-06
Update spark job name to reflect spark application name and execution node

collado-mike:
This change changes the job naming behavior creating a new job name for each query execution in the spark application. For each query execution, a new job is created, with a unique `runId` and a parent facet pointing to the run identified by the parameters passed into the agent. 

As an example, one job name I generated in testing was `orders_dump_to_gcs.execute_insert_into_hadoop_fs_relation_command`, where `orders_dump_to_gcs` is the spark application name and `execute_insert_into_hadoop_fs_relation_command` is the node name returned by the `DataWritingCommandExec` physical plan node.

### DIFF
--- a/integrations/spark/integrations/sparksql/4.json
+++ b/integrations/spark/integrations/sparksql/4.json
@@ -2,7 +2,7 @@
   "eventType": "COMPLETE",
   "eventTime": "2021-01-01T00:00:00Z",
   "run": {
-    "runId": "ea445b5c-22eb-457a-8007-01c7c52b6e54",
+    "runId": "fake_run_id",
     "facets": {
       "parent": {
         "_producer": "https://github.com/MarquezProject/marquez/tree/0.12.0/integrations/spark",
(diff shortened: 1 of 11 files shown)


integrations/spark/integrations/sparksql/4.json
```diff
@@ -2,7 +2,7 @@
   "eventType": "COMPLETE",
   "eventTime": "2021-01-01T00:00:00Z",
   "run": {
-    "runId": "ea445b5c-22eb-457a-8007-01c7c52b6e54",
+    "runId": "fake_run_id",
     "facets": {
       "parent": {
```
wslulciuc, line 7: _Minor_: Did you also want to update the parent runID used in the test?

wslulciuc (approved):
This looks great, @collado-mike! The naming convention now more closely aligns with our Airflow integration. Excited to see how a spark job launched via Airflow can be linked via to its parent runID (= the operator that submitted the spark job) and displayed in our lineage graph.

PR #1193, opened 2021-04-07
Update versioning for python modules

wslulciuc:
This PR simplifies the `setup.py` files for both `marquez-python` and `marquez-airflow` by controlling how versions are bumped via `setup.cfg`. The following sections were added to  `setup.cfg`:

```
[bumpversion:file:setup.py]
search = version="{current_version}"
replace = version="{new_version}"

[bumpversion:file:marquez_client/__init__.py]
search = __version__ = "{current_version}"
replace = __version__ = "{new_version}"
```

Now, when running `bump2version`, you don't need to specify the `version.py` file. For example:

```bash
$ bump2version patch
```


mobuchowski (approved):
LGTM :+1:

PR #1200, opened 2021-04-09
Eenable release via CI

wslulciuc:
This PR updates the usage [`new-release.sh`](https://github.com/MarquezProject/marquez/blob/main/new-version.sh) with the following:

```
$ ./new-version.sh
usage: ./new-version.sh <NEW_VERSION> <NEXT_VERSION>
```

where `<NEW_VERSION>` is the version of Marquez to release, and `<NEXT_VERSION>` is the next development version.

**The scripts performs the following release steps:**

1. Bump python module versions
2. Bump java module versions
3. Prepare release commit
4. Prepare release tag
5. Prepare next development version
6. Prepare next development version commit
7. Push commits and tag

After step `7`, the CI will kickoff the release pipeline triggered only via a tag. Note, since we now manage bumping versions via a script, I've removed the plugin [`gradle-release`](https://github.com/researchgate/gradle-release).

### DIFF
--- a/new-version.sh
+++ b/new-version.sh
@@ -19,12 +19,12 @@
 #   * You're on the 'main' branch
 #   * You've installed 'bump2version'
 #
-# Usage: $ ./new-version.sh <NEW_VERSION> <NEXT_VERSION>
+# Usage: $ ./new-version.sh <RELEASE_VERSION> <NEXT_VERSION>
 
 set -e
 
 usage() {
-  echo "usage: ./$(basename -- ${0}) <NEW_VERSION> <NEXT_VERSION>"
+  echo "usage: ./$(basename -- ${0}) <RELEASE_VERSION> <NEXT_VERSION>"
   exit 1
 }
 
@@ -52,11 +52,11 @@ if [[ $# -eq 0 ]] ; then
   usage
 fi
 
-NEW_VERSION="${1}"
+RELEASE_VERSION="${1}"
 NEXT_VERSION="${2}"
 
 # Ensure valid versions
-VERSIONS=($NEW_VERSION $NEXT_VERSION)
+VERSIONS=($RELEASE_VERSION $NEXT_VERSION)
 for VERSION in "${VERSIONS[@]}"; do
   if [[ ! "${VERSION}" =~ ${SEMVER_REGEX} ]]; then
     echo "Error: Version '${VERSION}' must match '${SEMVER_REGEX}'"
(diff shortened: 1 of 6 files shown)


new-version.sh
```diff
@@ -15,68 +15,74 @@
...
 
 set -e
 
-# PYTHON
+usage() {
+  echo "usage: ./$(basename -- ${0}) <NEW_VERSION> <NEXT_VERSION>"
+  exit 1
+}
+
+readonly SEMVER_REGEX="^[0-9]+(\.[0-9]+){2}(-rc\.[0-9]+)?(-SNAPSHOT)?$" # X.Y.Z
```
collado-mike, line 27: would we use both RC.X _and_ -SNAPSHOT? I'd imagine one or the other.

new-version.sh
```diff
@@ -15,68 +15,74 @@
...
+
+# Change working directory to project root
+project_root=$(git rev-parse --show-toplevel)
+cd "${project_root}"
 
 # Verify bump2version is installed
 if [[ ! $(type -P bump2version) ]]; then
- echo "bump2version not installed! Please see https://github.com/c4urself/bump2version#installation"
- exit 1
+  echo "bump2version not installed! Please see https://github.com/c4urself/bump2version#installation"
```
collado-mike, line 37: It would be nice to also put this requirement in the usage comments, so readers don't have to go through the whole script to understand requirements.

new-version.sh
```diff
@@ -15,68 +15,74 @@
...
+  exit 1
 fi
 
-if [[ -z "${type}" ]]; then
-  # Default to 'patch'
-  type="patch"
+branch=$(git symbolic-ref --short HEAD)
+if [[ "${branch}" != "main" ]]; then
+  echo "Error: You may only release on 'main'!"
```
collado-mike, line 43: Ditto re: this requirement

new-version.sh
```diff
@@ -15,68 +15,74 @@
...
-  gradle.properties
-  api/gradle.properties
-  integrations/spark/gradle.properties
-)
-
-for FILE in "${GRADLE_PROPS[@]}"
-do
-  sed -i 's/\(^version=.*$\)/version='"${NEW_VERSION}"'/g' "$FILE"
+NEW_VERSION="${1}"
+NEXT_VERSION="${2}"
```
collado-mike, line 52: To avoid confusion here, can we make them `RELEASE_VERSION` and `NEXT_VERSION`? I got confused below cause I didn't remember which was `NEW` and which was `NEXT` (short attn span :) )

collado-mike (approved):
Very awesome 👍🏽 Just a couple of minor nits, just for the easily confused, like myself 😉

new-version.sh
```diff
@@ -15,68 +15,74 @@
...
+done
+
+# (2) Bump java module versions
+sed -i "" "s/version=.*/version=${NEW_VERSION}/g" gradle.properties
+
+# (3) Prepare release commit
+git commit -am "Prepare for release ${NEW_VERSION}"
+
+# (4) Prepare release tag
+git fetch --all --tags > /dev/null 2>&1
```
phixMe, line 76: This command has failed for me in the past if the local tags are desynchronized from the remote server. I would recommend adding a `-f` (to force the fetch) or at least catching the error here.

phixMe (approved):

new-version.sh
wslulciuc: Woah! Nice catch 👍

new-version.sh
wslulciuc: Hmm, I had originally used `RELEASE_VERSION`, but second guests myself. Never again!

new-version.sh
wslulciuc: I'll catch the error 👍

PR #1222, opened 2021-04-13
Updated lineage endpoint to use latest dataset version on reads

collado-mike:
When using the OpenLineage endpoint, a new dataset version is being created every time a job reads a dataset. This causes lineage for the runs to be broken, as there is nothing tying the output of one job run to the input of another. I don't know that it really makes sense for a dataset version to be created on read if a version already exists. It makes sense to create one if a job is reading from a previously unknown dataset. But a read job is unlikely to be creating a new version of its input; more likely, any difference in the dataset properties will be due to inconsistent reporting between the reads and the writes (e.g., maybe the write schema includes all columns, but the read schema includes only a subset). This is common, for example, when reading Avro data, as the read and write schemas don't have to match as long as the data is compatible.

This also adds some test cases to validate behavior for multiple reads and writes.

### DIFF
--- a/api/src/main/java/marquez/db/OpenLineageDao.java
+++ b/api/src/main/java/marquez/db/OpenLineageDao.java
@@ -44,7 +44,9 @@ import marquez.db.models.UpdateLineageRow;
 import marquez.db.models.UpdateLineageRow.DatasetRecord;
 import marquez.service.models.LineageEvent;
 import marquez.service.models.LineageEvent.Dataset;
+import marquez.service.models.LineageEvent.DatasetFacet;
 import marquez.service.models.LineageEvent.Job;
+import marquez.service.models.LineageEvent.SchemaDatasetFacet;
 import marquez.service.models.LineageEvent.SchemaField;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import org.jdbi.v3.sqlobject.transaction.Transaction;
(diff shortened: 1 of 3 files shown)


api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -366,26 +368,42 @@ default DatasetRecord upsertLineageDataset(
...
-            datasetVersionDao.toPgObjectSchemaFields(fields),
-            dsNamespace.getName(),
-            ds.getName());
-
-    datasetDao.updateVersion(datasetRow.getUuid(), now, datasetVersionRow.getUuid());
-
+        datasetRow
+            .getCurrentVersionUuid()
+            .filter(v -> isInput) // only fetch the current version if this is a read
+            .flatMap(datasetVersionDao::findRowByUuid)
```
julienledem, line 381: in this case I feel an if statement or ternary operator would be more readable.
```
final DatasetVersionRow datasetVersionRow;
if (isInput) {
  datasetVersionRow = datasetRow
            .getCurrentVersionUuid()
            .map(datasetVersionDao::findRowByUuid)
} else {
  UUID versionUuid = version(...);
  datasetVersionRow = datasetVersionDao.upsert(...)
  datasetDao.updateVersion(...)
}
```

julienledem (approved):
LGTM!
Just some cosmetic comment

wslulciuc (approved):
Sweet work, @collado-mike 💯

PR #1256, opened 2021-04-21
Rewrite lineage IO sql queries to avoid job_versions_io_mapping_* tables

collado-mike:
Simplify the SQL queries that read the job and dataset lineage for the `LineageService`. This eliminates the reads from the `job_versions_io_mapping_outputs` and `job_versions_io_mapping_inputs` tables and reads all the IO from the `job_versions_io_mapping` table only. It also adds some pretty comprehensive test coverage for the lineage queries to verify we're returning the right jobs and datasets in different cases.

### DIFF
--- a/api/src/main/java/marquez/db/LineageDao.java
+++ b/api/src/main/java/marquez/db/LineageDao.java
@@ -44,12 +44,17 @@ public interface LineageDao {
    * @return
    */
   @SqlQuery(
+      // dataset_ids: all the input and output datasets of the current version of the specified jobs
       "WITH dataset_ids AS (\n"
           + "    SELECT DISTINCT dataset_uuid\n"
           + "        FROM jobs j\n"
           + "        INNER JOIN job_versions_io_mapping io ON io.job_version_uuid=j.current_version_uuid\n"
           + "        WHERE j.uuid IN (<jobIds>)\n"
           + ")\n"
+          // SELECT DISTINCT j.uuid:
+          // all the jobs that have ever read from or written to those datasets.
+          // note that this includes previous versions of jobs that no longer interact with those
+          // datasets
           + "SELECT DISTINCT j.uuid\n"
           + "    FROM job_versions_io_mapping io\n"
           + "    INNER JOIN dataset_ids ON io.dataset_uuid=dataset_ids.dataset_uuid\n"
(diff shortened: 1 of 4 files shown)


api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -33,81 +33,85 @@
...
+          + "SELECT ds.*, COALESCE(job_ids, '{}') AS job_ids, dv.fields\n"
+          + "FROM datasets ds\n"
+          + "    INNER JOIN dataset_ids ON dataset_ids.dataset_uuid=ds.uuid\n"
+          + "    LEFT JOIN (SELECT dataset_uuid, ARRAY_AGG(DISTINCT v.job_uuid) AS job_ids\n"
+          + "        FROM job_versions_io_mapping io2\n"
+          + "        INNER JOIN job_versions v on io2.job_version_uuid = v.uuid\n"
+          + "        WHERE io2.io_type='OUTPUT'\n"
+          + "        GROUP BY dataset_uuid\n"
+          + "        ) io ON io.dataset_uuid=ds.uuid\n"
+          + "    LEFT JOIN dataset_versions dv on dv.uuid = ds.current_version_uuid;")
```
julienledem, line 114: With this change using the job_versions_io_mapping table again it looks like this could be a single query that aggregates inputs and outputs in separate fields. 
Just a thought, this could be a follow up item.

julienledem (approved):
This looks great

api/src/main/java/marquez/db/LineageDao.java
collado-mike: that makes sense. I'll add a todo

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -33,81 +33,85 @@
...
+          + "    SELECT DISTINCT dataset_uuid\n"
+          + "        FROM jobs j\n"
+          + "        INNER JOIN job_versions_io_mapping io ON io.job_version_uuid=j.current_version_uuid\n"
+          + "        WHERE j.uuid IN (<jobIds>)\n"
+          + ")\n"
+          + "SELECT DISTINCT j.uuid\n"
+          + "    FROM job_versions_io_mapping io\n"
+          + "    INNER JOIN dataset_ids ON io.dataset_uuid=dataset_ids.dataset_uuid\n"
+          + "    INNER JOIN job_versions jv ON jv.uuid=io.job_version_uuid\n"
+          + "    INNER JOIN jobs j ON j.uuid=jv.job_uuid")
```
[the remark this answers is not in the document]
collado-mike: line 53 is actually
```
// SELECT DISTINCT j.uuid:
// all the jobs that have ever read from or written to those datasets.
// note that this includes previous versions of jobs that no longer interact with those
// datasets
```

PR #1257, opened 2021-04-22
Added docs and diagrams to illustrate dataset version creation from job run status

collado-mike:


STATE_TRANSITIONS.MD
```diff
@@ -0,0 +1,224 @@
+# Run State Transitions
+Run states change based on the success or failure of a job run. The datasets consumed and/or produced
+by a job run are immutable and do not change based on the success or failure of the run.
+
+# Dataset State Transitions
+Dataset states change based on the success or failure of job runs that consume or generate those 
+datasets. Typically, Runs generate dataset versions as outputs, but sometimes a job run may result 
+in an input dataset version being created if that dataset did not previously exist in the database.
+Once a dataset exists, a new version will only be created if the dataset is the output of a job run.
```
julienledem, line 9: The cases bellow expose all the possible paths taken in an algorithm.
Could you add a summary at the top with the pseudo code? That will make it easier to follow each of the individual paths bellow.
I would expect a few nested if-else that describe this logic.

STATE_TRANSITIONS.MD
```diff
@@ -0,0 +1,224 @@
...
+* Then
+    * Create JobA@Run1
+    * Create JobB@Run1
+    * Create DatasetX@V1
+    * Set DatasetX current version to V1
+    * Create DatasetY@V1
+    * Set DatasetY current version to V1
+    * JobA@Run1 produces DatasetX@V1
+    * JobB@Run1 consumes DatasetX@V1
+    * JobB@Run1 produces DatasetY@V1
```
julienledem, line 81: Please order this chronologically, that will be easier to follow.

STATE_TRANSITIONS.MD
```diff
@@ -0,0 +1,224 @@
...
+* Given
+    * JobA does not exist
+    * JobB does not exist
+    * DatasetX does not exist
+    * DatasetY does not exist
+* When
+    * JobA@Run1 **succeeds**
+    * JobA produces DatasetX
+    * JobB consumes DatasetX
+    * JobB produces DatasetY
```
julienledem, line 71: In previous cases we have `DS X -> Job A -> DS Y`
I would assume we want to keep the same shape in all cases.
So the case should be `DS X -> Job A -> DS Y -> Job B -> DS Z`, correct?

julienledem: summarizing this at the top would help

STATE_TRANSITIONS.MD
```diff
@@ -0,0 +1,224 @@
...
+    * Create JobA@Run1
+    * Create JobB@Run1
+    * Create DatasetX@V1
+    * Set DatasetX current version to V1
+    * Create DatasetY@V1
+    * Set DatasetY current version to V1
+    * JobA@Run1 produces DatasetX@V1
+    * JobB@Run1 consumes DatasetX@V1
+    * JobB@Run1 produces DatasetY@V1
+![Consume and produce dataset](./docs/assets/dot/job_input_dataset_creation.dot.png)
```
julienledem, line 82: the image does not match the cases `JobB@Run1 produces DatasetY@V` is wrong? Do you mean consume?
https://github.com/MarquezProject/marquez/blob/state_transition_docs/docs/assets/dot/job_input_dataset_creation.dot.png

julienledem (commented):
Before going in all the possible paths, I think it would help a lot, to define Job A, B, C and DS X, Y, Z upfront.
Let's keep it consistent and summarize the algorithm at the top. Then the individual cases can illustrate that logic.

STATE_TRANSITIONS.MD
```diff
@@ -0,0 +1,227 @@
+# Run State Transitions
+Run states change based on the success or failure of a job run. The datasets consumed and/or produced
+by a job run are immutable and do not change based on the success or failure of the run.
+
+# Dataset State Transitions
```
wslulciuc, line 5: Though I agree that datasets have _states_, I think the title can be confused as _states_ is an overloaded term in the context of a run. Mind updating the title to something like: _Dataset Versioning Lifecycle_ or something similar?

collado-mike: My thinking was primarily around documenting like a state machine, hence the wording. I think something like _Dataset Versioning Lifecycle_ sounds good, though.

wslulciuc (changes requested):
These docs our amazing (❤️ the diagrams), @collado-mike! Mind if we move the markdown to `docs/` (as per our [documentation](https://github.com/MarquezProject/marquez#documentation) section in the `README.md`) and rename the file to `run-state-transitions.md` to be consistent with the docs that are? Otherwise, the flows you outlined are super clear 👍

wslulciuc (approved):
LGTM 👍

PR #1258, opened 2021-04-22
Updated OpenLineage impl to only update dataset version on run completion

collado-mike:
This addresses the issue of dataset versions being created and updated when Job Runs fail. This breaks the lineage for a downstream run, since a consumer job will see the latest dataset version created, but the output of a job run is only updated if the job run completes. Thus, the consumed dataset version will have no upstream producer, so the consumer's lineage isn't tied to anything. 

I think we established in an offline discussion that we'd allow the dataset version creation to continue in the OpenLineageDao, as it is now, but that the dataset `currentVersionUuid` change will only happen on run completion. This means that we get a record of a dataset version being created, but downstream consumers won't see the new version unless the run that created it completes.

Added a test case that illustrates the new behavior.

Signed-off-by: Michael Collado <mike@datakin.com>

### DIFF
--- a/api/src/test/java/marquez/service/OpenLineageServiceTest.java
+++ b/api/src/test/java/marquez/service/OpenLineageServiceTest.java
@@ -1,5 +1,6 @@
 package marquez.service;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.mock;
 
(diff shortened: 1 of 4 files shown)


wslulciuc (approved):
LGTM 👍

api/src/test/java/marquez/service/OpenLineageServiceTest.java
```diff
@@ -196,6 +216,118 @@ public void serviceCalls(List<URI> uris, ExpectedResults expectedResults) {
...
+                .build())
+        .get();
+
+    // still version 1 is consumed since the second producer job run failed
+    assertThat(datasetVersionDao.findInputsByRunId(secondRunId))
+        .hasSize(1)
+        .map(DatasetVersionRow::getVersion)
+        .contains(dsVersion1Id);
+  }
+
```
wslulciuc, line 330: Solid test! 💯

PR #1259, opened 2021-04-23
Great expectations based data quality facet support 

mobuchowski:
Currently, supporting only GreatExpectationsOperator, extracting valid data only if users defined it in their GreatExpectations config.


integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+                        result = parser.parse_expectation_result(expectation)
+                        facet_data[result.facet_key] = result.value
+                for parser in _COLUMN_EXPECTATIONS_PARSER:
+                    if parser.can_accept(expectation):
+                        result = parser.parse_expectation_result(expectation)
+                        facet_data['columnMetrics'][result.column_id][result.facet_key] \
+                            = result.value
+
+            for key in facet_data['columnMetrics'].keys():
+                facet_data['columnMetrics'][key] = ColumnMetric(**facet_data['columnMetrics'][key])
```
collado-mike, line 131: If I read this right, we're only collecting the actual observed values for everything and sticking it in the `ColumnMetric` type, is that right? Should we be collecting the assertions and the results?

collado-mike (commented):
Nice. I have a question about the column data we're collecting and whether or not the metrics we have are all we want from this integration

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
mobuchowski: I assumed that we want to implement the API presented in the google doc, and the expectations themselves are just a way to get this data. If we want to collect them, then sure, we can.

julienledem: I would add a greatexpectations custome facet for the assertions

mobuchowski: Let's do it in follow up PR then

collado-mike (approved):

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -0,0 +1,347 @@
...
+            )
+            # To match dataset name we need canonical datasource name
+            name = batch_kwargs.get('datasource', None)
+            path = batch_kwargs.get('path', None)
+            return Dataset(
+                source=Source(
+                    name=name,
+                    type='GREAT_EXPECTATIONS',
+                    connection_url=path
+                ),
```
collado-mike, line 100: Will the `type` argument here impact the ability to correctly match the dataset? I think in the marquez impl, the answer is no, but should that be guaranteed by the Open Lineage spec?

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def extract_on_complete(self, task_instance) -> Optional[StepMetadata]:
+        if self.result:
+            dataset = self.parse_result()
+            if not dataset:
+                return None
+            return StepMetadata(
+                name=get_job_name(task=self.operator),
+                outputs=[dataset]
+            )
```
collado-mike, line 154: I think this should be inputs, not outputs. The Great Expectations job doesn't create the dataset- it reads it. Marking the dataset as an output will create a new dataset version, which may or may not share lineage with other downstream consumers (depending on which job completes first).

collado-mike (commented):
Added a couple of thoughts on impl

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
mobuchowski: So, we don't actually use it in this way. DataSource facet requires just name and connection_url... 
I think we should get rid of type here, or build the URL somewhere later. It does not add anything if we already have URL conformant to https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -43,20 +40,27 @@
...
+            self.extractor_mapper = extractor_mapper
+        else:
+            self.extractor_mapper = Extractors()
+
+    def add_task(self, task):
+        super().add_task(task)
+
+        # Purpose: some extractors, called patchers need to hook up to internal components of
+        # operator to extract necessary data. The hooking up is done on instantiation
+        # of extractor TODO ADD
```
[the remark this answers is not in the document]
mobuchowski: Done

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+class ExpectationsParserResult:
+    """
+    Internal class to represent actual expectation values, per table and optionally per column
+    """
+    facet_key: str = attr.ib()
+    value: Any = attr.ib()
+    column_id: Optional[str] = attr.ib(default=None)
+
+
+class GreatExpectationsExtractorImpl(BaseExtractor):
```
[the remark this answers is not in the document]
mobuchowski: Done

integrations/airflow/marquez_airflow/extractors/extractors.py
```diff
@@ -0,0 +1,44 @@
...
+    BigQueryExtractor,
+    GreatExpectationsExtractor
+]
+
+_patchers = [
+    GreatExpectationsExtractor
+]
+
+
+class Extractors:
```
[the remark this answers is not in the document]
mobuchowski: Done

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -117,7 +121,7 @@ def _register_dagrun(self, dagrun: DagRun, is_external_trigger: bool, execution_
                     exc_info=True)
 
     def handle_callback(self, *args, **kwargs):
-        self.log.debug(f"handle_callback({args}, {kwargs})")
+        self.log.info(f"handle_callback({args}, {kwargs})")
```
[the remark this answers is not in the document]
mobuchowski: Done

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+                ),
+                name=name,
+                type=DatasetType.DB_TABLE,
+                custom_facets={
+                    'dataQuality': data_quality_facet
+                }
+            )
+
+        except ValueError:
+            pass
```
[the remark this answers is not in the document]
mobuchowski: Done.

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
collado-mike: yeah, we can search for input facets for a given dataset version, so there shouldn't be any concern there.

collado-mike (approved):

integrations/airflow/tests/integration/docker/up.sh
```diff
@@ -52,4 +52,5 @@ marquez-python
 ${MARQUEZ_AIRFLOW_WHL}
 EOL
 
+docker-compose down
 docker-compose up --build --exit-code-from integration
```
wslulciuc, line 56: I would use the [`--force-recreate`](https://docs.docker.com/compose/reference/up/) flag to force recreating the containers if they are running.

wslulciuc (approved):
Left a minor comment on the `docker/up.sh` used for the airflow integration test, otherwise really great work on adding support for _Great Exceptions_, @mobuchowski! 💯 🥇

PR #1261, opened 2021-04-23
Add openlineage facets support to API response models 

wslulciuc:
This PR adds support to return OpenLineage facets to the API response models (`Dataset`, `Job`, and `Runs`). For example, if we send an OpenLineage event with dataset facets, you can now retrieve the dataset meta with facets returned:

### `GET` `/api/v1/namespaces/my-namespace/datasets/my-dataset` 
```
{
  "id": {
    "namespace": "my-namespace",
    "name": "my-dataset"
  },
  "type": "DB_TABLE",
  "name": "my-dataset",
  "physicalName": "my-dataset",
  "createdAt": "2020-12-29T03:51:01.641Z",
  "updatedAt": "2020-12-29T03:51:01.641Z",
  "namespace": "my-namespace",
  "sourceName": "string",
  "fields": [
    {
      "name": "column1",
      "type": "VARCHAR",
      "tags": [],
      "description": "string"
    }
  ],
  "tags": [],
  "lastModifiedAt": null,
  "description": "canonical representation of entity Foo",
  "facets": {
    "schema": {
      "fields": [
        {
          "name": "column1",
          "type": "VARCHAR",
          "description": "string"
        }
      ],
      "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
      "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet"
    },
    "dataSource": {
      "uri": "string",
      "name": "string",
      "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
      "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet"
    },
    "documentation": {
      "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
      "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet",
      "description": "canonical representation of entity Foo"
    },
    "additionalProp1": {
      "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
      "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet",
      "additionalProp1": {}
    },
    "additionalProp2": {
      "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
      "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet",
      "additionalProp1": {}
    }
  }
}
```


api/src/main/java/marquez/db/JobDao.java
```diff
@@ -60,11 +60,17 @@
   void updateVersion(UUID rowUuid, Instant updatedAt, UUID currentVersionUuid);
 
   String JOB_SELECT =
-      "SELECT j.*, jc.context "
+      "SELECT j.*, jc.context, "
+          + "(SELECT event->'job'->'facets' "
+          + "   FROM lineage_events AS le "
+          + "  WHERE jv.latest_run_uuid::text = le.run_id "
+          + "  ORDER BY event_time DESC "
+          + "  LIMIT 1) AS facets "
```
collado-mike, line 68: I think the lineage events are supposed to be cumulative for a run. So it's conceivable that a user may post multiple partial lineage events, with each event contributing to the whole. This `LIMIT 1` assumes the latest event contains all of the information, which isn't necessarily so, right?

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -79,7 +79,12 @@
   void updateEndState(UUID rowUuid, Instant transitionedAt, UUID endRunStateUuid);
 
   String SELECT_RUN =
-      "SELECT r.*, ra.args, ra.args, ctx.context "
+      "SELECT r.*, ra.args, ra.args, ctx.context, "
+          + "(SELECT event->'run'->'facets' "
+          + "   FROM lineage_events AS le "
+          + "  WHERE r.uuid::text = le.run_id "
+          + "  ORDER BY event_time DESC "
+          + "  LIMIT 1) AS facets "
```
collado-mike, line 87: Same comments as above

api/src/main/java/marquez/db/mappers/DatasetMapper.java
```diff
@@ -88,12 +93,13 @@ public Dataset map(@NonNull ResultSet results, @NonNull StatementContext context
...
           stringOrNull(results, Columns.DESCRIPTION),
-          Optional.ofNullable(uuidOrNull(results, Columns.CURRENT_VERSION_UUID)));
+          Optional.ofNullable(uuidOrNull(results, Columns.CURRENT_VERSION_UUID)),
+          toFacetsOrNull(results));
     }
   }
 
   private URL getUrl(ResultSet results, String column) throws SQLException, MalformedURLException {
-    if (!hasColumn(results, column)) {
+    if (!Columns.columnExists(results, column)) {
```
collado-mike, line 102: If the tendency is to always use the class name with the method, as opposed to a static import, I'd prefer to write it as `Columns.exists(results, column)`

api/src/main/java/marquez/db/mappers/DatasetVersionMapper.java
```diff
@@ -65,7 +66,8 @@ public DatasetVersion map(@NonNull ResultSet results, @NonNull StatementContext
               toFields(results, "fields"),
               columnNames.contains("tags") ? toTags(results, "tags") : null,
               stringOrNull(results, Columns.DESCRIPTION),
-              null);
+              null,
+              toFacetsOrNull(results));
```
collado-mike, line 70: Can we use `Optional<Facet>`? Looks like Jackson has support for it in the `jackson-datatype-jdk8` package

api/src/main/java/marquez/db/mappers/MapperUtils.java
```diff
@@ -1,15 +1,26 @@
...
 import java.util.Set;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
+import marquez.common.Utils;
+import marquez.common.models.Facets;
+import marquez.db.Columns;
 
 @Slf4j
-public class MapperUtils {
+public final class MapperUtils {
```
collado-mike, line 20: I think there's a `@Utils` lombok annotation to generate final classes with private constructors

api/src/main/java/marquez/service/models/Dataset.java
```diff
@@ -61,6 +62,7 @@
   @Nullable private final Instant lastModifiedAt;
   @Nullable private final String description;
   private final Optional<UUID> currentVersionUuid;
+  @Nullable Facets facets;
```
collado-mike, line 65: It would be nice if we can just use `Optional` fields and let Jackson handle the de/serialization

clients/java/src/main/java/marquez/client/models/Run.java
```diff
@@ -45,14 +46,16 @@ public Run(
       @Nullable final Instant startedAt,
       @Nullable final Instant endedAt,
       @Nullable final Long durationMs,
-      @Nullable final Map<String, String> args) {
+      @Nullable final Map<String, String> args,
+      @Nullable final Object facets) {
```
collado-mike, line 50: Why are the facets all `Object` in the client?

collado-mike (commented):
We should answer the question about cumulative lineage events. My understanding is that events should be cumulative in the service, so that we should always return the full collection of facets we've received for a given run. If that's not the case, ever client has to keep track of every facet that's supposed to be returned for a given run, which, I think, has impact on the ability of users to return custom facets, e.g., from a Spark job, without modifying the integration to specifically do so.

clients/java/src/main/java/marquez/client/models/Run.java
wslulciuc: I think it'd make sense to use `Map<String, Object>` instead. I'll make the change 👍

api/src/main/java/marquez/db/mappers/MapperUtils.java
wslulciuc: Woah, very cool. Let's use `@UtilityClass`!

api/src/main/java/marquez/db/mappers/DatasetMapper.java
wslulciuc: Sure, reduces the wordiness 👍

api/src/main/java/marquez/common/models/Facets.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.util.StdConverter;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+
+/** Defines the {@code facets} type. */
```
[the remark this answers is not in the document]
wslulciuc: I realized after trying to write some docs on why we need to keep the `JsonNode` around, we should use `ImmutableMap<String, Object>` instead and avoid defining the `Facet` type all together.

api/src/main/java/marquez/db/mappers/DatasetVersionMapper.java
wslulciuc: We can begin using optionals as parameters or drop them all together. I'll open an issue about uses of optionals in the code.

api/src/main/java/marquez/db/DatasetVersionDao.java
```diff
@@ -170,19 +170,33 @@ default void updateDatasetVersionMetric(
...
-          + "from datasets d\n"
-          + "inner join dataset_versions dv on d.uuid = dv.dataset_uuid\n"
-          + "left outer join stream_versions sv on sv.dataset_version_uuid = d.current_version_uuid\n";
+  String BASE_DATASET_VERSION_SELECT =
+      "SELECT d.type, d.name, d.physical_name, dv.namespace_name, d.source_name, d.description, "
+          + "dv.created_at, dv.version, dv.fields, dv.run_uuid AS \"createdByRunUuid\", sv.schema_location, "
+          + "ARRAY(SELECT t.name FROM tags AS t "
+          + "       INNER JOIN datasets_tag_mapping AS m ON m.tag_uuid = t.uuid "
+          + "       WHERE d.uuid = m.dataset_uuid) AS tags, "
+          + "(SELECT JSON_AGG(tmp_facets) "
```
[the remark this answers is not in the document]
wslulciuc: Nice! I'll make the change 👍

PR #1277, opened 2021-04-29
Changed LineageDao to stop included jobs that previously were connected...

collado-mike:
to a graph, but no longer are


wslulciuc (approved):
Nice! 💯

wslulciuc:
@collado-mike Cool to merge, but looks like a lineage test is failing.

collado-mike:
> @collado-mike Cool to merge, but looks like a lineage test is failing.

Interestingly, this only started failing after I rebased on main after https://github.com/MarquezProject/marquez/commit/8636604e5dcffe1832ea39d9b72e9bbab7b272db was merged. It makes sense that this test would start failing, but I'm surprised it didn't fail in your PR. Haven't looked too deeply, but I think we _do_ expect failed jobs to be returned in the lineage now that your change is in. I've updated the tests accordingly and added a case for incomplete jobs to ensure they are still excluded.

julienledem (approved):

PR #1282, opened 2021-05-04
Updated dataset lineage query to find most recent job that wrote to it

collado-mike:
Currently, the dataset lineage query ends up picking the first job that touched the dataset- in some cases, this includes jobs that queried the dataset a long time ago and have never run since or have changed their input/output lineage. This changes the query to prefer the most recent job that wrote to a dataset, so that upstream lineage is visible.


julienledem (approved):
That makes sense. Thanks

PR #1286, opened 2021-05-06
Fix web deps

wslulciuc:
This PR fixes the package version for libs that causes the web UI to fail loading.


phixMe (approved):
Looks great! Thanks @wslulciuc

julienledem (commented):
👍

PR #1288, opened 2021-05-06
Unpin requests dep in airflow integration

wslulciuc:
Only Airflow `1.10.15` has `2.25.1` of `requests` installed, unpinning to fix broken integration for later versions of Airflow.

### DIFF
--- a/integrations/airflow/setup.py
+++ b/integrations/airflow/setup.py
@@ -21,7 +21,7 @@ with open("README.md") as readme_file:
 
 requirements = [
     "attrs==19.3",
-    "requests==2.25.1",
+    "requests>=2.25.1",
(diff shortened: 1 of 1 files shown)


integrations/airflow/setup.py
```diff
@@ -21,7 +21,7 @@
 
 requirements = [
     "attrs==19.3",
-    "requests==2.25.1",
+    "requests>=2.25.1",
```
julienledem, line 24: I think here we need to adjust it to a lower version.

wslulciuc: Yeah, I've lowered it to `2.24.0`

julienledem (approved):

PR #1304, opened 2021-05-12
Augment tutorial instructions & screenshots

rossturk:
Signed-off-by: Ross Turk <ross@rossturk.com>

This is a small set of changes that I think will help clarify the example. Included
are a screenshot of Airflow that shows the DAGs correctly running, some additions
to the sample `tree` output, and a bit of text that makes it clear that users should
start in examples/airflow.


wslulciuc (approved):
Thanks for improving the airflow tutorial, @rossturk!

PR #1305, opened 2021-05-12
Add HttpClient argument to MarquezClient

lewiesnyder:
Added HttpClient to MarquezClient and MarquezHttp to allow for special configurations. Example: supplying key or trust store for TLS via SSLContext.

Also exposed the UserAgent value to that it can be added to the supplied HttpClient for consistency as it can't be added to an existing HttpClient.


KevinMellott91 (commented):
Looks good to me - ability to make calls over HTTPS without impacting existing client functionality. Thanks @lewiesnyder for the add!

wslulciuc (changes requested):
@lewiesnyder, thanks for adding SSL support! There's only one minor change I would make in your implementation. The `marquez-java` client should encapsulate the instantiation of the underlying `HttpClient`. In the future, we'll want to allow callers to pass in `MarquezClientConfig` to the client that would enable overrides for: _conn timeouts_, _req timeouts_, _max retries_, etc.

To ensure the `MarquezClient` instantiates the underlying client consistently, I would have the caller pass in the `SSLContext`:

```java
MarquezClient client = MarquezClient.builder()
    .sslContext(sslContext)
    .baseUrl("https://localhost:5000")
    .build();
```

The caller is then only responsible for creating the SSL context. This would also ensure the user agent is correctly set and also allow use to swap out the underlying `HttpClient` with, for example, [OkHttp](https://github.com/square/okhttp) if we wanted.

I would simply overload the `MarquezHttp.create()` factory method to accept an `SSLContext` object that can be used by the [client builder](https://github.com/MarquezProject/marquez/blob/main/clients/java/src/main/java/marquez/client/MarquezClient.java#L316):

```java
MarquezHttp.create(SSLContext, MarquezClient.Version, String)
```

lewiesnyder:
@wslulciuc thanks for the feedback. Passing SSLContext was my original solution but opted for HttpClient to provide more flexibility for the implementation.

I should be able to have this ready later today.

wslulciuc:
> Passing SSLContext was my original solution but opted for HttpClient to provide more flexibility for the implementation.

What sort of flexibility did you have in mind? Mainly around configuring the client?

lewiesnyder:
@wslulciuc just saw this comment now.
Really nothing in particular other than SSL, as an engineer I like to have options

PR #1309, opened 2021-05-14
Add SSLContext to MarquezClient

lewiesnyder:
Add SSLContext argument to MarquezClient create method to allow for SSL configurations


wslulciuc:
@lewiesnyder: So, it didn't work! I'll make an exception and approve the DCO check, but will look into why it didn't work.

wslulciuc (approved):
LGTM 👍

lewiesnyder:
@wslulciuc awesome, I'm not familiar with DCO but definitely curious what the deal is here.

thanks for the help!

PR #1353, opened 2021-05-24
Refactored ValuesAverageExpectationParser to sum and count

collado-mike:
For parsing and displaying metrics, I think collecting count and sum separately is more versatile and useful than just reporting average. Average is easy to compute from count and sum, but it's impossible to go the other way.


mobuchowski:
The change makes sense. You need to update `ColumnMetric` definition as well - now it does not expect `sum` to be there.

mobuchowski (changes requested):
The change makes sense. Though, you need to update ColumnMetric definition as well - now it does not expect sum to be there.

mobuchowski (approved):

wslulciuc (commented):
LGTM 👍

PR #1376, opened 2021-06-01
Fix path for imgs in run-state-transitions.md

wslulciuc:
This PR fixes the relative image paths in `run-state-transitions.md`.


collado-mike (approved):
🤦🏽‍♂️

PR #1379, opened 2021-06-02
Replace run calls with OpenLineage and add data model section

wslulciuc:
This PR updates our quickstart guide to use OpenLineage and adds a section on the Marquez data model as well as other helpful clarifications.


rossturk (approved):
Looks good to me. I think these are solid improvements.

PR #1387, opened 2021-06-08
Airflow task run macro

collado-mike:
Added a macro function that returns the generated run_id for a task to forward on to child tasks. E.g., in the Spark documentation, the run_id is a randomly generated UUID that has no relation to the original airflow dag/task. The macro allows access to the generated run id so that parent and child tasks maintain their hierarchy when the Spark task reports its parent run facet. 

I did notice that the previously used function to generate task ids was replaced with one that simply concatenates dag_id and task_id. Per discussion in https://github.com/OpenLineage/OpenLineage/issues/29 , we should be generating valid UUIDs for the task run ids.


integrations/airflow/marquez_airflow/utils.py
```diff
@@ -157,6 +158,10 @@ def get_custom_facets(task, is_external_trigger: bool):
     }
 
 
+def new_run_id(dag_run_id: str, task_id: str) -> str:
```
mobuchowski, line 161: Does this function needs parameters now?

mobuchowski (approved):
Good idea with macro. One question, otherwise looks good.

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -41,9 +40,44 @@
 _MARQUEZ = MarquezAdapter()
 
 
+@provide_session
+def task_run_id(run_id, task, session=None):
```
wslulciuc, line 44: Given that Airflow also has a notion of task runs, we might want to be explicate here and name the function `marquez_task_run_id` or similar.

integrations/airflow/marquez_airflow/dag.py
```diff
@@ -158,7 +192,7 @@ def _report_task_instance(self, task_instance, dagrun, session):
         step = self._extract_metadata(dagrun, task, task_instance)
 
         job_name = self._marquez_job_name(self.dag_id, task.task_id)
-        run_id = self._marquez_run_id(dagrun.run_id, task.task_id)
+        run_id = new_run_id(dagrun.run_id, task.task_id)
```
wslulciuc, line 195: Similar comment as above, maybe `new_marquez_run_id` for clarity?

integrations/airflow/marquez_airflow/utils.py
collado-mike: I added the parameters so that the tests could override the function. I suppose I could have changed the test to return a list of expected return values, but I thought it would be simpler with a deterministic function so we don't have to keep track of which ids are returned in which order.

integrations/airflow/marquez_airflow/dag.py
collado-mike: I did have a similar thought. I left it as is just for brevity, but it's probably better to be more explicit.

wslulciuc (approved):
LGTM 👍

PR #1389, opened 2021-06-09
Updated SparkListener to extend Spark's SparkListener abstract class

collado-mike:
This change alters the SparkListener class to extend the real SparkListener abstract class and changes the method signatures to match the listener interface. This avoids the unnecessary Proxy class generation. Further, it uses strictly information already available from the listener events to access job information, avoiding the need to weave bytecode into the ActiveJob class. This allows users to get started in a more familiar and intuitive manner, such as 

```python
spark = (SparkSession.builder.master('local').appName('sample_spark')
             .config('spark.jars.packages', 'io.github.marquezproject:marquez-spark:0.15.+,org.postgresql:postgresql:42.2.+')
             .config('spark.extraListeners', 'marquez.spark.agent.SparkListener')
             .config('openlineage.url', 'http://marquez:5000/api/v1/namespaces/spark_integration/?api_key=XYZ')
             .config('spark.sql.repl.eagerEval.enabled', 'true')
             .getOrCreate())
```

The `marquez-spark` jar can be downloaded directly from maven, with support for version ranges, as supported by Ivy (see https://ant.apache.org/ivy/history/2.1.0/settings/version-matchers.html ) so users don't have to go through the hassle of downloading the jar and updating their configuration for each upgrade (users can, of course, pin versions for dependencies by specifying exactly the jar they want). Platform owners can construct Spark clusters and embed these parameters in the configuration files deployed to their cluster. That would allow every Spark job that runs on the cluster to report lineage without the application developers adding any custom configuration (parent run ids still have to be passed in via configuration in the case of airflow job submission). This also enables users of serverless Spark platforms, like AWS Glue, to report lineage from their Spark jobs without needing access to the JVM startup parameters.


wslulciuc (approved):
Love ❤️ the cleaner approach! Great work, @collado-mike!

PR #1398, opened 2021-06-11
Add endpoint to create tags

hanbei:
- Created an endpoint PUT /api/v1/tags that allows to create tags.
- Allow to create tags using MarquezCLient
- Test new tag creation api

Signed-off-by: Florian Schulz <florian.schulz@hanbei.de>


wslulciuc (changes requested):
Thanks for the contribution, @hanbei! Mind also adding the create tag method to the mqz [`python`](https://github.com/MarquezProject/marquez/tree/main/clients/python) client?

api/src/main/java/marquez/api/TagResource.java
```diff
@@ -51,6 +54,17 @@ public Response list(
     return Response.ok(new Tags(tags)).build();
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @PUT
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  public Response create(@Valid Tag tag) {
```
wslulciuc, line 63: I'd use the path `/tags/{name}` with the HTTP body as the tag description for consistency with the other endpoints. And, agree that this can be a `POST`, but making it a `PUT` makes a lot of sense given that the caller is providing the unique identifier for the tag.

hanbei: The HTTP body still as json in the form
```
{
  "description" : "some string"
}
```
or plain text?

wslulciuc: Json, as it's likely that we'll add properties in the future.

hanbei:
I added the method to the python client and wrote tests for the tags methods

api/src/test/java/marquez/TagIntegrationTest.java
```diff
@@ -29,6 +30,29 @@ public void testApp_testTags() {
...
+    assertThat(tagWithoutDescription.getDescription()).isEmpty();
+  }
+
+  @Test
+  public void testApp_testUpsertTag() {
+    Tag tag = client.createTag("tag", "Description");
+    assertThat(tag.getName()).isEqualTo("tag");
+    assertThat(tag.getDescription()).contains("Description");
+
+    Tag tagWithoutDescription = client.createTag("tag", "New Description");
```
hanbei, line 50: Maybe I found a bug in TagDao. Is it correct behaviour that I can't change the description after the initial creation of the tag.

wslulciuc: Nope, it definitely should! Mind opening an issue? And feel free to assign it to yourself if you wanted to take it on :)

clients/python/tests/test_marquez_client.py
```diff
@@ -857,3 +868,47 @@ def test_mark_job_run_as_failed(mock_post, client):
...
+        headers=mock.ANY,
+        json={
+            "description": DESCRIPTION
+        },
+        timeout=mock.ANY
+    )
+
+
+@mock.patch('requests.get')
+def test_list_tags(mock_get, client):
```
wslulciuc, line 896: Thanks for adding this test! 💯

clients/python/marquez_client/client.py
```diff
@@ -348,6 +348,16 @@ def list_tags(self, limit=None, offset=None):
             }
         )
 
+    def create_tag(self, name, description=None):
+        payload = {
```
wslulciuc, line 352: _Minor:_ We may want to check that the `description` is not `None`, otherwise the `None` value would be set to the description in the payload?

wslulciuc (approved):
This looks great, @hanbei! Left a minor comment that we may want to address, otherwise 👍

clients/python/marquez_client/client.py
hanbei: A `None` value will be transformed to `null` in the payload json. According to the `tag` schema the description may be null and the java api also allows a `null` value here. It explicitly uses an Optional as return value for getDescription.  
We could use `""` (empty string) as the default value but that would kind of break the contract in the java api. Not sure though if that is the intended behaviour of the tags api

wslulciuc: Sounds good. As long as `None` maps to a `null`, that's the intended behavior. My comment was more around whether we wanted to add the following check:

```
if description:
     payload['description'] = description
```

PR #1399, opened 2021-06-11
Fix release and version change for marquez_common python module

collado-mike:
Signed-off-by: Michael Collado <mike@datakin.com>

### DIFF
--- a/integrations/airflow/Dockerfile.tests
+++ b/integrations/airflow/Dockerfile.tests
@@ -2,10 +2,12 @@ FROM python:3.6-slim
 RUN mkdir /app
 ADD airflow/setup.py /app
 ADD airflow/README.md /app
-COPY common/marquez /app/marquez
+COPY common /tmp/marquez-integration-common
+RUN cd /tmp/marquez-integration-common && pip wheel --wheel-dir=/tmp/marquez-integration-common/wheel .
 COPY airflow/marquez_airflow /app/marquez_airflow
 COPY airflow/openlineage /app/openlineage
 WORKDIR /app
 RUN ls -la
-RUN pip wheel --wheel-dir=/app/wheel .[tests]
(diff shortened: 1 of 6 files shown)


mobuchowski:
Tried that approach and hit a wall here with pip20.3> dependency resolution - in that PR moved to having fake package and explicitely copying files to target packages:

https://github.com/MarquezProject/marquez/pull/1301/files

integrations/airflow/Dockerfile.tests
```diff
@@ -2,10 +2,12 @@ FROM python:3.6-slim
 RUN mkdir /app
 ADD airflow/setup.py /app
 ADD airflow/README.md /app
-COPY common/marquez /app/marquez
+COPY common /tmp/marquez-integration-common
+RUN cd /tmp/marquez-integration-common && ls . && pip wheel --wheel-dir=/tmp/marquez-integration-common/wheel .
```
wslulciuc, line 6: _Minor_: Do you still need run the `ls .` cmd?

integrations/airflow/setup.py
```diff
@@ -20,9 +20,10 @@
     readme = readme_file.read()
 
 requirements = [
-    "attrs>=19.3",
+    "attrs==19.3.0",
```
wslulciuc, line 23: I'm wondering if we want to fix this version?

wslulciuc (approved):
LGTM 👍

integrations/airflow/Dockerfile.tests
collado-mike: nope

integrations/airflow/setup.py
collado-mike: Sorry- that was done to verify the cause of a particular problem. I think we can get away without this

Issue #1363, opened 2021-05-29
Ensure airflow integration follow the open lineage naming convention for datasets


julienledem:
We need to go over the integrations (postgres, bigquery, snowflake, redshift, GE) and make sure that the Inputs and outputs follow this convention.
https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md
In particular, the namespace

mobuchowski:
https://github.com/MarquezProject/marquez/pull/1453

Issue #1397, opened 2021-06-11
"Ambiguous Segment in URI" when trying to retrieve namespaces with "s3://" in it


APMoctezuma:
Per OpenLineage [specifications](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#s3) a dataset coming from S3 should have a namespace in the form of `s3://<bucketname>`. This is working when it comes to publishing a run event  but when trying to retrieve a job or other information based on this namespace both endpoints of `/api/v1/namespaces/<namespace>` and `/api/v1-beta/lineage?nodeId=<dataset>:<namespace>:<name>` return a 400 error with the message "Ambiguous Segment in URI" when using the encoding of s3%3A%2F%2F for s3:// (since : and / are special characters I imagine we have to use encoding)

dominiquetipton:
Since I looked into this a bit already I figured I'd leave my findings here. The issue specifically comes from the `%2F` (the `/` character) in the url. It seems Dropwizard and/or Jetty isn't allowing it because it can lead to potential security issues. However, I didn't have any issues with `%2F` being in a query string though (tested with Marquez 0.14.2).

mobuchowski:
Seems like setting `httpCompliance: RFC2616_LEGACY` in marquez.yml is an workaround.

mobuchowski:
Fixed by https://github.com/MarquezProject/marquez/pull/1419

PR #1407, opened 2021-06-14
source names: conform to naming.md

mobuchowski:
Make changes necessary to conform to https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md

Refactor Source and Dataset classes to map more closely with this concept. 
Refactor Postgres and Snowflake connection handling, to cleanly extract database name.

After this PR, job namespace and dataset/source namespace are separate concepts. This does not necessarily map well to Marquez's concept of namespaces: job can be in namespace `food_delivery`, while it's datasets could be in `postgres://postgres:5432`. 

Getting proper data from GreatExpectations operator is not cleanly possible - integration requires providing this data now. 

Solves https://github.com/MarquezProject/marquez/issues/1363


mobuchowski:
#1397 is the reason for the integration test failure. It works locally on my laptop when applied `httpCompliance: RFC2616_LEGACY`

PR #1410, opened 2021-06-16
api, client: Allow slashes in URLs in both encoded and raw form

mobuchowski:
This allows valid OpenLineage namespaces and datasets like `s3://bucket` be created and retrieved using Marquez API and Marquez client. 

Fixes https://github.com/MarquezProject/marquez/issues/1397

### DIFF
--- a/clients/java/src/main/java/marquez/client/MarquezPathV1.java
+++ b/clients/java/src/main/java/marquez/client/MarquezPathV1.java
@@ -15,6 +15,7 @@ class MarquezPathV1 {
     /*
      Converts path template (prepended with BASE_PATH), where path parts are separated
      by slashes, and formats strings filling pathArgs in place of %s placeholder.
+     This is required to properly handle double slashes in URLs using URIBuilder.
      Example:
      Path template "/namespaces/%s/datasets/%s/versions/%s"
      with args
--- a/clients/java/src/main/java/marquez/client/MarquezPathV1.java
+++ b/clients/java/src/main/java/marquez/client/MarquezPathV1.java
@@ -5,72 +5,98 @@ import javax.annotation.Nullable;
 import lombok.NonNull;
 import marquez.client.models.RunState;
 
-class MarquezPathV1 {
+import java.util.ArrayList;
+import java.util.List;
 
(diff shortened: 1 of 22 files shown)


clients/java/src/main/java/marquez/client/MarquezPathV1.java
```diff
@@ -1,76 +1,102 @@
...
+     by slashes, and formats strings filling pathArgs in place of %s placeholder.
+     This is required to properly handle double slashes in URLs using URIBuilder.
+     Example:
+     Path template "/namespaces/%s/datasets/%s/versions/%s"
+     with args
+     "nName", "dName", "vName"
+     is converted to list
+     ["api", "v1", "namespaces", "nName", "datasets", "dName", "versions", "vName"]
+    */
+    List<String> resultPath = new ArrayList<>();
```
OleksandrDvornik, line 26: I didn't get why do we need that, but what about stream approach?
I have only one concern, what if number of %s and pathArgs mismatch, should we throw some exception, or put empty string instead of %s?
```
    String pathTemplate = "/namespaces/%s/datasets/%s/versions/%s";
    String[] pathArgs = new String[]{"nName", "dName", "vName"};
    Iterator<String> iterator = Arrays.stream(pathArgs).iterator();
    List<String> parts = Stream.of(pathTemplate.split("/"))
            .filter(StringUtils::isNotBlank)
            .map(it->it.equals("%s") ? iterator.next() : it)
            .collect(Collectors.toList());
    assertThat(parts).isEqualTo(Arrays.asList("namespaces", "nName", "datasets", "dName", "versions", "vName"));
```

marquez.dev.yml
```diff
@@ -2,6 +2,7 @@ server:
   applicationConnectors:
   - type: http
     port: 5000
+    httpCompliance: RFC2616_LEGACY
```
dominiquetipton, line 5: Would it be better to use `RFC7230_LEGACY` instead since `RFC7230` is the [default for Dropwizard](https://www.dropwizard.io/en/latest/manual/configuration.html#http)? With my basic testing this also allows forward slashes.

PR #1414, opened 2021-06-16
Updated bigquery and snowflake operators to avoid throwing import exceptions

collado-mike:
Snowflake and BigQuery extractors throw errors in environments that don't have those libraries installed. Updated them to wrap the imports in try/except blocks.

I also added a new `build.sh` script in the examples/airflow directory that allows spinning up the docker envs with the marquez libraries built locally. This enabled me to test that the snowflake/bigquery changes actually work in a docker image that has no snowflake or bigquery libraries installed.


wslulciuc (approved):
NICE! 💯 🥇

PR #1415, opened 2021-06-16
Add openAPI docs gen step to new-version.sh

wslulciuc:


new-version.sh
```diff
@@ -58,6 +59,12 @@ if [[ ! $(type -P bump2version) ]]; then
   exit 1;
 fi
 
+# Verify redoc-cli is installed
+if [[ ! $(type -P redoc-cli) ]]; then
+  echo "redoc-cli not installed! Please see https://www.npmjs.com/package/redoc-cli"
+  exit 1;
+fi
```
collado-mike, line 66: One thing we did at my last job to avoid requiring these dependencies on everyone's machine was to use docker images. E.g., https://hub.docker.com/r/redocly/redoc/

collado-mike (approved):

new-version.sh
wslulciuc: Yeah, thought about going this route. Let me make this change now 👍

wslulciuc: Hmm, it actually isn't as easy as I'd hoped. I update the command to use docker in a follow PR.

PR #1418, opened 2021-06-17
api: allow urlencoded path parameters, java client: properly urlencode path parameters

mobuchowski:
Use HttpCompliance.RFC7230_LEGACY to allow potentially ambiguous URIs with encoded path segments.
Refactor Marquez Java client to properly urlencode path segments.

### DIFF
--- a/clients/java/src/main/java/marquez/client/MarquezPathV1.java
+++ b/clients/java/src/main/java/marquez/client/MarquezPathV1.java
@@ -15,6 +15,7 @@ class MarquezPathV1 {
     /*
      Converts path template (prepended with BASE_PATH), where path parts are separated
      by slashes, and formats strings filling pathArgs in place of %s placeholder.
+     This is required to properly handle double slashes in URLs using URIBuilder.
      Example:
      Path template "/namespaces/%s/datasets/%s/versions/%s"
      with args
--- a/clients/java/src/main/java/marquez/client/MarquezPathV1.java
+++ b/clients/java/src/main/java/marquez/client/MarquezPathV1.java
@@ -27,16 +27,22 @@ class MarquezPathV1 {
      is converted to list
      ["api", "v1", "namespaces", "nName", "datasets", "dName", "versions", "vName"]
     */
-    int placeholderAmount = pathTemplate.split("%s").length;
+    // Find amount of placeholders
+    int placeholderAmount = 0;
+    for (int pos = pathTemplate.indexOf("%s"); pos >= 0; pos = pathTemplate.indexOf("%s", pos + 1)) {
+      placeholderAmount++;
+    }
+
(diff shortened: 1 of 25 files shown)


clients/java/src/main/java/marquez/client/MarquezPathV1.java
```diff
@@ -1,76 +1,113 @@
...
 
 class MarquezPathV1 {
-
   @VisibleForTesting static final String BASE_PATH = "/api/v1";
 
   @VisibleForTesting
-  static String path(String pathTemplate, @Nullable String... pathArgs) {
-    return BASE_PATH + String.format(pathTemplate, (Object[]) pathArgs);
+  static List<String> path(String pathTemplate, @Nullable String... pathArgs) {
+    /*
```
OleksandrDvornik, line 18: As I got MarquezPathV1 is for preparing path :)
I would prefer to put encoding here for pathArgs. Cause if it's not encoded path is not ready, then why do we need a class with separate responsibility, which do only half of the task

clients/java/src/main/java/marquez/client/MarquezPathV1.java
```diff
@@ -1,76 +1,113 @@
...
+     Example:
+     Path template "/namespaces/%s/datasets/%s/versions/%s"
+     with args
+     "nName", "dName", "vName"
+     is converted to list
+     ["api", "v1", "namespaces", "nName", "datasets", "dName", "versions", "vName"]
+    */
+    // Find amount of placeholders
+    int placeholderAmount = 0;
+    for (int pos = pathTemplate.indexOf("%s");
```
OleksandrDvornik, line 31: you can use 
`StringUtils.countMatches(pathTemplate, "%s");`

clients/java/src/main/java/marquez/client/MarquezUrl.java
```diff
@@ -45,14 +46,14 @@ static MarquezUrl create(URL url) {
...
+  URL from(List<String> path) {
     return from(path, ImmutableMap.of());
   }
 
   @VisibleForTesting
-  URL from(String path, @Nullable Map<String, Object> queryParams) {
+  URL from(List<String> path, @Nullable Map<String, Object> queryParams) {
     try {
-      final URIBuilder builder = new URIBuilder(baseUrl.toURI()).setPath(baseUrl.getPath() + path);
+      final URIBuilder builder = new URIBuilder(baseUrl.toURI()).setPathSegments(path);
```
OleksandrDvornik, line 56: we can with 
`final URIBuilder builder = new URIBuilder(URI.create(baseUrl.toURI()+path));`
if it would be encoded previously in MarquezPathV1

OleksandrDvornik (commented):
I would prefer to stick to the current contract and encode path param in MarquezPathV1 path method

PR #1419, opened 2021-06-17
api: allow urlencoded path parameters, java client: properly urlencode path parameters

mobuchowski:
Use HttpCompliance.RFC7230_LEGACY to allow potentially ambiguous URIs with encoded path segments.
Refactor Marquez Java client to properly urlencode path segments.

### DIFF
--- a/clients/java/src/main/java/marquez/client/MarquezPathV1.java
+++ b/clients/java/src/main/java/marquez/client/MarquezPathV1.java
@@ -1,17 +1,62 @@
 package marquez.client;
 
 import com.google.common.annotations.VisibleForTesting;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import lombok.NonNull;
 import marquez.client.models.RunState;
+import org.apache.maven.shared.utils.StringUtils;
 
 class MarquezPathV1 {
-
   @VisibleForTesting static final String BASE_PATH = "/api/v1";
 
   @VisibleForTesting
   static String path(String pathTemplate, @Nullable String... pathArgs) {
-    return BASE_PATH + String.format(pathTemplate, (Object[]) pathArgs);
+    /*
+     Converts path template (prepended with BASE_PATH), where path parts are separated
+     by slashes, and formats strings filling pathArgs in place of %s placeholder.
+     This is required to properly handle double slashes in URLs using URIBuilder.
+     Example:
+     Path template "/namespaces/%s/datasets/%s/versions/%s"
+     with args
+     "s3://bucket", "dName", "vName"
+     is converted to
+     "/api/v1/namespaces/s3%3A%2F%2Fbucket/datasets/nName/versions/vName"
+    */
+    int argsLength = pathArgs == null ? 0 : pathArgs.length;
+    if (StringUtils.countMatches(pathTemplate, "%s") != argsLength) {
+      throw new MarquezClientException(
+          String.format(
+              "Amount of placeholders %s differ from amount of provided path arguments %s",
+              pathTemplate.split("%s").length - 1, argsLength));
+    }
+
+    pathTemplate = BASE_PATH + pathTemplate;
+    if (pathArgs == null) {
+      return pathTemplate;
+    }
+
+    // Replace placeholders with path templates, removing empty strings
+    Iterator<String> iterator = Arrays.stream(pathArgs).iterator();
+    return Stream.of(pathTemplate.split("/"))
+        .filter(it -> it != null && !it.isEmpty())
+        .map(it -> it.equals("%s") ? iterator.next() : it)
+        .map(MarquezPathV1::encode)
+        .collect(Collectors.joining("/", "/", ""));
+  }
+
+  static String encode(String input) {
+    try {
+      return URLEncoder.encode(input, StandardCharsets.UTF_8.toString());
+    } catch (UnsupportedEncodingException e) {
+      throw new MarquezClientException(e);
+    }
   }
 
   static String listNamespacesPath() {
--- a/clients/java/src/main/java/marquez/client/MarquezWriteOnlyClientImpl.java
+++ b/clients/java/src/main/java/marquez/client/MarquezWriteOnlyClientImpl.java
@@ -33,9 +33,8 @@ class MarquezWriteOnlyClientImpl implements MarquezWriteOnlyClient {
     this.backend = backend;
   }
 
-  private static String path(String path, Map<String, Object> queryParams) {
-    StringBuilder pathBuilder = new StringBuilder();
-    pathBuilder.append(path);
+  private static String pathWithQueryParams(String path, Map<String, Object> queryParams) {
+    StringBuilder pathBuilder = new StringBuilder(path);
     if (queryParams != null && !queryParams.isEmpty()) {
       boolean first = true;
       for (Entry<String, Object> entry : queryParams.entrySet()) {
@@ -89,7 +88,7 @@ class MarquezWriteOnlyClientImpl implements MarquezWriteOnlyClient {
   public void markRunAs(String runId, RunState runState, Instant at) {
     Map<String, Object> queryParams =
         at == null ? null : ImmutableMap.of("at", ISO_INSTANT.format(at));
-    backend.post(path(runTransitionPath(runId, runState), queryParams));
+    backend.post(pathWithQueryParams(runTransitionPath(runId, runState), queryParams));
   }
 
   @Override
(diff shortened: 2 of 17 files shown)


clients/java/src/main/java/marquez/client/MarquezWriteOnlyClientImpl.java
```diff
@@ -89,7 +88,7 @@ public void createRun(String namespaceName, String jobName, RunMeta runMeta) {
   public void markRunAs(String runId, RunState runState, Instant at) {
     Map<String, Object> queryParams =
         at == null ? null : ImmutableMap.of("at", ISO_INSTANT.format(at));
-    backend.post(path(runTransitionPath(runId, runState), queryParams));
+    backend.post(queryPath(runTransitionPath(runId, runState), queryParams), null);
```
wslulciuc, line 91: Can we keep the method name as `path()`? or `pathWithQueryParams()`? Using `queryPath()` reads more like the query params for the path.

clients/java/src/test/java/marquez/client/MarquezWriteOnlyClientTest.java
```diff
@@ -98,26 +97,26 @@ public void testMarkRunAs() throws UnsupportedEncodingException {
...
+    verify(backend, times(1)).post("api/v1/jobs/runs/" + runId + "/start?at=" + atParam, null);
     client.markRunAsAborted(runId, at);
-    verify(backend, times(1)).post("/api/v1/jobs/runs/" + runId + "/abort?at=" + atParam);
+    verify(backend, times(1)).post("api/v1/jobs/runs/" + runId + "/abort?at=" + atParam, null);
     client.markRunAsCompleted(runId, at);
-    verify(backend, times(1)).post("/api/v1/jobs/runs/" + runId + "/complete?at=" + atParam);
+    verify(backend, times(1)).post("api/v1/jobs/runs/" + runId + "/complete?at=" + atParam, null);
     client.markRunAsFailed(runId, at);
-    verify(backend, times(1)).post("/api/v1/jobs/runs/" + runId + "/fail?at=" + atParam);
+    verify(backend, times(1)).post("api/v1/jobs/runs/" + runId + "/fail?at=" + atParam, null);
```
wslulciuc, line 106: I'm not sure I understand the usage of `null` being passed to `path()`. There is a method that already handle such case, see [`Backend.post(path)`](https://github.com/MarquezProject/marquez/blob/main/clients/java/src/main/java/marquez/client/Backend.java#L14)

clients/java/src/main/java/marquez/client/MarquezPathV1.java
```diff
@@ -1,17 +1,58 @@
...
+        .map(it -> it.equals("%s") ? iterator.next() : it)
+        .map(MarquezPathV1::encode)
+        .collect(Collectors.joining("/"));
+  }
+
+  static String encode(String input) {
+    try {
+      return URLEncoder.encode(input, StandardCharsets.UTF_8.toString());
+    } catch (UnsupportedEncodingException e) {
+      throw new RuntimeException(e);
```
wslulciuc, line 54: I would throw a `MarquezClientException` here as well.

clients/java/src/main/java/marquez/client/MarquezPathV1.java
```diff
@@ -1,17 +1,58 @@
...
+     "api/v1/namespaces/s3%3A%2F%2Fbucket/datasets/nName/versions/vName"
+    */
+    int argsLength = pathArgs == null ? 0 : pathArgs.length;
+    if (StringUtils.countMatches(pathTemplate, "%s") != argsLength) {
+      throw new MarquezClientException(
+          String.format(
+              "Amount of placeholders %s differ from amount of provided path arguments %s",
+              pathTemplate.split("%s").length - 1, argsLength));
+    }
+
```
OleksandrDvornik, line 39: we can add here 
`if(pathArgs == null ) return pathTemplate;`
and then we can use only iterator
`Iterator<String> iterator = pathArgs == null ? null : Arrays.stream(pathArgs).iterator();`
to
`Iterator<String> iterator = Arrays.stream(pathArgs).iterator();`
that would reduce cost of splitting

clients/java/src/main/java/marquez/client/Utils.java
```diff
@@ -59,7 +59,11 @@ public static String toJson(@NonNull final Object value) {
 
   public static URL toUrl(@NonNull final String urlString) {
     try {
-      return new URL(urlString);
+      String url = urlString;
+      if (!urlString.endsWith("/")) {
```
OleksandrDvornik, line 63: what about one liner?
`urlString.endsWith("/") ? urlString : urlString + "/"`

clients/java/src/test/java/marquez/client/MarquezPathV1Test.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+package marquez.client;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+@org.junit.jupiter.api.Tag("UnitTests")
+public class MarquezPathV1Test {
+
```
OleksandrDvornik, line 22: what about Parametrized test?

clients/java/src/test/java/marquez/client/MarquezWriteOnlyClientTest.java
```diff
@@ -89,7 +88,7 @@ public void testCreateRunWithId() {
     RunMeta runMeta = RunMeta.builder().id(id).build();
     client.createRun("namespaceName", "jobName", runMeta);
     verify(backend, times(1))
-        .post("/api/v1/namespaces/namespaceName/jobs/jobName/runs", runMeta.toJson());
```
OleksandrDvornik, line 92: what do you think about handling this case also in method?
`private URL url(String path) {`

clients/java/src/main/java/marquez/client/MarquezWriteOnlyClientImpl.java
mobuchowski: What about `attachQueryParams`? I think it's more accurate about what this function does.

clients/java/src/main/java/marquez/client/MarquezPathV1.java
mobuchowski: Done. It just won't ever happen :wink:  
The Java 1.0 classes have terrible ergonomics...

clients/java/src/test/java/marquez/client/MarquezWriteOnlyClientTest.java
mobuchowski: Any idea of something better than 
```java
      if (baseUrl.toString().endsWith("/") && path.startsWith("/")) {
        path = path.substring(1);
      } else if (!baseUrl.toString().endsWith("/") && !path.startsWith("/")) {
        path = "/" + path;
      }
```

clients/java/src/main/java/marquez/client/Utils.java
mobuchowski: Done

clients/java/src/test/java/marquez/client/MarquezPathV1Test.java
mobuchowski: Done

clients/java/src/test/java/marquez/client/MarquezWriteOnlyClientTest.java
mobuchowski: Done

clients/java/src/main/java/marquez/client/MarquezWriteOnlyClientImpl.java
mobuchowski: Done

mobuchowski:
@wslulciuc @OleksandrDvornik can we merge it?

OleksandrDvornik:
looks good for me

clients/java/src/main/java/marquez/client/HttpBackend.java
```diff
@@ -36,6 +36,11 @@ public URL getBaseUrl() {
 
   private URL url(String path) {
     try {
+      if (baseUrl.toString().endsWith("/") && path.startsWith("/")) {
```
wslulciuc, line 39: The caller passes in the `baseUrl`, so I agree stripping the `/` character is a valid check and should be removed. But, we control the `path` and also the base path (=`/api/v1`). So, I'm not sure changing the paths in the client makes sense given that they are internal and never exposed to the caller. Mind reverting the path changes in this PR, but still keep the stripping of the trailing `/` in the `baseUrl`?

wslulciuc (commented):
@mobuchowski, changes look great! I left a comment about reverting some changes related to the paths that are internal. Once reverted, we can merge!

clients/java/src/main/java/marquez/client/HttpBackend.java
wslulciuc: I'd also add the logic to strip the trailing `/` when a [new `MarquezUrl`](https://github.com/MarquezProject/marquez/blob/main/clients/java/src/main/java/marquez/client/MarquezUrl.java#L45) object is constructed rather than on each call to `put()`

mobuchowski: Done in Utils.toUrl() which is MarquezUrl's package-private API.

wslulciuc (approved):
LGTM 👍 💯 🥇

Issue #1361, opened 2021-05-27
Openlineage API does not set Dataset current_version_uuid


vitoravancini:
It seems that using the openlineage api the current version id is not populated for new datasets and the join that fetches the fields  for the dataset never works.

This is the join that fails, 'dv' is never actually found and dv.fields is always null
![image](https://user-images.githubusercontent.com/4209575/119858537-f0697580-beea-11eb-98f0-f650fced93a1.png)

The screenshot is for the DatasetDao.java file.
https://github.com/MarquezProject/marquez/blob/41d4073420b34a2217a05ceb52a65a94b95c22c2/api/src/main/java/marquez/db/DatasetDao.java#L81

Thank you!

wslulciuc:
Thanks for opening this issue, @vitoravancini! As of Marquez `0.15.+`, the _input_ datasets for an OpenLineage (=OL) event are no longer automatically register for a given run ID, see [`56f9...ef20`](https://github.com/MarquezProject/marquez/commit/56f97f91dbb7661346a24070dfdbaee1c5777a45#diff-ef203e3c4dd3b82cb08d8395b0c36868f4396e967c2bb172df629ec8fc8daaf3). This was to ensure output dataset versions were only created on run _completion_. You can view the job and its outputs as a single unit, where the output datasets provide a clear _public_ contract for downstream job consumption. The change was introduced in PR https://github.com/MarquezProject/marquez/pull/1258. But, we may want to reconsider the logic (or rather, refine it). Below, I outline the current flow when an OL event is received, then, a proposal for an alternative flow.

### How are input / output datasets for an OpenLineage event handled?

If we reference the OL events in the Marquez [quickstart](https://marquezproject.github.io/marquez/quickstart.html), the `my-input` dataset is not created in **Step 1**. In the Marquez backend, the run is initially created with the ID provided (as well as any run specific facets), but the input dataset is ignored and assumed to have been already registered by an upstream job (or directly using the [DatasetAPI](https://marquezproject.github.io/marquez/openapi.html#tag/Datasets/paths/~1namespaces~1{namespace}~1datasets~1{dataset}/put)). Then, in **step 2**, the OL event has `my-output` as an output dataset that is subsequently registered and versioned, then linked to the run ID. This results in the behavior you outlined in this issue.

### Should an input dataset be registered and versioned if present in an OL event?

As an alternative, we could register all output datasets present in an OL event. But, we'd have to consider:

1. How should input dataset ownership be assigned?
[...]

PR #1420, opened 2021-06-17
Fix open lineage api to update current version for created input datasets

collado-mike:
Quick and dirty fix for https://github.com/MarquezProject/marquez/issues/1361


wslulciuc (approved):
LGTM 👍

PR #1421, opened 2021-06-17
Fixing the display for job sql

phixMe:
This fixes the display for job SQL. It was being transcluded in our React component instead of included as a prop. We perhaps wish to show other keys here as well, but it's out of scope for this fix.
image


wslulciuc (approved):
❤️

PR #1424, opened 2021-06-18
Allow upsert of tags

### LINKED ISSUE #1412: Upserting tags doesn't update tag
The method `upsertTag` in [TagDao](https://github.com/MarquezProject/marquez/blob/009c865b141e71c2ab07f6d0a167a13e44890bea/api/src/main/java/marquez/db/TagDao.java) doesn't really update the tag description. The tag may be created but may never be changed afterwards.

hanbei:
Fixes #1412 that didn't allow to change the tag description after creation

Signed-off-by: Florian Schulz <florian.schulz@hanbei.de>


wslulciuc (approved):
❤️

PR #1451, opened 2021-06-21
Add a facet reporting unknown inputs outputs in marquez-spark

OleksandrDvornik:


integrations/spark/src/main/java/marquez/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -73,24 +77,43 @@ public void start(SparkListenerJobStart jobStart) {
...
             .outputs(outputDatasets)
             .run(
                 buildRun(
                     buildRunFacets(
-                        buildLogicalPlanFacet(queryExecution.optimizedPlan()),
-                        null,
-                        buildParentFacet())))
+                        buildParentFacet(),
+                        AdditionalFacetWrapper.compose(
+                            "spark.logicalPlan",
```
mobuchowski, line 114: I think we should move to follow [custom facet naming](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#custom-facet-naming)

OleksandrDvornik: so, we should go with 'SparkLogicalPlanRunFacet'? 
P.S. it was there already, I've just move it up

mobuchowski: And "spark_logicalPlan" key style, unless we already need to keep current names. @collado-mike do you know something about it?

mobuchowski:
We should add JSON Schema facet definition, and link to it in facet like https://github.com/MarquezProject/marquez/blob/main/integrations/spark/facets/spark-2.4/v1/logical-plan-facet.json and LogicalPlanFacet

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/SparkSQLExecutionContext.java
collado-mike: The existing spark facets were in place before we discussed any naming conventions. I think they ought to be migrated to match the expected conventions.

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/UnknownEntryFacetListener.java
```diff
@@ -0,0 +1,60 @@
...
+    List<FacetEntry> inputs =
+        seqAsJavaList(root.collectLeaves()).stream()
+            .map(this::mapEntry)
+            .filter(Objects::nonNull)
+            .collect(Collectors.toList());
+    return output == null && inputs.isEmpty() ? null : new UnknownEntryFacet(output, inputs);
+  }
+
+  private FacetEntry mapEntry(LogicalPlan x) {
+    if (processedNodes.containsKey(x)) return null;
```
collado-mike, line 42: I'm not sure I understand when `mapEntry` will ever get past the first line. If I read the `PlanTraversal` code correctly, the `accpet` method here gets called for every visited node when we call `collect` on the LogicalPlan. In what circumstances will we not visit a node that gets returned by the `collectLeaves` method call here in `build`?

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/UnknownEntryFacetListener.java
```diff
@@ -0,0 +1,60 @@
...
+    List<UnknownEntryFacet.AttributeField> output = attributeFields(x.outputSet());
+    List<UnknownEntryFacet.AttributeField> input = attributeFields(x.inputSet());
+    return new FacetEntry(x.toJSON(), input, output);
```
collado-mike, line 45: The logical plan visitor already gets the output of this `toJSON` call. I think what we'll need is something beyond the built-in `toJSON` converter- maybe something that includes that information, but also adds the contents of private fields? Especially if those private fields are themselves LogicalPlan implementations (e.g., a subquery contains a reference to the LogicalPlan of the query).

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/UnknownEntryFacetListener.java
OleksandrDvornik: we will not visit a node if we don't have partial function which is going to map it.
At every `PartialFunction` we have method `isDefinedAt`, which is saying that this function can map a Node to Dataset.
We should skip such Nodes, cause it's known for us.
In `accept` method we collect all "known" nodes, and then in `mapEntry` we have a check/filter to skip already visited nodes.
```
private FacetEntry mapEntry(LogicalPlan x) {
    if (processedNodes.containsKey(x)) return null;
```

integrations/spark/src/main/java/marquez/spark/agent/lifecycle/plan/UnknownEntryFacetListener.java
OleksandrDvornik: it's good idea, will take a look on that. We want to have similar JSON format. right?

OleksandrDvornik: @collado-mike  maybe you have some example, how can I test that?

collado-mike (commented):
Any luck with exploring the `json4s` library to extract extra fields from the plan nodes?

OleksandrDvornik:
having some issue with matching, trying to add custom format to it.
`scala.MatchError: MethodType(`

julienledem:
This now moves to https://github.com/OpenLineage/OpenLineage/tree/main/integration

julienledem:
closing as it will be reopened in OpenLineage

PR #1453, opened 2021-06-21
source names: conform to naming.md

mobuchowski:
Make changes necessary to conform to https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md

Refactor Source and Dataset classes to map more closely with this concept.
Refactor Postgres and Snowflake connection handling, to cleanly extract database name.

After this PR, job namespace and dataset/source namespace are separate concepts. This does not necessarily map well to Marquez's concept of namespaces: job can be in namespace food_delivery, while it's datasets could be in postgres://postgres:5432.

Getting proper data from GreatExpectations operator is not cleanly possible - integration requires providing this data now.

Solves #1363

Signed-off-by: Maciej Obuchowski maciej.obuchowski@getindata.com


integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -79,6 +79,11 @@ def _get_schema() -> str:
         return "https://github.com/MarquezProject/marquez/tree/main/integrations/airflow/marquez_airflow/extractors/ge-assertions-dataset-facet.json"  # noqa
 
 
+def provide_dataset_info(operator, namespace: str, name: str):
```
wslulciuc, line 82: Mind if we add docs on the importance of setting the `_expectations*` mappings? And I'd call the method `set_dataset_info()` given that it will be called to mutate great expectations values for the extractor.

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -79,6 +79,11 @@ def _get_schema() -> str:
...
+    operator._expectations_namespace = namespace
+    operator._expectations_name = name
```
wslulciuc, line 84: _Minor:_ Should we be explicate an use `_great_expectations_*`?

integrations/airflow/tests/extractors/test_bigquery_extractor.py
```diff
@@ -97,8 +97,10 @@ def test_extract(self, mock_client, mock_hook):
             'bigquery-public-data.usa_names.usa_1910_2013'
 
         assert step_meta.inputs[0].fields is not None
+        assert step_meta.inputs[0].source.name == \
+               'bigquery:'
```
wslulciuc, line 101: Is `bigquery:` a valid connection string for BigQuery?

integrations/airflow/tests/extractors/test_great_expectations_extractor.py
```diff
@@ -119,3 +122,53 @@ def test_great_expectations_operator_batch_kwargs_success():
...
+        fail_task_on_validation_failure=False
+    )
+
+    provide_dataset_info(operator, namespace, 'schema.table')
+
+    extractor = GreatExpectationsExtractor(operator)
+    result = operator.execute({})
+    step = extractor.extract_on_complete(None)
+
+    log.info(result)
```
wslulciuc, line 150: Can we remove this log from the test?

integrations/airflow/tests/integration/integration.py
```diff
@@ -29,7 +25,7 @@
 
 log = logging.getLogger(__name__)
 
-NAMESPACE_NAME = 'food_delivery'
+JOB_NAMESPACE_NAME = 'food_delivery'
```
wslulciuc, line 28: _Minor:_ You can still keep `NAMESPACE_NAME` as it's apparent in the tests that the source name is used for the namespace in some cases. Also, reduces the diff in the PR.

integrations/airflow/tests/test_marquez_dag.py
```diff
@@ -155,7 +155,7 @@ def test_marquez_dag(job_id_mapping, mock_get_or_create_openlineage_client,
                 "parentRun": ParentRunFacet.create(
                     runId=DAG_RUN_ID,
                     namespace=DAG_NAMESPACE,
-                    job_name=f"{DAG_ID}.{TASK_ID_COMPLETED}"
+                    name=f"{DAG_ID}.{TASK_ID_COMPLETED}"
```
wslulciuc, line 158: Can we keep `job_name`? It's not obvious what `name` is referring to exactly. That is: is it the run? is it the job?

wslulciuc:
> After this PR, job namespace and dataset/source namespace are separate concepts. This does not necessarily map well to Marquez's concept of namespaces: job can be in namespace food_delivery, while it's datasets could be in postgres://postgres:5432.

While this is true, a namespace is a way of group datasets and associating ownership. So, a data source is certainly valid. Meaning, a `MySql` source can have one or more tables under it's namespace.

integrations/airflow/tests/test_marquez_dag.py
mobuchowski: Need to change that here then first: https://github.com/OpenLineage/OpenLineage/blob/main/client/python/openlineage/facet.py#L55

Before this PR, airflow integration was using it's own buildin OpenLineage client - which was there before client was officially in OL repo. Now, it's using OL client, reducing code duplication, and possibility of diverging changes like this one :slightly_smiling_face:

integrations/airflow/tests/extractors/test_bigquery_extractor.py
mobuchowski: Sources conflict with that. For example, [rest API docs](https://cloud.google.com/bigquery/docs/reference/rest) use https://bigquery.googleapis.com. On the other hand, "official 3rd party" JDBC driver uses `https://www.googleapis.com/bigquery/v2`. On the third hand, official client libraries do not use concept of connection string, since there is one, global bigquery instance, and only thing that needs to be sorted is authentication.

integrations/airflow/tests/integration/integration.py
mobuchowski: Done.

integrations/airflow/tests/extractors/test_great_expectations_extractor.py
mobuchowski: Done.

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
mobuchowski: I've provided info in integration's readme, and now added docstring with example to the function.

.circleci/config.yml
```diff
@@ -195,6 +195,7 @@ jobs:
     steps:
       - *checkout_project_root
       - run: ./../.circleci/get-docker-compose.sh
+      - run: bash -c '(cd ~/marquez && docker build -t marquezproject/marquez . )'
```
wslulciuc, line 198: Ahh, this allows us to use the Marquez docker image we build from source. Nice!

integrations/airflow/README.md
```diff
@@ -42,40 +42,27 @@ $ python3 setup.py install
...
-MARQUEZ_BACKEND=FILE
-```
+`marquez-airflow` uses OpenLineage client to push data to Marquez.
 
-> **Note:** Metadata will be written to `/tmp/marquez/client.requests.log`, but the location can be overridden with `MARQUEZ_FILE`.
+OpenLineage client depends on environment variables:
 
-### `HTTP` Backend Authentication
+* `OPENLINEAGE_URL` - point to service which will consume OpenLineage events
+* `OPENLINEAGE_PRODUCER` - name of producer that client will send along with your events
```
wslulciuc, line 53: We'll be dropping support for `OPENLINEAGE_PRODUCER`, see https://github.com/OpenLineage/OpenLineage/issues/72. Mainly, it's to ensure callers configure the producer via the `py` client constructor. Though, `OPENLINEAGE_URL` and `OPENLINEAGE_API_KEY` make sense to keep as env vars, but also have support to set via the client constructor.

integrations/airflow/README.md
```diff
@@ -93,6 +80,11 @@ If you do nothing, Marquez will receive the `Job` and the `Run` from your DAGs,
 
 It's important to understand the inputs and outputs are lists and relate directly to the `Dataset` object in Marquez.  Datasets also include a source which relates directly to the `Source` object in Marquez.
 
+#### Great Expectations
+
+`great_expectations` extractor requires more care than that. For technical reasons, you need to manually provide dataset
+name and namespace for dataset provided to great expectations operator by calling function `marquez_airflow.extractors.great_expectations_extractor.provide_dataset_info`.
```
wslulciuc, line 86: _Minor:_ I think you'l need to update the function call to use `set_dataset_info()`?

integrations/airflow/marquez_airflow/extractors/great_expectations_extractor.py
```diff
@@ -79,6 +79,30 @@ def _get_schema() -> str:
         return "https://github.com/MarquezProject/marquez/tree/main/integrations/airflow/marquez_airflow/extractors/ge-assertions-dataset-facet.json"  # noqa
 
 
+def set_dataset_info(operator: GreatExpectationsOperator, namespace: str, name: str):
```
wslulciuc, line 82: 💯

integrations/airflow/tests/extractors/test_bigquery_extractor.py
wslulciuc: Thanks for the explanation, @mobuchowski!

integrations/airflow/tests/test_marquez_dag.py
wslulciuc: Ahh, I see. I'll give the `py` client a review before the OL `0.1.0` release 👍

wslulciuc (approved):
Left some minor comments, otherwise great work @mobuchowski 💯 💯 💯

integrations/airflow/README.md
mobuchowski: Added notification. Although, `marquez_airflow` users don't configure connection by client themselves now. This would require changes of how users use Airflow integration - aka it won't be just drop-in `from marquez_airflow import DAG`.

On the other hand, this is good - we have to change it anyway to support 2.0 and stuff like [taskflow API](https://airflow.apache.org/docs/apache-airflow/stable/tutorial_taskflow_api.html)

Issue #1416, opened 2021-06-17
Promote the lineage API out of beta


julienledem:
currently the lineage api is on /api/v1-beta/lineage
See:
https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/LineageResource.java
It is time to promote it to /api/v1/lineage

Possibly adding GET methods here.
https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/OpenLineageResource.java

The old path should work for the duration of the deprecation cycle

wslulciuc:
Fixed #1457

PR #1457, opened 2021-06-24
Promote the lineage API out of beta. Issue ref#:1416

OleksandrDvornik:
Issue: https://github.com/MarquezProject/marquez/issues/1416
Not sure about one thing, should I move it to OpenLineageResource, or should it stay in the same LineageResource.

Notice one thing in Node model. Please check [type](https://github.com/MarquezProject/marquez/compare/main...OleksandrDvornik:promote_lineage_api?expand=1#diff-6c47b767832d73df61783f7df93e55a44459a08d2711517be993f00e5bee2959R10) in json we have two fields with same name:
"type": "DATASET",
"type": "DB_TABLE",

One is for subclassing, second is real data


OleksandrDvornik:
Update on double "type" field.
If we have only one type field:
1."type": "DATASET" -> getting exception 
`com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `marquez.db.models.DatasetData`, problem: type is marked non-null but is null`
1."type": "DB_TABLE" -> getting exception 
`com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Could not resolve type id 'DB_TABLE' as a subtype of `marquez.db.models.NodeData`: known type ids = [DATASET, JOB] (for POJO property 'data')`

So, two fields with same name is mandatory for now.
Good news - it's valid, and parser can handle that

_RFC 7159 talks about JSON documents, not parsers. It recommends (but not requires) the JSON documents to have unique object keys. This means documents with non-unique keys are still valid._

api/src/main/java/marquez/api/LineageResource.java
```diff
@@ -32,7 +32,7 @@ public LineageResource(@NonNull final ServiceFactory serviceFactory) {
   @ResponseMetered
   @ExceptionMetered
   @GET
-  @Path("/lineage")
+  @Path("{version:v1-beta|v1}/lineage")
   @Consumes(APPLICATION_JSON)
   @Produces(APPLICATION_JSON)
   public Response getLineage(
```
collado-mike, line 38: I think we discussed moving this whole method to the `OpenLineageResource`. I think it's fine to update the `@Path` annotations to support the `v1-beta` path as well as the `v1` path for now. When we remove the deprecated path, we can consolidate the `@Path` annotations again.

OleksandrDvornik: ok, sure

wslulciuc: Once you move the method `LineageResource.getLineage()` to `OpenLineageResource`, you can delete the class `LineageResource`.

wslulciuc:
> So, two fields with same name is mandatory for now.
Good news - it's valid, and parser can handle that

Thanks for letting us know about the `type` field being duplicated, @OleksandrDvornik! I've opened an issue https://github.com/MarquezProject/marquez/issues/1471

api/src/test/java/marquez/api/OpenLineageResourceTest.java
```diff
@@ -0,0 +1,67 @@
...
+        UNDER_TEST
+            .target("/api/v1-beta/lineage")
+            .queryParam("nodeId", "job:test")
+            .request()
+            .get()
+            .readEntity(Lineage.class);
+    Lineage response =
+        UNDER_TEST
+            .target("/api/v1/lineage")
+            .queryParam("nodeId", "job:test")
```
wslulciuc, line 60: Do we know what the expect response would be? A node ID can either be `dataset:<namespace>:<dataset>` or `job:<namespace>:<job>`. So, I _think_ the node ID `job:test` used in your test should throw a `500`.

api/src/main/java/marquez/api/LineageResource.java
```diff
@@ -19,8 +19,13 @@
 import marquez.service.ServiceFactory;
 import marquez.service.models.NodeId;
 
+/**
+ * @deprecated This resource is no longer supported
+ *     <p>Use {@link OpenLineageResource#getLineage(NodeId, int)} instead.
+ */
 @Path("/api/v1-beta")
 @Slf4j
+@Deprecated(since = "0.16")
```
wslulciuc, line 28: _Minor:_ I would set _since_ to `0.16.0` to be specific.

wslulciuc (approved):
A couple minor comments, otherwise 👍

api/src/test/java/marquez/api/OpenLineageResourceTest.java
collado-mike: The service mock on line 35 will return the contents of `node.json` no matter what its inputs are

collado-mike (approved):

api/src/test/java/marquez/api/OpenLineageResourceTest.java
wslulciuc: 👍

api/src/main/java/marquez/api/LineageResource.java
OleksandrDvornik: ok, will do it in next PR

api/src/test/java/marquez/api/OpenLineageResourceTest.java
OleksandrDvornik: so, value  "job:test" passes pattern matching in NodeId
  ```
private static final Pattern ID_PATTERN =
      Pattern.compile(
          String.format("^(%s|%s|%s):.*$", ID_PREFX_DATASET, ID_PREFX_JOB, ID_PREFX_RUN));
```
I guess we should rethink that pattern, cause we will fail later on.

```
@JsonIgnore
  public JobVersionId asJobVersionId() {
    String[] parts = parts(4, ID_PREFX_JOB);
    return new JobVersionId(
        NamespaceName.of(parts[1]), JobName.of(parts[2]), UUID.fromString(parts[3]));
  }

  @JsonIgnore
  public DatasetVersionId asDatasetVersionId() {
    String[] parts = parts(4, ID_PREFX_DATASET);
    return new DatasetVersionId(
        NamespaceName.of(parts[1]), DatasetName.of(parts[2]), UUID.fromString(parts[3]));
  }
```

PR #1474, opened 2021-06-30
Reverted delete of BigQueryNodeVisitor to work with vanilla SparkListener

collado-mike:
After https://github.com/MarquezProject/marquez/pull/1389 was merged, we can add back the native BigQueryNodeVisitor which we deleted because the javaagent jar was not loaded by the same ClassLoader as all other Spark dependencies. By referencing the marquez package in `spark.jars.packages`, the jar will be loaded in the same user classpath as all other dependencies, so we can the BigQuery classes directly, without modifying the bytecode.

We can keep both implementations simultaneously, as anyone using the javagent approach will not be able to use the BigQueryNodeVisitor, because of the classpath problem, and anyone using the `spark.jars.packages` configuration will not be able to use the bytecode instrumentation, as there is no bytecode transformation step.


integrations/spark/build.gradle
```diff
@@ -40,6 +40,8 @@ dependencies {
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}"
     compileOnly "org.apache.spark:spark-core_2.11:${sparkVersion}"
     compileOnly "org.apache.spark:spark-sql_2.11:${sparkVersion}"
+    compileOnly 'com.google.cloud.spark:spark-bigquery_2.11:0.19.1'
+    testImplementation 'com.google.cloud.spark:spark-bigquery-with-dependencies_2.11:0.19.1'
```
wslulciuc, line 44: The `spark-bigquery-with-dependencies` is already defined in the test dependencies below.

wslulciuc (approved):
LGTM 👍

integrations/spark/build.gradle
collado-mike: Thanks. Fixed

PR #1475, opened 2021-06-30
fix: bigquery URI should lose colon at the end

mobuchowski:
Java URI class does not accept URI with colon at the end,
Then, method getUrlOrPlaceholder changes it to http://bigquery: as a result.


integrations/airflow/tests/extractors/test_great_expectations_extractor.py
```diff
@@ -126,7 +126,7 @@ def test_great_expectations_operator_batch_kwargs_success():
 
 @pytest.mark.parametrize('namespace,scheme,authority', [
     ('postgres://localhost:5432', 'postgres', 'localhost:5432'),
-    ('bigquery:', 'bigquery', ''),
+    ('bigquery', 'bigquery', None),
```
wslulciuc, line 129: Do we need to test `bigquery` twice here?

wslulciuc (approved):
LGTM 👍

PR #1476, opened 2021-06-30
Fix packaging for dbt `py` modules

wslulciuc:
This PR fixes the way the `py` modules for DBT are versioned to align more with the way existing `py` modules are released / versioned.

Thanks @collado-mike for the suggestion on the fix :)


integrations/dbt/bigquery/setup.cfg
```diff
@@ -11,14 +11,14 @@ serialize =
 search = version="{current_version}"
 replace = version="{new_version}"
 
-[bumpversion:file:./setup.py]
-search = marquez-integration-common=={current_version}
```
collado-mike, line 15: Does this module not require the `marquez-integration-common` package?

wslulciuc: No, it was copied over by mistake. Here are the dependencies in [`setup.py`](https://github.com/MarquezProject/marquez/blob/fix/versioning-for-dbt-integrations/integrations/dbt/bigquery/setup.py):

```
requirements = [
    "dbt-core>=0.20.0b1",
    "dbt-bigquery>=0.20.0b1",
    "sqlparse>=0.3.1",
    "openlineage-python==0.0.1rc6"
]
```

collado-mike (approved):

integrations/dbt/bigquery/setup.cfg
collado-mike: I see. I guess the "reusable" bigquery utilities aren't that reusable, then?

wslulciuc: That was my initial thinking as well. But, with the DBT integration being experimental, and using _adopters_ it most likely differs greatly from our airflow integration. Perhaps @mobuchowski could provide more insight?

Issue #1360, opened 2021-05-27
Marquez 0.15.0 Facets in Output Dataset Missing


kachontep:
I have attempted to upgrade from Marquez 0.14.2 to 0.15.0. However I see that there is a missing facets in my output dataset from  my old previous code which  work properly in version 0.14.2.

I use VS code extension [REST-Client](https://marketplace.visualstudio.com/items?itemName=humao.rest-client) to reproduce the steps and list the steps as below:

```
@baseUrl = http://localhost:5000/api/v1

###
# Set up test namespace
PUT {{baseUrl}}/namespaces/test
Content-Type: application/json

{
    "ownerName": "Test",
    "description": "Test namespace"
}

###
# Set up test datasource
PUT {{baseUrl}}/sources/test
Content-Type: application/json

{
    "type": "POSTGRES",
    "connectionUrl": "jdbc:postgresql://db.example.com/testdb",
    "description": "Tes source"
}

###
# Set up test dataset
PUT {{baseUrl}}/namespaces/test/datasets/test
Content-Type: application/json

{
    "type": "DB_TABLE",
    "physicalName": "testtbl",
    "sourceName": "test",
    "description": "Test dataset"
}

###
# Set up test job
PUT {{baseUrl}}/namespaces/test/jobs/test
Content-Type: application/json

{
    "type": "BATCH",
    "inputs": [],
    "outputs": [{
        "namespace": "test",
        "name": "test"
    }],
    "location": "https://github.com/test/testjob",
    "context": {
        "SQL": "SELECT * FROM EVERYTHING"
    },
    "description": "Test job"
}

###
# Create a job run
# @name createJobRun
POST {{baseUrl}}/namespaces/test/jobs/test/runs
Content-Type: application/json

{
    "id": "{{$guid}}"
}

###
# Start a job run
POST {{baseUrl}}/jobs/runs/{{createJobRun.response.body.id}}/start

###
# Send facet to OpenLineage API with START 
POST {{baseUrl}}/lineage
Content-Type: application/json

{
    "eventType": "START",
    "eventTime": "{{$datetime iso8601}}",
    "run": {
        "runId": "{{createJobRun.response.body.id}}",
        "facets": {
            "startEventFacet": "{{$datetime iso8601}}"
        }
    },
    "job": {
        "namespace": "test",
        "name": "test"
    },
    "inputs": [],
    "outputs": [{
        "namespace": "test",
        "name": "test",
        "facets": {
[...]

PR #1477, opened 2021-06-30
Fixed DatasetDao queries to return input and output facets

collado-mike:
I think this addresses https://github.com/MarquezProject/marquez/issues/1360 . Tested by spinning up the local docker image with seed data and ran the following bash
```

RUN_ID=60b858cc-faaf-498a-a242-e774f98a937a
curl -X POST http://localhost:5000/api/v1/namespaces/food_delivery/jobs/example.etl_delivery_7_days/runs -H 'Content-Type: application/json' -d @- <<EOL
{
    "id": "$RUN_ID"
}
EOL

###
# Start a job run
curl -X POST http://localhost:5000/api/v1/jobs/runs/$RUN_ID/start

###
# Send facet to OpenLineage API with START 
curl -X POST http://localhost:5000/api/v1/lineage -H 'Content-Type: application/json' -d @- <<EOL

{
    "eventType": "START",
    "eventTime": "2021-06-30T00:00:00Z",
    "run": {
        "runId": "$RUN_ID",
        "facets": {
            "startEventFacet": "2021-06-30T00:00:00Z"
        }
    },
    "job": {
        "namespace": "food_delivery",
        "name": "example.etl_delivery_7_days"
    },
    "inputs": [],
    "outputs": [{
        "namespace": "food_delivery",
        "name": "public.delivery_7_days",
        "facets": {
            "startEventFacet": "2021-06-30T00:00:00Z"
        }
    }],
    "producer": "https://github.com/kachontep/marquez-0.15.0/openlineage-facets-reproduce"
}
EOL


###
# Complete a job run
curl -X POST http://localhost:5000/api/v1/jobs/runs/$RUN_ID/complete

curl -X POST http://localhost:5000/api/v1/lineage -H 'Content-Type: application/json' -d @- <<EOL
{
    "eventType": "COMPLETE",
    "eventTime": "2021-06-30T00:00:00Z",
    "run": {
        "runId": "$RUN_ID",
        "facets": {
            "completeEventFacet": "2021-06-30T00:00:00Z"
        }
    },
    "job": {
        "namespace": "food_delivery",
        "name": "example.etl_delivery_7_days"
    },
    "inputs": [],
    "outputs": [{
        "namespace": "food_delivery",
        "name": "public.delivery_7_days",
        "facets": {
            "completeEventFacet": "2021-06-30T00:00:00Z"
        }
    }],
    "producer": "https://github.com/kachontep/marquez-0.15.0/openlineage-facets-reproduce"
}
EOL
```
Then read the dataset definition with
```
curl -s "http://localhost:5000/api/v1/namespaces/food_delivery/datasets/public.delivery_7_days" | jq | less
```

Initially, the `facets` field was empty, as reported in the issue. After the SQL change, both expected facets are present.

I also took the opportunity to test the query against a large dataset and updated the query to run faster. I also added/updated unit tests that should have caught this issue. Validated unit tests failed before the fix and pass after.


api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -130,32 +147,51 @@ default void setFields(Dataset ds) {
   Optional<DatasetRow> getUuid(String namespaceName, String datasetName);
 
   @SqlQuery(
-      "SELECT d.*, dv.fields, sv.schema_location, facets, t.tags\n"
-          + "FROM datasets d\n"
-          + "INNER JOIN dataset_versions dv on d.current_version_uuid = dv.uuid\n"
-          + "LEFT OUTER JOIN stream_versions AS sv ON sv.dataset_version_uuid = dv.uuid\n"
-          + "LEFT OUTER JOIN (\n"
+      "WITH selected_datasets AS (\n"
```
wslulciuc, line 150: _Minor_: I _think_ it'd be possible to factor out some of the common SQL into a constant for reuse?

wslulciuc (approved):
@collado-mike, thanks for fixing this bug and for adding a test suite for `DatasetDao`! 💯

api/src/main/java/marquez/db/DatasetDao.java
collado-mike: yeah, we need to start incorporating some strategy to reuse SQL- I felt this badly when I was implementing the jobVersion/datasetVersions in the Run model.

Here, the `selected_datasets` differs between the two queries- one specifies namespace and name, the other just specifies namespace and applies limit/offset. The rest of the SQL is the same until the end- the `findAll` query applies an `ORDER BY` clause at the end. It seems silly to me to extract the middle out and concatenate, so my preference is to leave it as is until we either start organizing reusable sql or refactoring tables so we don't need quite so much.

wslulciuc: Makes sense! I'll open an issue to begin evaluating some of the current SQL and a plan to support reusability. But, for now, this is great 👍

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -130,32 +147,51 @@ default void setFields(Dataset ds) {
...
+          + "    SELECT d.*\n"
+          + "    FROM datasets d\n"
+          + "    WHERE d.namespace_name=:namespaceName\n"
+          + "    ORDER BY d.name\n"
+          + "    LIMIT 10 OFFSET 0\n"
```
dominiquetipton, line 155: @collado-mike I know this is already merged, but is this supposed to be `LIMIT :limit OFFSET :offset` instead?

collado-mike: 🤦🏽‍♂️

That's what I get for going back to my editor for one more test.

dominiquetipton: We've all done it haha 😅  Speaking of tests, in mine the ORDER BY, LIMIT, AND OFFSET had to be at the end of the query in order for them to behave as expected (although I'm sure you noticed that too).

PR #1479, opened 2021-07-01
Set content type for dbt snowflake description to 'text/markdown'

wslulciuc:


phixMe (approved):
Looks good @wslulciuc Thanks

PR #1481, opened 2021-07-02
fix: add common dependency to dbt plugins

mobuchowski:


wslulciuc (changes requested):
Mind also adding the follow regex in each `setup.cfg` file to bump the common lib?

```
[bumpversion:file:./setup.py]
search = marquez-integration-common=={current_version}
replace = marquez-integration-common=={new_version}
```

For an example, see https://github.com/MarquezProject/marquez/blob/main/integrations/airflow/setup.cfg#L14

mobuchowski:
Fixed.

integrations/dbt/bigquery/setup.cfg
```diff
@@ -3,13 +3,13 @@ current_version = 0.16.0
...
-serialize = 
+serialize =
 	{major}.{minor}.{patch}{rc}
 	{major}.{minor}.{patch}
 
 [bumpversion:file:setup.py]
-search = version="{current_version}"
-replace = version="{new_version}"
+search = __version__ = "{current_version}"
+replace = __version__ = "{new_version}"
```
wslulciuc, line 12: Oh, nice! I think we can clean this up even more and update the other `py` modules to follow the convention. For now, this looks great 👍

wslulciuc (approved):

PR #1482, opened 2021-07-02
dbt packages should look for namespace packages

mobuchowski:
Building package without this results with **actual code not being packaged** because it uses PEP420 namespace packages. Of course, local install just adds this to the path, so it does not cause any problems in tests.


wslulciuc (approved):
LGTM 👍

PR #1493, opened 2021-07-09
Fix Spark openlineage.* configuration parameters and add integration test container execution

collado-mike:
Found out that we need to prefix Spark configuration parameters with `spark.` when supplying them from the CLI. Otherwise, they just get ignored. Made the fix, but needed a way to test it, so I added an integration test step that uses local containers to spin up marquez and a Spark shell to run arbitrary Spark scripts and validate the appropriate lineage is reported. This integration test step is now executed in the CircleCI build as a separate test step (I didn't want to run it in `gradle build` because it takes a long time).

I added two tests- a simple Spark SQL test and one with mixed RDD and SparkSQL. I found some gaps in the coverage for the RDD when using the plain SparkListener (without the javaagent bytecode instrumentation), so I covered that gap and now all tests are passing. I also made a minor improvement to the job naming for RDD jobs, so it's more similar to the Spark SQL approach.

I wrote the integ test to support multiple Spark versions, so a future PR can add the steps to test with multiple versions of Spark/Scala.

It's more than I'd normally like in a single PR, but this is how yak shaving goes...
image


.circleci/config.yml
```diff
@@ -187,6 +188,53 @@ jobs:
...
+      JDK8_HOME: /usr/lib/jvm/java-8-openjdk-amd64
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-integration-spark-{{ .Branch }}-{{ .Revision }}
+            - v1-integration-spark-{{ .Branch }}
+      - attach_workspace:
+          at: .
+      - run: docker load -i ./build/marquezproject-api-image.tar
```
wslulciuc, line 223: Just learned that you can load an image from a `.tar` file. Very cool!

wslulciuc (approved):
LGTM ❤️

PR #1494, opened 2021-07-09
Fix multi-statement sql parsing

### LINKED ISSUE #1492: The Postgres adapter in the airflow-marquez integration should support multi-statements SQL
https://github.com/MarquezProject/marquez/blob/89415e7bcc0dfdbb6c9051f9ac7315cb6e0ff988/integrations/common/marquez/sql/parser.py#L129
```
        # We assume only one statement in SQL
        tokens = TokenList(statements[0].tokens)
```
This logic should instead iterate over the statements and return the union of all inputs and outputs.

see doc of the parser library here: https://sqlparse.readthedocs.io/en/latest/

Here is an example statement for unit tests:
```
DROP TABLE IF EXISTS schema.table3;
CREATE TABLE schema.table3 (
  a VARCHAR(64),
  b VARCHAR(64)
)
DISTKEY(a)
INTERLEAVED SORTKEY(
  a,
  b
);
INSERT INTO schema.table3(
  a,
  b
)
SELECT
   a,
   b
from schema.table2
;
```

wslulciuc:
This PR fixes #1492 allowing for the parsing of multi-statement sql.


mobuchowski (approved):
LGTM, just fix the flake8 test :slightly_smiling_face:

PR #1495, opened 2021-07-10
Fixes for serializing and deserialing Run and JobData models

collado-mike:
I realized that Runs can't be deserialized from JSON once they've been serialized, due to the handling of the new jobVersion field. This addresses a fix in that model as well as the JobData model, which disallowed `null` values for several fields, even though they wouldn't be present at deserialization time.


api/src/main/java/marquez/service/models/Run.java
```diff
@@ -16,29 +16,41 @@
...
 import marquez.common.models.DatasetVersionId;
 import marquez.common.models.JobName;
 import marquez.common.models.JobVersionId;
 import marquez.common.models.NamespaceName;
 import marquez.common.models.RunId;
 import marquez.common.models.RunState;
 
+/**
+ * Service model for a job run. Job attributes are serialized as {@link JobVersion} attributes. The
+ * {@link #location} field is ignored entirely in the serialized output. Deserializing from JSON
```
wslulciuc, line 45: Why is the location missing when serialized?

wslulciuc (approved):
LGTM, just had a minor clarifying question.

api/src/main/java/marquez/service/models/Run.java
collado-mike: Actually, I don't know. Maybe you remember? 😉 https://github.com/MarquezProject/marquez/commit/524abc80879c1967ecaaf41951d1d58cebd576df

Issue #1501, opened 2021-07-12
URL-encode namespaces on calls to API backend


wslulciuc:
Namespaces can be URLs as of Marquez `0.16.0` (see https://github.com/MarquezProject/marquez/pull/1419). Currently, calls fail to the backend when attempting to fetch metadata with URLs as the namespace. To fix this issue, the UI should always URL encode the namespace.

PR #1502, opened 2021-07-12
Fixing job, dataset, and run url encodings.

phixMe:
Fixes: #1501 

This encodes certain entity names in our fetch URLs since they can include delimiter characters such as `/` that will not work.


wslulciuc (approved):
LGTM 👍

PR #1503, opened 2021-07-12
Fix DatasetVersionDao sql to support input and output facets

collado-mike:
The DatasetVersion SQL changes were pulling from the runs for a dataset version, rather than from the dataset versions table. This meant a single dataset version could have multiple rows, which violates the contract for Optional<Dataset> responses. This fixes the sql with added test cases


wslulciuc (approved):
Nice catch, @collado-mike! 💯 🥇

PR #1535, opened 2021-07-21
change Agent, Listener to openlineage wrapper

OleksandrDvornik:


collado-mike (approved):
This is great! One minor suggestion to just ditch the MarquezAgent class, as I don't think it's needed.

Issue #1527, opened 2021-07-15
Lineage response includes `type` twice in DatasetData


collado-mike:
The DatasetData implementation of `NodeData` has its own `type` field- either DB_TABLE or a STREAM. However, the `NodeData` interface is annotated to include `@JsonTypeInfo` using the `type` field. This results in the DatasetData rendering with two `type` fields.

Example response below
```json
{
  "graph": [
    {
      "id": "dataset:food_delivery:public.categories",
      "type": "DATASET",
      "data": {
        "type": "DATASET",
        "id": {
          "namespace": "food_delivery",
          "name": "public.categories"
        },
        "type": "DB_TABLE",
        "name": "public.categories",
        "physicalName": "public.categories",
        "createdAt": "2021-03-09T02:33:18.468719Z",
        "updatedAt": "2021-07-15T05:06:46.064956Z",
        "namespace": "food_delivery",
        "sourceName": "analytics_db",
        "fields": [
          {
            "name": "id",
            "type": "INTEGER",
            "tags": [],
            "description": "The unique ID of the category."
          },
          {
            "name": "name",
            "type": "VARCHAR",
            "tags": [],
            "description": "The name of the category."
          },
          {
            "name": "menu_id",
            "type": "INTEGER",
            "tags": [],
            "description": "The ID of the menu related to the category."
          },
          {
            "name": "description",
            "type": "TEXT",
            "tags": [],
            "description": "The description of the category."
          }
        ],
        "tags": [],
        "lastModifiedAt": "2021-07-15T05:02:15.064956Z",
        "description": null
      },
      "inEdges": [
        {
          "origin": "job:food_delivery:example.etl_categories",
          "destination": "dataset:food_delivery:public.categories"
        }
      ],
      "outEdges": [
        {
          "origin": "dataset:food_delivery:public.categories",
          "destination": "job:food_delivery:example.etl_orders_7_days"
        }
      ]
    }
]
```

julienledem:
also covers: https://github.com/MarquezProject/marquez/issues/1471

PR #1537, opened 2021-07-22
Update openapi spec with type names appropriate for code generation

collado-mike:
Specifically with the goal of supporting better code generation, I've updated the openapi spec with type names that are more code friendly. I also corrected some invalid types (namely arrays with no `items` field) and called out some specific enum values. 

I also tried to consolidate the different `Run` types in the spec into two types- `IncompleteRun` and `Run`. The former was previously included as several different models in various states. I believe this consolidation captures the essential difference in the Run model as it is returned in various states. 

Finally, I tried to include a `discriminator` for the `GraphNode#data` field. As specified in the [OpenApi docs](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) and called out in this post on [common mistakes](https://redoc.ly/docs/resources/discriminator/#common-mistakes), the discriminator field should be nested within the types unioned in the spec, but the `type` field in this spec is actually in the outer `GraphNode` type. The `type` declared in the `Dataset` and `Job` types are actually the type of Dataset and Job, respectively. _However_, in reality the lineage API returns both the `DatasetData` and `JobData` structures with the `type` field set to the `NodeType` value, as called out in https://github.com/MarquezProject/marquez/issues/1527 .

Using the discriminator makes code generation easier, as the generated code itself knows how to check the `type` field in order to figure out which version of the node data to return. Given that https://github.com/MarquezProject/marquez/issues/1527 needs to be addressed, I think it would be good to make a decision on what to do with that field and fix it here in the spec.


wslulciuc (changes requested):
@collado-mike Mind regenerating the `openapi.html` based off your changes using the `redoc-cli bundle` [command](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#api-docs)?

spec/openapi.yml
```diff
@@ -503,7 +503,14 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/newTag'
+              $ref: '#/components/schemas/CreatedTag'
+      responses:
+        '201':
```
wslulciuc, line 508: Since this is a `PUT` call, we return a `200`, see [`/api/v1/tags/{name}`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/TagResource.java#L69)

spec/openapi.yml
```diff
@@ -1041,55 +1061,19 @@ components:
...
-              items:
-                $ref: '#/components/schemas/datasetVersionId'
-            outputVersions:
-              type: array
-              items:
-                $ref: '#/components/schemas/datasetVersionId'
+          $ref: '#/components/schemas/Run'
+        facets:
+          $ref: '#/components/schemas/DatasetFacets'
+      required:
```
wslulciuc, line 1067: Do we need the `required` section in the dataset version response? Usually that to let the caller know what's required when making a request.

collado-mike: For the code generation, filling out the `required` section lets the generator know whether to generate nullable types. E.g., in Typescript, the fields can be typed as `string | undefined`, which forces the user to check for nulls before using the field. In Java, the code can be generated with `@NonNull` annotations so we don't accidentally return a null value when a real value is expected.

wslulciuc (commented):
Nice work on getting our openAPI spec in tip-top shape, @collado-mike! Had no idea about `discriminator`. So, awesome to learn about its usage. Left some comments, but otherwise this get's use one step closer to generated our core API and client models using the spec ❤️.

spec/openapi.yml
collado-mike: Hmm, per the HTTP spec, we should return a 201 if a resources is created and a 200 if it already exists. I can add both options to the response here.

spec/openapi.yml
wslulciuc: Ahh, very cool 👍. We'll want to update the other responses (can be in a follow up PR) to follow the same convention.

spec/openapi.yml
wslulciuc: I see, currently we only return `200`s for all `PUT` calls. So, I'd prefer we have the spec aligned with what's implemented. Mind opening an issue to update our `PUT` calls to return `200` for updates and `201` for new resources?

wslulciuc (approved):

PR #1547, opened 2021-07-27
dbt: add macro files to source distribution using MANIFEST.in

mobuchowski:
Fixes https://github.com/OpenLineage/OpenLineage/issues/106 which was caused by not including relevant macro files in distibution, which caused dbt to use generic macros.


wslulciuc (approved):
👍

PR #1548, opened 2021-07-27
openlineage wrapper for spark

OleksandrDvornik:
Add Spark subclasses to extend OpenLineageSparkListener and javaagent from OpenLineage #1536


OleksandrDvornik:
@collado-mike  I've tested your suggestion for using SparkAgent directly - it works! So, I've removed MarquezAgent class also

wslulciuc (changes requested):
@OleksandrDvornik ~We don't want to remove the code for the Spark agent, but rather just deprecate (see https://github.com/MarquezProject/marquez/pull/1545). Meaning, we'll want to remove the spark integration from the root [build.gradle](https://github.com/MarquezProject/marquez/blob/main/build.gradle) file to no longer hook into the project build. I'd prefer we write a simple Spark app to test the integration with OpenLineage in CI with a running instance of Marquez to assert the metadata collected. We'll still want to fetch the snapshot from jfrog, but the Spark app would be its own separate project for testing.~

collado-mike (approved):
Per offline discussion, we'll wait for the OpenLineage release before merging this

wslulciuc (approved):
NIICE! ❤️

integrations/spark/build.gradle
```diff
@@ -33,24 +33,12 @@ ext {
 }
 
 dependencies {
-    implementation 'org.javassist:javassist:3.28.0-GA'
-    implementation 'com.github.ok2c.hc5:hc5-async-json:0.2.2'
-    implementation 'org.apache.httpcomponents.client5:httpclient5:5.1'
-    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
-    implementation "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
-    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}"
+    implementation 'io.openlineage:openlineage-spark:0.1.0-SNAPSHOT'
```
wslulciuc, line 36: Mind updating the `openlineage-spark` dep to `0.1.0`?

integrations/spark/build.gradle
```diff
@@ -164,8 +152,8 @@ shadowJar {
 
     manifest {
         attributes(
-                "Premain-Class": 'marquez.spark.agent.MarquezAgent',
-                "Agent-Class": 'marquez.spark.agent.MarquezAgent',
+                "Premain-Class": 'openlineage.spark.agent.SparkAgent',
```
wslulciuc, line 155: Since this jar will still be published by Marquez, we'll want the keep `marquez.*`

integrations/spark/src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -1,3 +1,3 @@
 Manifest-Version: 1.0 
-Premain-Class: marquez.spark.MarquezAgent
-Agent-Class: marquez.spark.MarquezAgent
\ No newline at end of file
+Premain-Class: openlineage.spark.agent.SparkAgent
+Agent-Class: openlineage.spark.agent.SparkAgent
```
wslulciuc, line 3: Do we need this file? We already define the manifest in [`spark/build.gradle`](https://github.com/MarquezProject/marquez/blob/main/integrations/spark/build.gradle#L165)

integrations/spark/build.gradle
OleksandrDvornik: we can't change that, cause we don't have agent in marquez, we have only listener. This is an entry for javaagent, so it will try to find such class. By Mike comment, I've removed that SparkAgent from marquez code.

integrations/spark/src/main/resources/META-INF/MANIFEST.MF
OleksandrDvornik: I guess not, testing that

OleksandrDvornik: we don't need it, gradle will create it for us.

integrations/spark/build.gradle
wslulciuc: Ahh, makes sense 👍

integrations/spark/src/main/resources/META-INF/MANIFEST.MF
wslulciuc: Nice 👍

Issue #1560, opened 2021-08-03
Eliminate FieldType enum in Marquez


collado-mike:
Discussed in https://github.com/MarquezProject/marquez/issues/1552 , we should eliminate the FieldType enum as we lose precision in the types that we can support (e.g., `VARCHAR(128)` vs `VARCHAR(256)`).

PR #1561, opened 2021-08-04
Drop `FieldType` enum

wslulciuc:
This PR removes the `FieldType` enum to allow for any string provided by the caller to be collected. This avoids maintaining an enum of all possible field types. The field type value can be normalized when displayed in the UI (i.e. `varchar(64)` will be normalized to `VARCHAR(64)`), but in the database, the raw value will be persisted or `null` if the field type is not provided (or known). Fixes https://github.com/MarquezProject/marquez/issues/1560

Note, I've opened an issue to discuss normalization of field types https://github.com/MarquezProject/marquez/issues/1562


api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -537,7 +537,11 @@ default UUID version(
                 fields == null
                     ? ImmutableList.of()
                     : fields.stream()
-                        .map(field -> versionField(field.getName(), field.getType()))
+                        .map(
+                            field ->
+                                versionField(
+                                    field.getName(),
+                                    Utils.toUpperCaseIfPresent(field.getType().toUpperCase())))
```
OleksandrDvornik, line 544: I guess it should be `Utils.toUpperCaseIfPresent(field.getType())))`

api/src/main/java/marquez/db/DatasetVersionDao.java
```diff
@@ -103,7 +102,7 @@ default DatasetVersionRow upsertDatasetVersion(
               UUID.randomUUID(),
               now,
               field.getName().getValue(),
-              field.getType().name(),
+              Utils.toUpperCaseIfPresent(field.getType()),
```
OleksandrDvornik, line 105: proposal, maybe we can add this Utils.toUpperCaseIfPresent() logic to setter? Cause in all places we should do that even on getters, but if we apply that on object creation, we will be sure that it's uppercased

api/src/main/java/marquez/common/models/Field.java
```diff
@@ -36,7 +36,7 @@
   @Getter
   private final FieldName name;
 
-  @Getter private final FieldType type;
+  @Nullable @Getter private final String type;
```
OleksandrDvornik, line 39: should we apply the same for SchemaField to keep them in sync?

OleksandrDvornik (commented):
Please take a look on my comments

api/src/main/java/marquez/common/models/Field.java
wslulciuc: We'll eventually want to use the classes published by OpenLineage, but yeah to keep the models in sync I can also update the class `SchemaField`.

julienledem (approved):
Looks good to me. 
Why did we need to change the python tests? The enum was uppercase before. I'm assuming all types had to be uppercase?

wslulciuc:
> Why did we need to change the python tests? The enum was uppercase before. I'm assuming all types had to be uppercase?

Yeah, when I made the field type optional, and made sure the python client didn't normalize the value by calling `.upper()` (based on https://github.com/MarquezProject/marquez/issues/1562), I had to change the field type casing in the test. We also no longer need to check that an exception is thrown when the field type is not present.

Issue #1472, opened 2021-06-30
Marquez client model out of date


collado-mike:
The Marquez client model is pretty out of sync from the server-side model. Ideally, both the client and server-side models should be generated from the open api spec and always kept up to date.

julienledem:
We should generate models from the spec

PR #1609, opened 2021-08-30
Feature/dataset current version

### LINKED ISSUE #1283: Add `currentVersion` to Dataset and Job models
Let's add `currentVersion` to the Dataset and Job models returned by the API. The `currentVersion` field would be a pointer to the current version for a given dataset or job. For example, let's say we have a dataset with the `currentVersion` set to `b1d626a2-6d3a-475e-9ecf-943176d4a8c6`, we can use the DatasetVersion API to retrieve that version:

## `GET` `/api/v1/namespaces/my-namespace/datasets/my-dataset/versions/b1d626a2-6d3a-475e-9ecf-943176d4a8c6`

```
{
  "id": {
    "namespace": "my-namespace",
    "name": "my-dataset"
  },
  "type": "DB_TABLE",
  "name": "my-dataset",
  "physicalName": "public.mytable",
  "createdAt": "2019-05-09T19:49:24.201Z",
  "version": "b1d626a2-6d3a-475e-9ecf-943176d4a8c6",
  "namespace": "my-namespace",
  "sourceName": "my-source",
  "fields": [
    {"name": "a", "type": "INTEGER","tags": []},
    { "name": "b","type": "TIMESTAMP","tags": []},
    {"name": "c", "type": "INTEGER","tags": []},
    {"name": "d", "type": "INTEGER","tags": []}
  ],
  "tags": [],
  "description": "My first dataset!",
  "createdByRun": {
    "id": "0e89b60a-d80d-42ae-a12b-bd79ae1fb576",
    "createdAt": "2019-05-09T18:52:35.620763Z",
    "updatedAt": "2019-05-09T18:53:41.461122Z",
    "nominalStartTime": null,
    "nominalEndTime": null,
    "state": "COMPLETED",
    "startedAt": "2021-02-05T18:53:33.245492Z",
    "endedAt": "2021-02-05T18:53:41.461122Z",
    "durationMs": 8216,
    "args": {
      "email": "me@example.com",
      "emailOnFailure": false,
      "emailOnRetry": true,
      "retries": 1
    }
  }
}
```

OleksandrDvornik:
Resolves #1283
Add currentVersion to Dataset and Job models


spec/openapi.yml
```diff
@@ -963,6 +963,10 @@ components:
           type: string
         facets:
           $ref: '#/components/schemas/DatasetFacets'
+        currentVersionUuid:
```
wslulciuc, line 966: For versions, we don't specify that they are `UUIDs`. For example, in our [DatasetVersionAPI](https://marquezproject.github.io/marquez/openapi.html#tag/Datasets/paths/~1namespaces~1{namespace}~1datasets~1{dataset}~1versions~1{version}/get), we use `Dataset.version`. Mind using `currentVersion`?

clients/java/src/test/java/marquez/client/models/StreamTest.java
```diff
@@ -17,16 +17,23 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.UUID;
+import marquez.client.Utils;
 import org.junit.jupiter.api.Test;
 
 @org.junit.jupiter.api.Tag("UnitTests")
 public class StreamTest {
-  private static final Dataset STREAM = newStream();
```
wslulciuc, line 24: I would add a helper method `newStream(UUID)`, where `UUID` is the current version you'd like to assert. In the test below, I would still us `Stream.fromJson(JSON)` and not expose the mapper.

spec/openapi.yml
OleksandrDvornik: I can, but we have already such field in Dataset.
https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/service/models/Dataset.java#L64
Just made in the same way as it is

clients/java/src/test/java/marquez/client/models/JobTest.java
```diff
@@ -18,16 +18,23 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.UUID;
+import marquez.client.Utils;
 import org.junit.jupiter.api.Test;
 
 @org.junit.jupiter.api.Tag("UnitTests")
 public class JobTest {
-  private static final Job JOB = newJobWith(newRun());
```
wslulciuc, line 25: I would add a helper method `newJobWith(UUID, UUID)`, where `UUID` if for the runID and the current version you'd like to assert.

clients/java/src/test/java/marquez/client/models/DbTableTest.java
```diff
@@ -17,16 +17,22 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.UUID;
+import marquez.client.Utils;
 import org.junit.jupiter.api.Test;
 
 @org.junit.jupiter.api.Tag("UnitTests")
 public class DbTableTest {
-  private static final Dataset DB_TABLE = newDbTable();
```
wslulciuc, line 24: I would add a helper method `newDbTableWith(UUID)`, where `UUID` is the current version you'd like to assert.

clients/java/src/main/java/marquez/client/models/Stream.java
```diff
@@ -44,7 +45,8 @@ public Stream(
       @Nullable final Instant lastModifiedAt,
       @Nullable final URL schemaLocation,
       @Nullable final String description,
-      @Nullable final Map<String, Object> facets) {
+      @Nullable final Map<String, Object> facets,
+      @Nullable final UUID currentVersionUuid) {
```
wslulciuc, line 49: I would use `currentVersion` to follow current convention for versions.

api/src/main/java/marquez/service/models/Job.java
```diff
@@ -60,7 +62,8 @@ public Job(
       @Nullable final ImmutableMap<String, String> context,
       @Nullable final String description,
       @Nullable final Run latestRun,
-      @Nullable final ImmutableMap<String, Object> facets) {
+      @Nullable final ImmutableMap<String, Object> facets,
+      @NonNull Optional<UUID> currentVersionUuid) {
```
wslulciuc, line 66: I would continue to follow the convention in the class and pass the `UUID` without the `Optional` wrapper and define a `getCurrentVersion()` method that returns `Optional<UUID>`

spec/openapi.yml
wslulciuc: Yeah, for public fields, I wouldn't specify the UUID format 👍

clients/java/src/test/java/marquez/client/models/StreamTest.java
OleksandrDvornik: `newStream()` method is used only in this test, so I'll just change method signature for it.
You are right `Stream.fromJson(JSON)` is better approach.

clients/java/src/test/java/marquez/client/MarquezClientTest.java
```diff
@@ -479,7 +486,7 @@ public void testCreateDbTable() throws Exception {
             .build();
 
     final String metaAsJson = JsonGenerator.newJsonFor(meta);
-    final String dbTableAsJson = JsonGenerator.newJsonFor(DB_TABLE);
```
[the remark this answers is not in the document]
OleksandrDvornik: I don't see any reason to use that class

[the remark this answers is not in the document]
OleksandrDvornik: but the cons:
 - It should be extended at any model change (like my case) vs automatic mapping via ObjectMapper
 - Additional class

api/src/main/java/marquez/service/models/Job.java
OleksandrDvornik: what is a reason for doing that? If we have mix of approaches?
cons:
- (more code) I need to add getter manually
- it would create Optional object at every call, instead of once in constructor

clients/java/src/test/java/marquez/client/MarquezClientTest.java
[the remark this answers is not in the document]
wslulciuc: The `JsonGenerator.newJsonFor()` is similar to defining a json file that you load in your tests that represents the serialized representation of the object that you want to deserialized, then assert on. If you just use the `ObjectMapper`, yeah you don't have to worry about defining json (via `JsonGenerator.newJsonFor()` or a json file), but here you're just confirming that the `ObjectMapper` result is equal to the string returned by calling the models `fromJson()` method.

The idea behind `JsonGenerator.newJsonFor()` was to use the objects generated by our `ModelGenerators` to ensure the json has valid values. We can move to defining json files for our models, but that also means the author of the test has to define the values. Maybe that's not the worst thing.

That said, I would still keep `JsonGenerator.newJsonFor()`, but instead of having a specific type (`Dataset`, `Job`, etc), I would just wrap the mapper call to serialize:

```java
public static String newJsonFor(final Object o) {
    return MAPPER.writeValueAsString(o);
}
```

Then, you can delete the other methods in [`JsonGenerator`](https://github.com/MarquezProject/marquez/blob/main/clients/java/src/test/java/marquez/client/models/JsonGenerator.java). I think this would help with overall test maintenance.

api/src/main/java/marquez/service/models/Job.java
wslulciuc: We will be revising all API models for Marquez core, clients, etc when we [generate models from the OpenAPI spec](https://github.com/MarquezProject/marquez/issues/1472). I know currently `Optional<>` are used differently depending on the class (which we'll want to add guidelines for), but ["Optional" should not be used for parameters](https://rules.sonarsource.com/java/RSPEC-3553). I would conform to what's currently used in the class, but when we generate the models, this will no longer be something we have to write ourselves.

OleksandrDvornik: I also don't like to have it as a param, I prefer to have UUID value as a param, and then wrap it in constructor. It's not recommended to use Optional as a field also, but it relates to serialization/deserialization which works fine with jackson java8module

clients/java/src/test/java/marquez/client/MarquezClientTest.java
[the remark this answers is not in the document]
OleksandrDvornik: I leave it and add field there

api/src/main/java/marquez/service/models/Job.java
OleksandrDvornik: I've changed it in your way, should be ok

wslulciuc (approved):
Great work, @OleksandrDvornik 💯 🥇

PR #1610, opened 2021-08-31
Remove notebook example from spark integration

wslulciuc:
This PR removes the notebook example from our spark integration to encourage users to use [`openlineage-spark`](https://github.com/OpenLineage/OpenLineage/tree/main/integration/spark) as `marquez-spark` has been deprecated.


integrations/spark/docker-compose.yml
```diff
@@ -1,13 +0,0 @@
-version: "3.7"
```
fm100, line 1: Can you please put an explanation in the PR description why this is being removed?

fm100 (approved):
👍

PR #1622, opened 2021-09-08
Add search endpoint

wslulciuc:
This PR adds the `/search` endpoint to Marquez. The endpoint returns all datasets and jobs that match the provided query; matching of datasets and jobs is string based and case-insensitive. For example, below we query for all jobs containing the text "etl" sorted by "name" limiting our search results to `5`:

```
curl -X GET http://localhost:8080/api/v1/search?q=etl&filter=job&sort=name&limit=5
```

```
{
  "totalCount": 10
  "results": [
    {
      "type": "JOB",
      "name": "example.etl_drivers",
      "updatedAt": "2021-08-25T23:16:04.214438Z",
      "namespace": "food_delivery",
      "nodeId": "job:food_delivery:example.etl_drivers"
    },
    {
      "type": "JOB",
      "name": "example.etl_menus",
      "updatedAt": "2021-08-25T23:11:42.214438Z",
      "namespace": "food_delivery",
      "nodeId": "job:food_delivery:example.etl_menus"
    },
    .
    .
  ]
}
```


api/src/main/java/marquez/db/mappers/SearchResultMapper.java
```diff
@@ -0,0 +1,53 @@
...
+            DatasetName.of(stringOrThrow(results, Columns.NAME)),
+            timestampOrThrow(results, Columns.UPDATED_AT),
+            NamespaceName.of(stringOrThrow(results, Columns.NAMESPACE_NAME)));
+      case JOB:
+        return SearchResult.newJobResult(
+            JobName.of(stringOrThrow(results, Columns.NAME)),
+            timestampOrThrow(results, Columns.UPDATED_AT),
+            NamespaceName.of(stringOrThrow(results, Columns.NAMESPACE_NAME)));
+      default:
+        throw new IllegalArgumentException(String.format("search type '%s' not supported", type));
```
phixMe, line 50: I think we should be able to cover this line in the tests, right?

wslulciuc: Yeah, it's should never happen, but I'll add a tests!

api/src/main/java/marquez/db/SearchDao.java
```diff
@@ -0,0 +1,53 @@
...
+  @SqlQuery(
+      "SELECT type, name, updated_at, namespace_name\n"
+          + "FROM (\n"
+          + "  SELECT 'DATASET' AS type, d.name, d.updated_at, d.namespace_name\n"
+          + "    FROM datasets AS d\n"
+          + "   WHERE  d.name ilike '%' || :query || '%'\n"
+          + "   UNION\n"
+          + "  SELECT 'JOB' AS type, j.name, j.updated_at, j.namespace_name\n"
+          + "    FROM jobs AS j\n"
+          + "   WHERE  j.name ilike '%' || :query || '%'\n"
```
phixMe, line 47: I think we might need some testing for an empty string here. I believe it will return everything. Maybe also worth mentioning in the API docs.

api/src/main/java/marquez/db/mappers/SearchResultMapper.java
wslulciuc: Actually, there's no way to add a test as the enum `SearchType` only defines `DATASET` and `JOB`. We throw an `IllegalArgumentException` just in case we expand search to support other types.

api/src/main/java/marquez/db/SearchDao.java
wslulciuc: I think we'd want to return no results if the string is empty? I can add a check to immediately return if `q=`.

phixMe: Ok, that works as well.

phixMe (approved):
Looks good. Thanks for the changes and doc updates.

api/src/test/java/marquez/db/SearchDaoTest.java
```diff
@@ -0,0 +1,216 @@
...
+import marquez.jdbi.MarquezJdbiExternalPostgresExtension;
+import org.glassfish.jersey.internal.guava.Lists;
+import org.jdbi.v3.core.Jdbi;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/** The test suite for {@link SearchDao}. */
+@Tag("IntegrationTests")
```
collado-mike, line 36: Why is this tagged as an integration test?

api/src/test/java/marquez/db/SearchDaoTest.java
```diff
@@ -0,0 +1,216 @@
...
+    final List<SearchResult> results = searchDao.search(query, null, SearchSort.NAME, LIMIT);
+
+    // Ensure search results contain N datasets and M jobs.
+    assertThat(results).hasSize(NUM_OF_DATASETS + NUM_OF_JOBS);
+
+    // Group search results by result type.
+    final List<List<SearchResult>> resultsGroupedByType = groupResultsByType(results);
+
+    // Ensure search results contain exactly N datasets.
+    final List<SearchResult> resultsWithOnlyDatasets = resultsGroupedByType.get(0);
```
collado-mike, line 76: I don't think you can always rely on this ordering- datasets might be first or it might be second. I'd return a map keyed by `result.getType()` then get the list for the `SearchResult.ResultType.DATASET` key

api/src/test/java/marquez/db/SearchDaoTest.java
wslulciuc: Ahh, good point, I should be using the `DataAccessTests` tag. I'll open a PR to fix the tag.

wslulciuc:
@collado-mike: I addressed your feedback in https://github.com/MarquezProject/marquez/pull/1624

Issue #1630, opened 2021-09-13
The Marquez integration tests fail with OpenLineage 0.2.2


julienledem:
https://app.circleci.com/pipelines/github/MarquezProject/marquez/4655/workflows/0c614b86-a41b-48db-a294-3414b4b6af16/jobs/34862

It looks like it is expected a hardcoded 1-0-0 version of teh spec which has evolved to 1-0-1

PR #1632, opened 2021-09-13
Add OL version deb to root build.gradle

wslulciuc:
This PR ensures that both `marquez-api` and `marquez-spark` are using the same `openlineage-java` version.

Depends on #1625 being fixed, see https://github.com/MarquezProject/marquez/issues/1630


OleksandrDvornik (approved):
Nice!

PR #1652, opened 2021-09-21
Add simple `python` client example

wslulciuc:
This PR adds a simple `python` client example to outline usage. This PR also fixes a bug where if `mark_as_running` was set to `True`, an exception was thrown and now uses the `runID` present in _create run_ API response.


clients/python/examples/simple.py
```diff
@@ -0,0 +1,81 @@
...
+marquez_client.create_dataset(
+    namespace_name="my-namespace",
+    dataset_type=DatasetType.DB_TABLE,
+    dataset_name="my-dataset", # You can also use the physical name as the dataset name.
+    dataset_physical_name="public.mytable",
+    source_name="my-source",
+    fields=[
+        {'name': 'field0', 'type': 'INTEGER'},
+        {'name': 'field1', 'type': 'INTEGER'},
+        {'name': 'field2', 'type': 'STRING'}
```
fm100, line 54: Very minor, but can we use double quotation here for the sake of consistency?

clients/python/examples/simple.py
```diff
@@ -0,0 +1,81 @@
...
+
+# Now, let's create a run for 'my-job' and capture any runtime arguments:
+active_run = marquez_client.create_job_run(
+    namespace_name="my-namespace",
+    job_name="my-job",
+    mark_as_running=True
+)
+
+# ID of active run to mark as 'COMPLETE'.
+run_id = active_run['id']
```
fm100, line 79: Same for here, double quotation for consistency sake.

clients/python/marquez_client/client.py
```diff
@@ -303,7 +303,7 @@ def create_job_run(self, namespace_name, job_name, run_id=None,
             payload=payload)
 
         if mark_as_running:
-            response = self.mark_job_run_as_started(run_id)
+            response = self.mark_job_run_as_started(response['id'])
```
fm100, line 306: Same here too.

fm100 (approved):
With very minor comments, LGTM.

clients/python/examples/simple.py
wslulciuc: Sure, but most of the `py` client code uses single quotes, so I'll default to that 👍

PR #1655, opened 2021-09-22
Counts for lists of jobs and datatsets

phixMe:
This adds a generalized list class that various list operations can use to serialize responses. It includes implementations for `jobs` and `datasets` as well as basic queries to count the number of items in a particular namespace. 

This is useful for clients that wish to have paging operations, but it can also be used for informational purposes.

### Jobs list response
```JSON
{
  "jobs": ["..."],
  "totalCount": 15
}
```

### Checklist
- [x] Add Tests
- [x] Generate Docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased](https://github.com/MarquezProject/marquez/compare/0.18.0...HEAD)
 
+### Fixed
+
+* location/connectionUrl should be in URI format on write [@OleksandrDvornik](https://github.com/OleksandrDvornik)
+
 ## [0.18.0](https://github.com/MarquezProject/marquez/compare/0.17.0...0.18.0) - 2021-09-14
 
 ### Added
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -27,10 +27,10 @@ java {
 }
 
 ext {
-    jdbi3Version = '3.21.0'
+    jdbi3Version = '3.23.0'
     prometheusVersion = '0.12.0'
     testcontainersVersion = '1.16.0'
-    sentryVersion = '5.2.0'
+    sentryVersion = '5.2.2'
 }
 
 dependencies {
--- a/api/src/main/java/marquez/api/OpenLineageResource.java
+++ b/api/src/main/java/marquez/api/OpenLineageResource.java
@@ -15,6 +15,7 @@
 package marquez.api;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
 
 import com.codahale.metrics.annotation.ExceptionMetered;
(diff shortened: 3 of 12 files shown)


phixMe:
I wanted to push this into review a little early (before writing some tests) to get some feedback to see if we're comfortable with the implementation for `jobs` and `datasets`. Ultimately, we will want to expand this to other `list` type operations like datasets versions and runs, but I feel that this is a good first step that's non-breaking and strictly augments the existing API.

api/src/main/java/marquez/api/models/ListResult.java
```diff
@@ -0,0 +1,31 @@
+package marquez.api.models;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Collections;
+import java.util.Map;
+import lombok.NonNull;
+
+public class ListResult<T> {
```
collado-mike, line 10: Maybe `ResultsPage<T>`? Maybe also include the `offset` in this structure?

wslulciuc: I'd prefer the use of `ResultsPage<T>` as well, as it makes it clear the class is used for offset-based pagination. In addition to returning the `offset` and `totalCount`, I'd also add the `limit` to the response.

For clarification, `totalCount` is equal to the total number of _current_ entries in the `datasets`, `jobs`, etc tables?

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -194,6 +194,9 @@ default void setFields(Dataset ds) {
           + "ORDER BY d.name")
   List<Dataset> findAll(String namespaceName, int limit, int offset);
 
+  @SqlQuery("SELECT count(*) FROM datasets AS j WHERE j.namespace_name = :namespaceName")
+  int countFindAll(String namespaceName);
```
wslulciuc, line 198: I'd define the following methods for counts, as `findAll()` usually means that the method would return `n` rows:

```java
// total count
int count();

// total count by namespace
int countFor(namespaceName);
```

I would do that same for the `JobDao`

spec/openapi.yml
```diff
@@ -1424,6 +1427,9 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Job'
+        totalCount:
+          type: number
+          description: The total number of jobs for the given namespace
 
     CreatedRun:
```
wslulciuc, line 1434: Should we also document how to use the `limit` and `offset` for offset-based pagination?

wslulciuc (changes requested):
We would also need to update the following `marquez-java` client methods:

* [`MarquezClient.listDatasets()`](https://github.com/MarquezProject/marquez/blob/main/clients/java/src/main/java/marquez/client/MarquezClient.java#L168)
* [`MarquezClient.listJobs()`](https://github.com/MarquezProject/marquez/blob/main/clients/java/src/main/java/marquez/client/MarquezClient.java#L210)

And define a similar `ResultsPage<T>` class in the client

api/src/main/java/marquez/api/models/ListResult.java
phixMe: Sounds good on renaming the class.

Aren't `offset` and `limit` already known values? Those values would be reflecting input parameters back at the user; I'm not sure I see the value in that.

The `totalCount` value that I'm adding is so a client can know how many pages of data there are total when requesting any page.

api/src/main/java/marquez/db/DatasetDao.java
phixMe: Good idea!

Issue #1650, opened 2021-09-20
Binary incompatibility after deleting LineageEvent classes


collado-mike:
The change in https://github.com/MarquezProject/marquez/pull/1593 made the `marquez-api` jar incompatible with code that had depended on the `LineageEvent` class and its related classes. Any code that depended on those models must now be rewritten to rely on the `OpenLineage.*` models, which have a very different construction model, thus require a major effort to rewrite. 

Moreover, the current OpenLineage API has introduced new fields in the `InputDataset` and `OutputDataset` models, which were never present in the Marquez implementation of the OpenLineage models. The `LineageEvent` model is annotated with `@JsonIgnoreProperties` so any new fields in the JSON are simply dropped during deserialization. Therefore, simply reverting the `LineageEvent` models would make the Marquez backend incompatible with the new OpenLineage models as new facets would be dropped from the model before storing.

I think we should revert #1593 and alter the models to support unknown fields. Some options for this are
1. Add a `Map<String, Object>` field annotated with `@JsonAnySetter` so that any unknown fields are added to the map, rather than dropped.
    * This is little work up front and offers backward and forward compatibility, as any unknown fields are automatically supported. There is some maintainability concern, as we need to update the Marquez model alongside the OL one. 
2. Extend or wrap (using `@JsonUnwrapped`) Jackson `ObjectNode` so that objects are automatically deserialized into JsonNodes and setters/getters are written to work with expected properties in a compatible API
    * This is the most up-front work, but offers the most compatibility and least maintenance. Each model is backward and future compatible with any event POSTed and will always be serialized back into an exact replica of the original event. Accessor methods must be hand-written to replace the lombok-generated ones in order to maintain API compatibility.
3. Wrap new `OpenLineage` model classes with existing Marquez models
[...]

PR #1658, opened 2021-09-23
Revert lineage event

collado-mike:
Addresses https://github.com/MarquezProject/marquez/issues/1650 . Reverted the change in #1593 , but kept the Spark client references to the new `io.openlineage.*` package names. Nearly all of this PR is the revert. 

The commit at https://github.com/MarquezProject/marquez/commit/29769d5c44d79c65a658e2dfffd209286b0277bc adds a new base class for the `LineageEvent` and related models to support an `additionalProperties` map to store unknown fields. This makes the `LineageEvent` class forward and backward compatible with the OpenLineage `RunEvent` model and all other related models. To assert compatibility, I added a test for isomorphism between `RunEvent` and `LineageEvent`.


api/src/main/java/marquez/db/RunDao.java
```diff
@@ -193,7 +193,7 @@
           + "RETURNING *")
   ExtendedRunRow upsert(
       UUID runUuid,
-      UUID externalId,
+      String externalId,
```
wslulciuc, line 196: Good catch. An `externalId` isn't required to be a `UUID`, but rather a `UUID` is generated for the corresponding `externalId`

api/src/main/java/marquez/service/OpenLineageService.java
```diff
@@ -74,12 +75,29 @@ public OpenLineageService(BaseDao baseDao, RunService runService) {
...
   }
 
+  /**
+   * Try to convert the run id to a UUID. If it isn't a properly formatted UUID, generate one from
+   * the string bytes
+   *
+   * @param run
+   * @return the {@link UUID} for the run
+   */
+  private UUID runUuidFromEvent(LineageEvent.Run run) {
```
wslulciuc, line 91: _Minor_: A runID for OpenLineage should be a `UUID` according to the spec, but assuming this is for backwards compatibility?

api/src/main/java/marquez/service/models/ExtensibleJsonModel.java
```diff
@@ -0,0 +1,23 @@
+package marquez.service.models;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/** Base class for JSON models that supports additional properties for forward compatibility. */
+public class ExtensibleJsonModel {
```
wslulciuc, line 10: _Minor_: Mind if we use `BaseJsonObject` instead? just to keep aligned with the current naming `Base*`.

wslulciuc (approved):
Left a minor comment on naming for the class supporting Json forward compatibility I think we should rename, but otherwise great work 👍

Also opened an issue to capture our early discussion on _eventually_ deprecating the handwritten OpenLineage classes in Marquez, see https://github.com/MarquezProject/marquez/issues/1659

api/src/main/java/marquez/service/OpenLineageService.java
collado-mike: This was the original logic. I simply reverted the prior commit.
But yeah, this does seem like a good backward compatibility mechanism- the worst case is that the catch block is never executed.

api/src/main/java/marquez/service/models/ExtensibleJsonModel.java
collado-mike: I'm fine with that, but it may be worth pointing out that many of the JSON objects don't extend from this- i.e., all of the facets, as they are already extensible.

wslulciuc: Hmm, yeah fair point. And it's important to make that distinction. This is super nit-picky, so feel free to ignore, but  `ExtensibleJsonObject` just reads better 👌

collado-mike: Ah! I already merged before I saw your last comment. `BaseJsonObject` is what got through 🤷🏽‍♂️

Issue #1651, opened 2021-09-21
Proposal: Add support for `STRUCT` dataset field type


wslulciuc:
In Marquez, when collecting dataset metadata using the [DatasetAPI](https://marquezproject.github.io/marquez/openapi.html#tag/Datasets/paths/~1namespaces~1{namespace}~1datasets~1{dataset}/put), the caller can provide the fields of the dataset. The API can be used, for example, to collect the schema for a DB table. For a given dataset field, the following metadata is collected:

```
type
name
tags
description
``` 

The current size for the `type` column in the `dataset_fields` table is `VARCHAR(64)`. The size was set to allow for _primitive_ data types: `INT`, `STRING`, `BOOLEAN`, etc but not `STRUCT`s as supported by [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type), [Spark](https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/types/StructType.html), Redshift, etc. As a result, an exception is thrown (see below) when receiving a dataset with a `STRUCT` field type resulting in the dataset metadata not being registered with Marquez. For example, below is the metadata for the dataset field `file`:

```json
{
  "name": "file",
  "type": "STRUCT<filename STRING, project STRING, version STRING, type STRING>",
  "tags": [],
  "description": null
}
```

### `Exception`: value too long for type character `varying(64)`

```
Executed SQL: 
/* DatasetFieldDao.upsert */ 
INSERT INTO dataset_fields (uuid, type, created_at, updated_at, dataset_uuid, name, description) VALUES (?, ?, ?, ?, ?, ?, ?) ON CONFLICT(dataset_uuid, name, type) DO UPDATE SET updated_at = EXCLUDED.updated_at, description = EXCLUDED.description RETURNING *
JDBI SQL: 
/* DatasetFieldDao.upsert */ 
INSERT INTO dataset_fields (uuid, type, created_at, updated_at, dataset_uuid, name, description) VALUES (:uuid, :type, :now, :now, :datasetUuid, :name, :description) ON CONFLICT(dataset_uuid, name, type) DO UPDATE SET updated_at = EXCLUDED.updated_at, description = EXCLUDED.description RETURNING *
[...]

Issue #1649, opened 2021-09-20
`STRUCT` data type not supported for dataset field types


wslulciuc:
Marquez throws an exception when the dataset field type is **not** a _primitive_ data type:

```
Executed SQL: 
/* DatasetFieldDao.upsert */ 
INSERT INTO dataset_fields (uuid, type, created_at, updated_at, dataset_uuid, name, description) VALUES (?, ?, ?, ?, ?, ?, ?) ON CONFLICT(dataset_uuid, name, type) DO UPDATE SET updated_at = EXCLUDED.updated_at, description = EXCLUDED.description RETURNING *
JDBI SQL: 
/* DatasetFieldDao.upsert */ 
INSERT INTO dataset_fields (uuid, type, created_at, updated_at, dataset_uuid, name, description) VALUES (:uuid, :type, :now, :now, :datasetUuid, :name, :description) ON CONFLICT(dataset_uuid, name, type) DO UPDATE SET updated_at = EXCLUDED.updated_at, description = EXCLUDED.description RETURNING *
Binding: {positional:{0:44df86f3-78ee-4063-9112-022271f32647,1:2021-09-20T18:09:45.277712Z,2:file,3:STRUCT<filename STRING, project STRING, version STRING, type STRING>,4:null,5:4809a000-b457-404f-ad15-49789f46061f}, named:{now:2021-09-20T18:09:45.277712Z,name:file,description:null,type:STRUCT<filename STRING, project STRING, version STRING, type STRING>,datasetUuid:4809a000-b457-404f-ad15-49789f46061f,uuid:44df86f3-78ee-4063-9112-022271f32647}, finder:[]}
```

wslulciuc:
Fixed #1669

PR #1669, opened 2021-09-29
Change `type` column in `dataset_fields` table to `TEXT`

wslulciuc:
This PR changes the `type` column in `dataset_fields` table to `TEXT`. Note, this is a short term solution to support `STRUCTS` (see #1651), but also fixes exceptions thrown when the dataset field type is `> 64` characters (see #1649)


phixMe (approved):
Thanks @wslulciuc

fm100 (approved):
👍

PR #1670, opened 2021-09-29
Updated ZonedDateTime parsing to support optional offsets

collado-mike:
Currently, the GreatExpectations integration in OpenLineage is omitting the timezone offset from the event time. This needs to be fixed in the OpenLineage integration, but I think we should also be resilient to such errors in Marquez. Currently, we simply throw away the whole record.

This change allows for optional timezones and simply defaults to the system timezone if it is omitted. The GreatExpectations integration also posts fractions of a second in nanoseconds, rather than millis, so I added a test to validate that we can handle that format.

Also, I deleted the `testRequiredObjectMapper` test, as I'm not really sure what its purpose is. The test fails with this change because the ObjectMapper defined in the test code isn't configured to work with the timestamps in the test data. I could update the ObjectMapper to have the same configuration as the production one, but... I'm not really sure what the point of having that test would be... The test was originally introduced [here](https://github.com/MarquezProject/marquez/pull/880/files), but there are no comments explaining why

### DIFF
--- a/api/src/test/java/marquez/service/models/LineageEventTest.java
+++ b/api/src/test/java/marquez/service/models/LineageEventTest.java
@@ -2,18 +2,23 @@ package marquez.service.models;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import com.fasterxml.jackson.annotation.JsonInclude.Include;
-import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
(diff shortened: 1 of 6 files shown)


api/src/test/java/marquez/service/models/LineageEventTest.java
```diff
@@ -55,28 +68,25 @@ public void testSerialization(String input) throws IOException {
     testSerialization(Utils.newObjectMapper(), input);
   }
 
-  // Test object mapper with listed jackson requirements
-  @ParameterizedTest
-  @MethodSource("data")
-  public void testRequiredObjectMapper(String input) throws IOException {
```
wslulciuc, line 61: 👍

wslulciuc (approved):
LGTM 👍

PR #1674, opened 2021-09-30
Add ci job to build docs

wslulciuc:
This PR ensures our API docs are built in CI


.circleci/config.yml
```diff
@@ -15,6 +15,15 @@ only-on-release: &only-on-release
       ignore: /.*/
 
 jobs:
+  build-docs:
+    working_directory: ~/marquez
+    docker:
+      - image: circleci/node:12.21.0
```
fm100, line 21: `cimg/node:12.21` might be better: https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/

and omitting patch version will always pull the latest one among minor versions.

wslulciuc: Ahh, good point 👍

fm100 (approved):
👍

PR #1675, opened 2021-09-30
Navigate Dataset Versions

phixMe:
We're continuing to add new features to expose the knowledge of our Marquez model in our web interfaces. For this PR, we are adding the ability to navigate across dataset versions in the bottom panel within the lineage page. We also want to expose facet information (if we have it) for these dataset versions.

Preview Screenshots:
image
image
image


web/webpack.dev.js
```diff
@@ -14,7 +14,7 @@ const webpackDev = {
     },
     proxy: {
       '/api': {
-        target: `http://${process.env.MARQUEZ_HOST || 'localhost'}:${process.env.MARQUEZ_PORT || 8080}/`,
```
phixMe, line 17: Change this back.

wslulciuc:
@phixMe The UI for displaying dataset versions looks great! 💯 I only have a few minor suggestions:

1. For the dataset tab `CURRENT`, I would use either `SCHEMA` or `CURRENT SCHEMA` to make it clear that it's the schema for the datasets. That way, the `VERSIONS` table implies the `N` schema versions for the dataset
2. For the dataset tab `VERSIONS`, I would use the columns: `Version`, `Created At` and `Field Count`. I think it would be helpful to know which run ID created the dataset version, but we can expand on this later
3. Also, very minor, but I would use `Field` instead of `Attribute` when displaying the dataset schema

phixMe:
@wslulciuc All changes have been made.

wslulciuc (approved):
LGTM ❤️

PR #1677, opened 2021-10-01
Fix OpenLineage API to return correct response codes for exceptions

collado-mike:
Exceptions thrown during the OpenLineage API handling ends up always returning a 500, even if the underlying cause is something else. I added some code to handle exceptions in the async function execution and make some attempt at returning a proper response


api/src/test/java/marquez/OpenLineageIntegrationTest.java
```diff
@@ -42,6 +48,35 @@
...
+
+    final CompletableFuture<Integer> resp =
+        this.sendLineage(Utils.toJson(event))
+            .thenApply(HttpResponse::statusCode)
+            .whenComplete(
+                (val, error) -> {
+                  if (error != null) {
+                    Assertions.fail("Could not complete request");
+                  }
+                });
```
fm100, line 74: Aren't we expecting an error here? `Assertions.fail()` on error doesn't make sense to me but test seems to be passing. Am I misunderstanding something?

collado-mike: Seems to be an error only when there is no response from the server- e.g., can't connect

fm100: Oh I see.

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -65,13 +68,27 @@ public void create(
...
               }
             });
   }
 
+  private int determineStatusCode(Throwable e) {
+    if (e instanceof CompletionException) {
+      return determineStatusCode(e.getCause());
+    } else if (e instanceof IllegalArgumentException) {
+      return BAD_REQUEST.getStatusCode();
+    } else if (e instanceof MarquezHttpException) {
```
wslulciuc, line 83: Outside the integration tests, I don't think it's possible for an `MarquezHttpException` to be thrown. Do we need to handle the exception?

PR #1678, opened 2021-10-01
Add FlexibleDateTimeDeserializer mixin to the Dropwized Jackson ObjectMapper

collado-mike:
This change is needed to use the FlexibleDateTimeDeserializer in the Dropwizard Jackson ObjectMapper which is used to deserialize POST bodies


wslulciuc (approved):
LGTM 👍

PR #1688, opened 2021-10-06
Add option for docker/up to run in the background

rossturk:
Signed-off-by: Ross Turk <ross@rossturk.com>

I found myself hacking `docker/up.sh` regularly to add `-d` to the list of arguments so that I could keep Marquez running persistently. I tend to build models over several working sessions. Maybe this could be useful for others.

### DIFF
--- a/docker/up.sh
+++ b/docker/up.sh
@@ -19,7 +19,7 @@ title() {
 }
 
 usage() {
-  echo "usage: ./$(basename -- ${0}) [--api-port PORT] [--web-port PORT] [--tag TAG] [--build] [--seed]"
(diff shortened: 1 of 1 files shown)


docker/up.sh
```diff
@@ -47,6 +47,7 @@ usage() {
   title "FLAGS:"
   echo "  -b, --build           build images from source"
   echo "  -s, --seed            seed HTTP API server with metadata"
+  echo "  -d, --daemon          run in the background"
```
wslulciuc, line 50: Mind if we use `--detach` instead to match the docker option for running containers in [detached mode](https://docs.docker.com/engine/reference/run/#detached--d)?

rossturk: Ah, and here I always thought -d stood for "daemon" 😀

I'll update it.

Issue #1729, opened 2021-10-29
[Bug] Exception when dataset name contains `struct<>`


fm100:
Exception happens from `NodeId` class when the dataset node has name containing `struct<>`.
```
UUID string too large
java.lang.IllegalArgumentException: UUID string too large
	at java.base/java.util.UUID.fromString(UUID.java:199)
        at marquez.service.models.NodeId.asDatasetVersionId(NodeId.java:229)
```

Reproducible code:
```
    String datasetId = "dataset:my_namespace:my_struct<a:bigint,b:bigint,c:string>#54a90a2d-183c-4c3e-848c-d1622d6fd4ec";
    NodeId nodeId = NodeId.of(datasetId);
    nodeId.asDatasetVersionId();
```

https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/service/models/NodeId.java#L169-L198

`parts` method doesn't seem to be properly handling this case because it contains colon(`:`) inside dataset name.

PR #1730, opened 2021-10-30
Fix NodeId parsing bug when dataset name contains struct<>

fm100:
Signed-off-by: Minkyu Park <minkyu@datakin.com>

### Problem
For the dataset which contains `struct<>` inside its name, `NodeId.asDatasetVersionId` fails due to the parsing error. `struct<>` contains its sub fields and their types which contains colon(`:`), for example `struct<a:bigint,b:string>`, and the current regex only takes URL scheme(`://`) and port number(`:<port>`) into account.

Closes: #1729 

### Solution
This PR changes the logic of `asDatasetVersionId` and `asJobVersionId` so that they get their name along with the version, and then parse the version separately.

### Checklist


julienledem (approved):
+1
LGTM

Issue #1800, opened 2021-12-15
Add support for table operations reported from OpenLineage


collado-mike:
Some of the OpenLineage integrations can report table operations, such as
* Truncate
* Drop
* Rename

We should record these operations in Marquez and reflect the new state of the table after the operation has completed.

pawel-big-lebowski:
Some thought on how to achieve this. 

- **Collecting table state change information.**  - Table state change is a custom facet within outputs section of OpenLineage events. 
   - Add new column `operation` to `dataset_versions` table
   - Update `marquez.db.OpenLineageDao#updateBaseMarquezModel` and `marquez.db.OpenLineageDao#upsertLineageDataset` methods to store table operation when present. 
 -  **Exposing with API** 
    - Method `marquez.api.DatasetResource#getDataset` should be modified to contain optionally the latest table state change information. 
- **Presenting on Web UI** 
   - This information should be shown on the popup box with dataset info similarly to "Run duration" & "Run State"

TODO: how to reflect `rename`?

collado-mike:
This makes sense to me. 

As we discussed offline, a good approach to `rename` might be to add the original dataset as an input and the target dataset as an output, while also creating a new version of the original dataset with a `drop` operation. From the UI perspective, users might see the original dataset as no longer existing, while the lineage persists to the new table.

PR #1847, opened 2022-01-25
add lifecycleStateChange support

pawel-big-lebowski:

### Problem

Lacking support for table operations reported from OpenLineage

Closes: #1800

### Solution

Store `stateChange` field in backend database and expose the property over API. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased](https://github.com/MarquezProject/marquez/compare/0.21.0...HEAD)
 
+### Added
+
+* Add support for `LifecycleStateChangeFacet` with an ability to softly delete datasets. 
+
 ## [0.21.0](https://github.com/MarquezProject/marquez/compare/0.20.0...0.21.0) - 2022-03-03
 
 ### Added
--- a/api/src/main/java/marquez/common/Utils.java
+++ b/api/src/main/java/marquez/common/Utils.java
@@ -204,6 +204,7 @@ public final class Utils {
    * @param sourceName The source name of the dataset.
    * @param physicalName The physical name of the dataset.
    * @param datasetName The dataset name.
+   * @param lifecycleState The dataset change like CREATE, DROP, TRUNCATE.
    * @param fields The fields of the dataset.
    * @param runId The UUID of the run linked to the dataset.
    * @return A {@link Version} object based on the specified job meta.
@@ -213,6 +214,7 @@ public final class Utils {
       String sourceName,
       String physicalName,
       String datasetName,
+      String lifecycleState,
       List<LineageEvent.SchemaField> fields,
       UUID runId) {
     DatasetVersionData data =
@@ -221,6 +223,7 @@ public final class Utils {
             .sourceName(sourceName)
             .physicalName(physicalName)
             .datasetName(datasetName)
+            .lifecycleState(lifecycleState)
             .schemaFields(fields)
             .runId(runId)
             .build();
@@ -259,6 +262,7 @@ public final class Utils {
                 data.getPhysicalName(),
                 data.getSchemaLocation(),
                 data.getFields().stream().map(Utils::joinField).collect(joining(VERSION_DELIM)),
+                data.getLifecycleState(),
                 data.getRunId())
             .getBytes(UTF_8);
     return Version.of(UUID.nameUUIDFromBytes(bytes));
@@ -275,6 +279,7 @@ public final class Utils {
     private String sourceName;
     private String physicalName;
     private String datasetName;
+    private String lifecycleState;
     private String schemaLocation;
     private Set<Triple<String, String, String>> fields;
     private UUID runId;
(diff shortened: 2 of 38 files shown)


pawel-big-lebowski:
Zrzut ekranu 2022-01-26 o 17 15 42

wslulciuc:
@pawel-big-lebowski: though I agree we should handle the [`TableStateChangeFacet`](https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/src/main/common/java/io/openlineage/spark/agent/facets/TableStateChangeFacet.java), I'm not sure the facet is applicable to all dataset types (ex: `streams`, `files`, etc). That is, displaying the table state change in the web UI would be relevant for tables, but not for a file in `s3`.

After looking over the initial issue on [adding table operation support](https://github.com/MarquezProject/marquez/issues/1800) (and the facet definition), Marquez should perform certain _actions_ based on the table state change. That is, for each state changes the following actions are defined:

* `CREATE`: the dataset would be created and associated with the runID; but, based on the [`CreateTableCommandVisitor`](https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/CreateTableCommandVisitor.java) this state change would be a no-op as the table is registered as an output of the job run and would automatically be created by Marquez.
* `DROP`: the dataset would be a _soft_ delete; that is, Marquez would created a new _deleted_ version of the dataset and no longer add the dataset to the _current_ lineage graph view.
* `TRUNCATE` / `OVERWRITE`: this is essentially is an update on the dataset; on a completed job run, a new dataset version will be created automatically and associated with the runID.

If we refer back to issue #1800, @collado-mike mentions:

> We should record these operations in Marquez and reflect the new state of the table after the operation has completed.

The actions defined above would reflect the new state of the table, which is the intended goal. As for displaying or recording the state change as proposed in this PR?

1. If we decide **not** to display the state change in the web UI as proposed [here](https://github.com/MarquezProject/marquez/pull/1847#issuecomment-1022366982) (as least not initially), we still record all the events in the `lineage_events` table, and reflect the new state of the table after the action mapped to state change is applied to the dataset.
2. If we decide to display the state change in the web UI, we would need to think in terms of dataset lifecycle states (the lifecycle of a dataset is something that Marquez manages): `CREATED`, `DELETED`, `UPDATED`, etc. These lifecycle states would be generic enough to be associated with `streams`, `files`, etc. To be make it clear that the state change is associated with a given lifecycle state, I would use the column names `lifecycle_state` and `latest_lifecycle_state` in the tables `dataset_versions` and `datasets` respectively. I personally would prefer standardizing on the lifecycle states (defined as an `enum`) and mapping incoming state changes to them.

pawel-big-lebowski:
Thanks @wslulciuc for the extensive comment. I think the issue is that we started with table state changes in Spark and wanted to implement it in Marquez, while not being aware of a whole dataset context within Marquez. Like you said, the approach may not fit well other dataset types like files or streams. I agree with the column names `lifecycle_state` and `latest_lifecycle_state` which are more generic. 

Based on the knowledge collected, I think we should start with: https://github.com/OpenLineage/OpenLineage/issues/518
to define global facets' spec that matches tables, files, streams etc. Then we should implement it in Spark and later revisit this Marquez PR.

pawel-big-lebowski:
* I renamed the fields in JSON and backend DB according to naming convention introduced in spec (`lifecycleStateChange`). 
* I implemented soft-delete mechanism, so that in case of `drop` operation, a dataset is marked with a `is_deleted` flag equal to true. 

The still valid open question is: what do we want to do with this information on the frontend? There are at least two options:
 * Store dropped datasets in DB and do not show it on Web UI. This is implemented within current version. We collect the pretty useful information but we do not make any use of it at the moment. 
 * Display dropped datasets in UI while marking them deleted. The potential drawback is the amount of deleted datasets can increase over time and blur the graph. Perhaps we could assume configurable 7 days expiry time and just show datasets dropped within the expiry time.

@julienledem @wslulciuc What do you think?

pawel-big-lebowski:
We decided to proceed in a following way:
 * Datasets marked as `deleted` should be returned by the API.
 * Datasets marked as `deleted` should not be presented in UI. Filtering has to be done on the frontend side.
 * This approach is good enough for now and can be extended further on the frontend side later on.

wslulciuc (changes requested):
Great work, @pawel-big-lebowski! I have a few minor suggestions before we can merge this sweet feature

1. Mind using just `latest_lifecycle_state` and `lifecycle_state` for the column names to represent states? In OpenLineage, the facet does capture the state change (so the naming is appropriate), but in Marquez the column will represent the current state at a given time in the dataset lifecycle. You can think about how the run states are stored in the `runs` table.
2. I think we should also add the column `latest_lifecycle_state` to the `datasets` table?
3. With #1925 merged, you'll need to update you sql migration file to use` V41` and above
4. Mind also updating the changelog?

wslulciuc:
> First, this does not require any extra join in postgres as whenever dataset is fetched, it is already joined with the latest dataset_version. Secondly, there are other columns obtained from dataset_version table in a similar way like fields

Totally agree, thanks for pointing this out! The `datasets` table already has a reference to the current version (=`current_version_uuid `), which will also contain the current lifecycle state of the table. So, there's a stronger argument not to add the `latest_lifecycle_state` to `datasets` given that it can be derived for `dataset_versions` 👍

wslulciuc (approved):
Great work, @pawel-big-lebowski 💯 💯 💯

PR #1860, opened 2022-02-04
Add Postgres sub-chart to Helm deployment

KevinMellott91:
### Problem

The current Helm chart depends on an external Postgres database to function, which introduces challenges to fully automating quality checks via CI/CD.

Additionally, the recent SPDX license update #1846 is preventing the Helm chart from installing and produces the following error message.

`Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: apiVersion not set`

### Solution

Introduced the [Postgres Helm chart](https://github.com/bitnami/charts/tree/master/bitnami/postgresql) as an optional component, which allows Marquez to deploy into Kubernetes and function without depending on an external Postgres instance. The current limitation of this approach is that the Marquez test data seeding is not available; however, it's expected that can be added as a future enhancement to the Helm chart.

SPDX license updates were reverted from the Helm chart to restore deployment functionality.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 * Add MDC to the `LoggingMdcFilter` to include API method, path, and request ID [@fm100](https://github.com/fm100)
+* Add Postgres sub-chart to Helm deployment for easier installation option [@KevinMellott91](https://github.com/KevinMellott91)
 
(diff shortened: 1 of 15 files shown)


wslulciuc (approved):
@KevinMellott91 , thanks for the fix and for adding the postgresql subchart!

PR #1882, opened 2022-02-17
GitHub Action workflow to validate changes to Helm chart

KevinMellott91:
### Problem

When changes are made to the Marquez Helm chart, they need to be manually verified by installing and testing the chart into a local instance of Kubernetes (ex: Minikube). This process is both time consuming and prone to error.

### Solution

An automated GitHub Actions workflow has been created that performs the following checks whenever a pull request is opened that affects the Helm chart.

- Lint the chart to verify it is valid to the Helm specification
- Create a [kind](https://kind.sigs.k8s.io/) Kubernetes cluster and use it to deploy the Helm chart
- Execute the Helm chart tests, which verify the deployment completed successfully and the pods/services are available

All of these checks can executed locally as well, using the [act](https://github.com/nektos/act) utility. Additional instructions have been added to the contribution guide.

### Checklist


.actrc
```diff
@@ -0,0 +1,2 @@
+# These default configurations will be applied when running act locally for GitHub Action testing
+-P ubuntu-latest=ghcr.io/catthehacker/ubuntu:act-latest
```
wslulciuc, line 2: When [first running `act`](https://github.com/nektos/act#first-act-run), the user is asked what default image to use. So, I see the `.actrc` file more of a user defined configuration that may change from user-to-user and not something we'd want to checkin.

.github/workflows/lint-test-chart.yaml
```diff
@@ -0,0 +1,44 @@
+name: Lint and Test Chart
```
wslulciuc, line 1: For workflows, we may want to define a naming convention for `.yml` files:

```bash
build-api.yml         # run build, unit tests, linter
build-client-java.yml # run build, unit tests, linter
test-helm.yml         # run linter, test against kind k8s cluster
...
```

That is, I would take a general-to-more-specific naming convention `[action]-[group]-[module]`, where `[group]` is optional. It's a similar approach we took for naming CI jobs in CircleCI. But, that might result in a growing number of workflows (well, maybe not, `< 10`).


We could also just want to map workflows to Marquez's sub-projects defined 
```
api.yml
web.yml
clients.yml
spec.yml
chart.yml
```

Personally, it might be better to have more discrete workflows, then define them per sub-projects. Thoughts?

chart/ci/ci-values.yaml
```diff
@@ -0,0 +1,4 @@
+## These value overrides are used within the GitHub Actions CI validation.
+##
+postgresql:
+  enabled: true
```
wslulciuc, line 4: Is this file defined for CI overrides only?

chart/templates/marquez/deployment.yaml
```diff
@@ -20,6 +20,53 @@ spec:
...
+      initContainers:
+        - name: wait-for-db
+          image: {{ include "common.images.image" (dict "imageRoot" .Values.postgresql.image "global" .Values.global) }}
+          imagePullPolicy: {{ .Values.postgresql.image.pullPolicy  }}
+          command:
+            - /bin/bash
+          args:
+            - -ec
+            - |
+              #!/bin/bash
```
wslulciuc, line 34: You may just want to use:

```bash
while ! pg_isready -h ${POSTGRES_HOST}; do echo "Waiting for postgres to become available..."; sleep 2; done;
```

ct.yaml
```diff
@@ -0,0 +1,9 @@
+# See https://github.com/helm/chart-testing#configuration
```
wslulciuc, line 1: In terms of organization, I'd prefer if this file was defined under `chart/`. I suggest:
1. We rename the `chart/ci/` to `chart/.ci/` to convey the directory is not part of the chart and for internal use only
2. Move / rename `ci-values.yaml` to `chart/.ci/values.yaml`
3. Move `ct.yaml` to `chart/.ci/ct.yaml`

```
chart/.ci/
  ct.yaml
  values.yaml
```

.actrc
KevinMellott91: I created this file so that the user wouldn't need to decide [which image](https://github.com/nektos/act#runners) to choose, since the minimal one doesn't have the necessary packages pre-installed to operate. I'm wondering if it would be better to remove this user setting file and clarify in the documentation section instead?

.github/workflows/lint-test-chart.yaml
KevinMellott91: Great idea! I like the option of combining the Marquez sub-module with the action being taken. How about this for the file in this PR?

_test-chart.yml_

If we were to extend the workflows to include Helm chart releasing, then we could add another file that would be called _release-chart.yml_. Not planning that right now, just wanted to test how the nomenclature could look.

chart/templates/marquez/deployment.yaml
KevinMellott91: That does look much simpler, thanks!

ct.yaml
KevinMellott91: Unfortunately the helm chart-testing utility is [hard-coded](https://github.com/helm/chart-testing/blob/0a093e464b884565148eb71dda6999de0c956b17/pkg/chart/chart.go#L225) to only recognize the folder named "ci" within a given chart. However, I might be able to configure the GitHub Actions workflow in a way that will allow for the ct.yaml file to be moved into the "ci" folder.

chart/ci/ci-values.yaml
wslulciuc: Nice! Thanks for clarifying.

.github/workflows/lint-test-chart.yaml
wslulciuc: Yeah, let's use `test-chart.yml`, but with the assumption that things will change. I need to review the docs for GitHub Actions as well. I'm thinking we might have:

```
.github/workflows/
  ci.yml
  release.yml
```
as a way to further simply our workflows. Just a thought of course, but will have a proper discussion when we begin to replace CircleCI with GitHub Actions.

.actrc
wslulciuc: Yeah, personally, I would have a section on how to configure the `act` cli in the `CONTRIBUTING.md` under the [development](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#development) section.

KevinMellott91:
@wslulciuc I've uploaded changes based on your comments - thanks for the great suggestions! Note that I also pinned the version of Postgres to 12.1 to match what is being used within the rest of the Marquez project.

wslulciuc (approved):
LGTM 💯

Issue #1758, opened 2021-11-06
Internal server error 500 - field name too long


mandy-chessell:
Marquez returns a 500 if the following open lineage event is sent to it:
```
{"producer":"https://egeria-project.org/","schemaURL":"https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/RunEvent","eventType":"START","eventTime":"2021-11-06T14:15:45.536Z","run":{"runId":"b978fa80-e28c-4891-908d-63e84b7c830a","facets":{"nominalTime":{"_producer":"https://egeria-project.org/","_schemaURL":"https://openlineage.io/spec/facets/1-0-0/NominalTimeRunFacet.json#/$defs/NominalTimeRunFacet","nominalStartTime":"2021-11-06T14:15:33.398Z"},"additionalProperties":{}},"additionalProperties":{}},"job":{"namespace":"governance-action-process:clinical-trials:drop-foot:weekly-measurements:onboarding","name":"provision-weekly-measurements-governance-action-type"},"inputs":[{"namespace":"governance-action-process:clinical-trials:drop-foot:weekly-measurements:onboarding","name":"sourceFile"}]}
```
This is the error log:
```
marquez-db                  | 2021-11-06 14:15:48.409 UTC [83] ERROR:  value too long for type character varying(64)
marquez-db                  | 2021-11-06 14:15:48.409 UTC [83] STATEMENT:  /* NamespaceDao.upsertNamespaceRow */ INSERT INTO namespaces ( uuid, created_at, updated_at, name, current_owner_name ) VALUES ($1, $2, $3, $4, $5) ON CONFLICT(name) DO UPDATE SET updated_at = EXCLUDED.updated_at RETURNING *
marquez-api                 | ERROR [2021-11-06 14:15:48,445] marquez.api.OpenLineageResource: Unexpected error while processing request
marquez-api                 | ! org.postgresql.util.PSQLException: ERROR: value too long for type character varying(64)
marquez-api                 | ! at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2565)
marquez-api                 | ! at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2297)
marquez-api                 | ! at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:322)
[...]

PR #1925, opened 2022-03-24
Add migration script to remove size limits from namespaces, dataset n…

collado-mike:
### Problem

Several columns in the database are restricted to 255 characters, perhaps without strong justification. Several of these cause problems, including namespaces, in #1758, as well as the `connection_url` when storing some Snowflake tables (the connection URLs contain a lot of extra parameters which aren't always scrubbed before sending the OpenLineage events). If there isn't a strong reason for restricting these columns, the default restriction on VARCHARs will work just as well with no performance penalty.

Closes: [#1758](https://github.com/MarquezProject/marquez/issues/1758)

### Solution

Remove the size restriction on various VARCHAR types

### Checklist


api/src/main/resources/marquez/db/migration/V40__alter_tables_resize_namespace_and_connection_url.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE dataset_versions ALTER COLUMN namespace_name TYPE VARCHAR;
+ALTER TABLE dataset_versions ALTER COLUMN dataset_name TYPE VARCHAR;
+ALTER TABLE datasets ALTER COLUMN name TYPE VARCHAR;
+ALTER TABLE datasets ALTER COLUMN physical_name TYPE VARCHAR;
+ALTER TABLE datasets ALTER COLUMN source_name TYPE VARCHAR;
+ALTER TABLE namespaces ALTER COLUMN name TYPE VARCHAR;
+ALTER TABLE runs ALTER COLUMN external_id TYPE VARCHAR;
+ALTER TABLE sources ALTER COLUMN name TYPE VARCHAR;
+ALTER TABLE sources ALTER COLUMN connection_url TYPE VARCHAR;
+ALTER TABLE tags ALTER COLUMN name TYPE VARCHAR;
```
fm100, line 10: So these are to change from VARCHAR(256) to VARCHAR, right?

fm100 (approved):
👍

api/src/main/resources/marquez/db/migration/V40__alter_tables_resize_namespace_and_connection_url.sql
collado-mike: yep

PR #1929, opened 2022-03-30
Web: Created by view for dataset versions

phixMe:

### Problem
Some Marquez users were eager to see the run that created a dataset version.

### Solution
We've added the `Created by Run` section on the dataset page as well as the dataset version list view. Also updated is syntax highlighting support for `sql` contexts and the empty states for context that is missing for us.
image
image
image
image

### Checklist

### DIFF
--- a/web/src/components/core/code/MqCode.tsx
+++ b/web/src/components/core/code/MqCode.tsx
@@ -1,31 +1,30 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { THEME_EXTRA } from '../../../helpers/theme'
-import { Theme, alpha } from '@material-ui/core/styles'
+import { THEME_EXTRA, theme } from '../../../helpers/theme'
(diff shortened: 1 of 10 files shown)


wslulciuc:
@phixMe, love to see the progress on this! I'm wondering, can we simplify the change by just linking the run that created the dataset? That is, in the version history tab, can we just have the `runID` be "clickable" and link out to the run metadata? I like the idea of showing the state of the run, but a dataset version is created on a successful job run, so it'd always be "green" (=complete run state). So, I think just the `runID` would be enough.

In the latest schema tab, I would also just display the `runID` (maybe on the top right?) as well and make that clickable. Given that the run metadata comes after the `fact` metadata, the run metadata would be far down the page if the run contained a large fact value.

wslulciuc:
@phixMe: To not block the merge, let's address any feedback as follow up work

wslulciuc (approved):
LGTM 🛳️

Issue #1928, opened 2022-03-29
Supporting Job grouping and hierarchy in Marquez


collado-mike:
# Problem

In [https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#parent-job-run-a-nested-hierarchy-of-jobs](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#parent-job-run-a-nested-hierarchy-of-jobs) we use parent jobs and job naming to describe an extended DAG of related jobs. However, Marquez never tracks parent/child job relationships.

As a concrete example, consider a data pipeline that calculates metrics for a website experimentation service:

Each hour:

	Collect clickstream data from the website and copy it to cloud storage

	Analyze clickstream data and determine samples for bucket A and bucket B

	Read current and past sample population buckets and calculate per customer metrics for each running experiment

	Aggregate hourly customer metrics for each running experiment

Each day:

	Aggregate daily customer metrics for each running experiment

	Aggregate cumulative experiment metrics for each running experiment

The workflow consists of two DAGs, one hourly and one daily. The lineage would look like the following

![experiment_metrics_workflows](https://user-images.githubusercontent.com/40346148/160709002-5060871b-1c4f-4473-a596-df3286b96d39.png)

Customer metrics are generated hourly and experiment level metrics are aggregated from them so that major regressions in high volume metrics (clicks, page latency) can be caught right away. Hourly customer metrics also aggregated to generate daily customer level metrics. Once daily customer metrics are available, aggregate experiment-level metrics are calculated for all active days of the experiment. The code for aggregating hourly customer-level metrics and daily customer-level metrics is the same. Only the input (hourly vs daily customer metrics) and output (experiment-level metrics) differ.

To complicate things, assume the experiment metrics workflow executes once per marketplace region- e.g., once for U.S., once for U.K., once for JP, etc., each in its local timezone (a day in JP ends at midnight JST). 

## Grouping and Naming by Parent Jobs

In [https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#parent-job-run-a-nested-hierarchy-of-jobs](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#parent-job-run-a-nested-hierarchy-of-jobs) , we state

[...]

PR #1935, opened 2022-04-05
Job parent hierarchy

collado-mike:
### Problem

Database migration as described in https://github.com/MarquezProject/marquez/issues/1928 . This implements the database migration necessary to create the new columns (`job_uuid` and `parent_run_id` in `runs` and `parent_job_uuid` in `jobs`.

### Solution
This PR addresses only the migration. After merging this, a deployment will have the new columns and values will be present, but aren't being actively written by the code and aren't being read. The migration is intended to be fully backward compatible,  so if a deployment needs to be rolled back, the original data/behavior is preserved. 

I added a `jobs_view` that calculates the fully qualified name rather than storing the FQN directly. With this implementation, a parent job can be renamed and child jobs can automatically be renamed (we'll need a separate issue to automatically redirect, as that's not currently handled in this impl). We also use the simple name of the job to display in the UI (currently relying on parsing the dot characters in the name). If we store the FQN directly, we'll need to parse the FQN to derive the simple name for display. On the other hand, constructing the FQN after storing the simple name is easy and cheap. While the view adds some complexity, I think it gives us direct access to the information we want without forcing us to use heuristics to figure it out.

> **Note:** All database schema changes require discussion. Please [link the issue](ttps://github.com/MarquezProject/marquez/issues/1928) for context.

### Checklist


api/src/main/java/marquez/db/JobDao.java
```diff
@@ -78,16 +78,24 @@
   }
 
   @SqlQuery(
-      "SELECT j.*, n.name AS namespace_name FROM jobs AS j "
-          + "INNER JOIN namespaces AS n "
-          + "  ON (n.name = :namespaceName AND "
-          + "      j.namespace_uuid = n.uuid AND "
-          + "      j.name = :jobName)")
+      """
```
wslulciuc, line 81: Love the new multiline formatting 😍

api/src/main/java/marquez/db/JobDao.java
```diff
@@ -107,10 +115,10 @@
           + "LIMIT :limit OFFSET :offset")
   List<Job> findAll(String namespaceName, int limit, int offset);
 
-  @SqlQuery("SELECT count(*) FROM jobs AS j")
+  @SqlQuery("SELECT count(*) FROM jobs_view AS j")
   int count(String namespaceName);
```
wslulciuc, line 119: _Minor:_ Not related to this PR at all, but `namespaceName` isn't used in the SQL

api/src/main/java/marquez/db/JobDao.java
```diff
@@ -78,16 +78,24 @@
...
-          + "INNER JOIN namespaces AS n "
-          + "  ON (n.name = :namespaceName AND "
-          + "      j.namespace_uuid = n.uuid AND "
-          + "      j.name = :jobName)")
+      """
+          SELECT j.* FROM jobs_view AS j
+          WHERE j.namespace_name = :namespaceName AND
+          j.name = :jobName
+      """)
   Optional<JobRow> findJobByNameAsRow(String namespaceName, String jobName);
```
wslulciuc, line 86: Does this query return the FQN of the job? I'm just trying to understanding the difference between `findJobByNameAsRow()` and `findJobBySimpleNameAsRow()` when using the view vs not.

wslulciuc: I see the `job_fqn` function defined in the migration `V41__alter_jobs_add_job_parent_id.sql` constructs the fully qualified name.

api/src/main/java/marquez/graphql/GraphqlDaos.java
```diff
@@ -163,87 +178,90 @@
...
+                  left outer join job_versions_io_mapping io_of_out
+                    on io_of_out.job_version_uuid = j_of_out.current_version_uuid
+                    and io_of_out.io_type = 'OUTPUT'
+                  group by io_of_out.dataset_uuid
+              ) o on o.dataset_uuid = io_out.dataset_uuid
+              WHERE io_out.io_type = 'INPUT'
+              group by io_out.job_version_uuid
+          ) d_out on d_out.job_version_uuid = j.current_version_uuid
+  """)
   List<JobResult> getLineage(String jobName, String namespaceName, Integer depth);
```
wslulciuc, line 266: Thanks for also updating the graphql SQL!

api/src/main/java/marquez/db/migrations/V43__UpdateRunsWithJobUUID.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * This migration is dependent on the migration found in the SQL script for V42. This updates the
+ * runs table to include the <code>job_uuid</code> value for each record. We update the table in
+ * batches to avoid table-level locks so that concurrent reads and writes can continue to take
+ * place. Auto-commit is enabled, so it is entirely possible that this migration will fail partway
+ * through and some records will retain the <code>job_uuid</code> value while others will not. This
+ * is intentional as no harm will come from leaving these values in place in case of rollback.
+ */
+@Slf4j
+public class V43__UpdateRunsWithJobUUID extends BaseJavaMigration {
```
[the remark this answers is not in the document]
collado-mike: TBH, I wrote it in Java because I wanted to be able to loop through the results and update a batch of run records at a time, rather than locking the whole table for writes. I _could_ do this in plpgsql, but I don't have much experience writing those kinds of looping functions, so I went with my comfort zone 😀. I can take a stab at doing this in pure SQL...

Re: the ordering, I could rename the Java class to be `V42a...` or some other suffix so that it comes alphanumerically after the `V42` SQL migration, but allow new migration scripts to continue with `V43__`, etc. But I'll leave that as a last resort if the pure SQL function doesn't pan out

api/src/main/java/marquez/graphql/GraphqlDaos.java
wslulciuc: I know, was visually looking to see if you missed anything 👀

api/src/main/resources/marquez/db/migration/V41__alter_jobs_add_job_parent_id.sql
```diff
@@ -0,0 +1,33 @@
+ALTER TABLE jobs ADD COLUMN IF NOT EXISTS parent_job_id uuid CONSTRAINT jobs_parent_fk_jobs REFERENCES jobs (uuid);
+ALTER TABLE runs ADD COLUMN IF NOT EXISTS parent_run_id uuid CONSTRAINT runs_parent_fk_runs REFERENCES runs (uuid);
```
wslulciuc, line 2: _Minor_: We use `*_uuid` internally for IDs within the database (when referencing the primary key). For example, the API returns a `run.id`, but in the database it's really just the `uuid`. This also avoids any confusion with `jobId` or `datasetId`s which is a `namespace` + `name` pairing.

for consistency, I recommend using:

```
parent_job_uuid
parent_run_uuid
```

api/src/main/resources/marquez/db/migration/V41__alter_jobs_add_job_parent_id.sql
```diff
@@ -0,0 +1,33 @@
...
+DROP INDEX IF EXISTS jobs_name_index;
```
wslulciuc, line 3: How expensive would it be to keep the index? (Wondering if that's just a good index to have).

api/src/main/java/marquez/graphql/GraphqlDaos.java
wslulciuc: (GraphQL is also in beta, but to move it out of beta we'd need a proper test suite).

api/src/main/resources/marquez/db/migration/V41__alter_jobs_add_job_parent_id.sql
```diff
@@ -0,0 +1,33 @@
...
+           j.type,
+           j.created_at,
+           j.updated_at,
+           j.namespace_uuid,
+           j.description,
+           j.current_version_uuid,
+           j.current_job_context_uuid,
+           j.current_location,
+           j.current_inputs
+           FROM job_fqn f, jobs j WHERE j.uuid=f.uuid
```
wslulciuc, line 33: What are your thoughts on being more explicit with the job _fully qualified name_? That is, if someone was querying the `jobs` table, the `name` column would be the _simple_ name, but when querying `jobs_view` the name is now the fully qualified name (though the `name` column is renamed for added context).

I find it confusing to repurpose the `name` column and would prefer to add a `fully_qualified_name` to `job_views`. That said, that comes with major drawbacks given the renaming nightmare. If we consider the response:

```
{
  "id": {
  "namespace": "my-namespace",
  "name": "parent-job.my-job"
  },
  "type": "BATCH",
  "name": "parent-job.my-job",
  .
  .
```

`name` in this context would be the fully qualified name, and as outlined in your [proposal](https://github.com/MarquezProject/marquez/issues/1928) we'd add a `simpleName` field. I find this reasonable.   This change would also mean, `jobId`s will now use the fully qualified name. So maybe I talked myself out of using `fully_qualified_name` 😅 

I guess my thoughts are that the view does provide guarantees for the migration, but we are left with long term "maintenance". For example, any columns added to `jobs` would also be added to `jobs_view`, any newcomers to the project would have to be aware of this view, etc. I think this is why initially I preferred `option 2` in your propsoal:

> 2. Do not backfill the jobs or runs tables. Jobs and runs can continue to use the simple names in tables and those names will be translated at runtime when the API response is constructed.

wslulciuc: That said, we don't expect the db model to be queried directly and the API model is the public contract. I'm just wondering if there's a scenario where a `simple_name` column is added to `jobs` and we backfill the `name` column in place if the `parent_job_uuid` is not `null`. If `null` both `name` and `simple_name` would contain the same value.

wslulciuc: But, you've also thought about this in more detail than I have! The goal is to support job grouping, so if that means we add views and ensure we minimize user downtime, then no objections on my end. Just thought I'd bring up a few things but certainly not a blocker to move forward.

collado-mike: Option 2 in the proposal requires either querying the view or changing the code to query the correct tables/columns. E.g., if a request comes in for `parent-job.my-job`, we can't simply query the unaltered `jobs` table, since the `name` column will only contain `my-job`. So the either the code will need to parse the job name (very bad idea, given that there are jobs whose names already contain a dot character) or it will have to query a table/view that has the FQN pre-constructed. So even if we did absolutely no backfilling, the view is likely the _most_ maintainable option going forward. To get away with no view at all requires option 1 - update all the job name in the `jobs`, `job_version`, `runs`, and `lineage_events` tables. We can add the `simple_name` column for the `jobs` table, but everything else would be the FQN. Assuming that migration is successful, there's 0 maintenance going forward. But that migration is potentially expensive.

api/src/main/resources/marquez/db/migration/V42__runs_job_versions_add_job_uuid.sql
```diff
@@ -0,0 +1,46 @@
...
+       r.updated_at,
+       job_version_uuid,
+       run_args_uuid,
+       nominal_start_time,
+       nominal_end_time,
+       current_run_state,
+       start_run_state_uuid,
+       end_run_state_uuid,
+       external_id,
+       location,
```
wslulciuc, line 37: _Minor:_ We've been meaning to rename the `location` column to `job_location` as the name is confusing and reads more like "the location of the run". Would be great to get the name right for the view 🙏

wslulciuc (approved):
Thanks for entertaining my thoughts and feedback @collado-mike. Left some minor comment, and suggestion to rename a few columns, but otherwise great work! 💯 🥇

api/src/main/resources/marquez/db/migration/V41__alter_jobs_add_job_parent_id.sql
collado-mike: It's replaced with a different index

api/src/main/resources/marquez/db/migration/V43__alter_jobs_add_job_parent_id.sql
```diff
@@ -0,0 +1,44 @@
+ALTER TABLE jobs ADD COLUMN IF NOT EXISTS parent_job_uuid uuid CONSTRAINT jobs_parent_fk_jobs REFERENCES jobs (uuid);
```
wslulciuc, line 1: _Minor: I would also update the migration script to reflect the name of the column added:

```
V43__alter_jobs_add_job_parent_uuid.sql
```

wslulciuc (approved):
Thanks for addressing my comments! Left one minor suggestion on naming, otherwise 👍

Issue #1967, opened 2022-04-25
Web not able to access API when using Helm charts 


raghanag:
Hi All, I have tried to use helm install in my local kind cluster but when I access `localhost:3000` I see the below error, looks like the expected proxy is not able to connect to `localhost:5000` where API is running at. 

I don't see the same issue when I used docker/up.sh, please let us know how to fix this.

here is the screenshot: 
![image](https://user-images.githubusercontent.com/11847063/165148261-bcc86eac-3fb6-4e4a-9bd6-c5509efb6634.png)

API running at `localhost:5000`: 
![image](https://user-images.githubusercontent.com/11847063/165148460-8adb481f-8c2b-4ba8-8fbe-d6455ec54c53.png)

raghanag:
@wslulciuc Looks like the web is not able to talk to API only in k8s, I can see that when I run docker-compose.yaml, it is working as expected.

raghanag:
error response: 
![image](https://user-images.githubusercontent.com/11847063/165222342-9083e329-1dfc-4c5f-95ec-b3991a1a7a6d.png)

GZack2000:
**Why is this happening?**
This is because in the helm chart `values.yaml`, `marquez.hostname` has the value `localhost`
```yaml
  ## Marquez will run using this hostname
  ##
  hostname: localhost
  ## Marquez API will run on this port
  ##
  port: 5000
```
This will not work when deployed in K8s as the web UI front-end and marquez API backend are in different pods. When the web application's proxy tries to connect to `localhost:5000`, it gives 'connection refused' as that refers to localhost:5000 within the web pod.

`localhost:5000` can only be used from within the marquez API pod, which is why the API endpoints are accessible when port-forwarding.

To enable UI frontend pod to talk to the marquez API backed pod deployed in K8s cluster, use K8s Service way of accessing (`<service-name>:<port>`) [1]

**How do I fix this?**
The workaround I followed was modifying the `MARQUEZ_HOST` to use name of the service as
```yaml
env:
  - name: MARQUEZ_HOST
    # value: {{ .Values.marquez.hostname | quote }}
    value: {{ include "common.names.fullname" . }}
```

So if I deploy the helm chart like `helm install marquez . --dependency-update --set postgresql.enabled=true`
the service name being `marquez`,  when UI tries to talk to the backend, it uses `http://marquez:5000`

***References***
[1] https://www.tutorialworks.com/kubernetes-pod-communication/

PR #1970, opened 2022-04-27
Helm chart: Enable web app to access API

GZack2000:
### Problem
In the helm chart `values.yaml`, `marquez.hostname` has the value `localhost`
```yaml
  ## Marquez will run using this hostname
  ##
  hostname: localhost
  ## Marquez API will run on this port
  ##
  port: 5000
```
This will not work when deployed in K8s as the web UI front-end and marquez API backend are in different pods. When the web application's proxy tries to connect to `localhost:5000`, it gives 'connection refused' as that refers to localhost:5000 within the web pod.

`localhost:5000` can only be used from within the marquez API pod, which is why the API endpoints are accessible when port-forwarding.

To enable UI frontend pod to talk to the marquez API backed pod deployed in K8s cluster, use K8s Service way of accessing (`<service-name>:<port>`) [1]

***References***
[1] https://www.tutorialworks.com/kubernetes-pod-communication/

Closes: #1967

### Solution

Modify the `MARQUEZ_HOST` environment variable used in the marquez-web deployment to use name of the service :
```yaml
env:
  - name: MARQUEZ_HOST
    value: {{ include "common.names.fullname" . }}
```
With this template, the web app should always be able to talk to the API service as it uses the K8s service address

Once the helm chart is deployed like `helm install marquez . --dependency-update --set postgresql.enabled=true`
the service name being `marquez`,  when UI tries to talk to the backend, it uses `http://marquez:5000`

### Checklist

### DIFF
--- a/chart/templates/web/deployment.yaml
+++ b/chart/templates/web/deployment.yaml
@@ -39,7 +39,7 @@ spec:
               port: http
           env:
             - name: MARQUEZ_HOST
-              value: {{ .Values.marquez.hostname | quote }}
+              value: {{ include "common.names.fullname" . }}
             - name: MARQUEZ_PORT
               value: {{ .Values.marquez.port | quote }}
           {{- if .Values.web.resources }}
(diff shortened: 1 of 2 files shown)


wslulciuc (approved):
Thanks for the fix, @GZack2000! 💯 🥇

PR #1980, opened 2022-05-07
Add java migrations for backfilling runs with job uuids and parents

collado-mike:
### Problem
Continuation of https://github.com/MarquezProject/marquez/pull/1935/files , this adds the migration scripts for populating job parents for Airflow runs and other runs that have explicit parent run ids. Special handling is given to the Airflow tasks to match the [behavior in OpenLineage](https://github.com/OpenLineage/OpenLineage/pull/664). Unit tests are provided to verify the behavior of the backfill scripts

> **Note:** All database schema changes require discussion. Please [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) for context.

### Checklist

- [X] You've [signed-off](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#sign-your-work) your work
- [X] Your changes are accompanied by tests (_if relevant_)
- [X] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [ ] You've updated the [`CHANGELOG.md`](https://github.com/MarquezProject/marquez/blob/main/CHANGELOG.md#unreleased) with details about your change under the "Unreleased" section (_if relevant, depending on the change, this may not be necessary_)
- [X] You've versioned your `.sql` database schema migration according to [Flyway's naming convention](https://flywaydb.org/documentation/concepts/migrations#naming) (_if relevant_)

PR #1992, opened 2022-05-16
Job parent hierarchy api changes

collado-mike:
### Problem

Final PR for #1928, continued from https://github.com/MarquezProject/marquez/pull/1980. This updates the behavior of the write APIs to update the job parent field for new events and updates the read APIs to return the `simpleName` field of the job as well as the FQN. Notably, parent jobs and parent runs are created if present in the OpenLineage event but not present in the Marquez database. This handles events from Airflow DAGs where the DAG is a parent job for all tasks even though no event is ever sent for the DAG itself. A few integration tests added to validate the behavior for receiving messages from Airflow and Spark.

Closes: #1928 

### Solution

### Checklist


api/src/main/java/marquez/db/JobDao.java
```diff
@@ -278,6 +296,34 @@ JobRow upsertJob(
...
+      UUID namespaceUuid,
+      String namespaceName,
+      String name,
+      String description,
+      UUID jobContextUuid,
+      String location,
+      UUID symlinkTargetId,
+      PGobject inputs) {
+    UUID jobUuid =
+        upsertJobNoParent(
```
wslulciuc, line 312: Wouldn't we want the `upsertJobNoParent()` call to the `JobRow` object similar to other upsert calls?

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -104,37 +113,155 @@ default UpdateLineageRow updateBaseMarquezModel(LineageEvent event, ObjectMapper
...
+            .map(t -> t.withZoneSameInstant(ZoneId.of("UTC")).toInstant())
+            .orElse(null);
+    Instant nominalEndTime =
+        Optional.ofNullable(event.getRun().getFacets())
+            .flatMap(f -> Optional.ofNullable(f.getNominalTime()))
+            .map(NominalTimeRunFacet::getNominalEndTime)
+            .map(t -> t.withZoneSameInstant(ZoneId.of("UTC")).toInstant())
+            .orElse(null);
+
+    Logger log = LoggerFactory.getLogger(OpenLineageDao.class);
```
wslulciuc, line 147: I would annotate the `OpenLineageDao` class with `@Slf4j`

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -104,37 +113,155 @@ default UpdateLineageRow updateBaseMarquezModel(LineageEvent event, ObjectMapper
...
+            });
+
+    // construct the simple name of the job by removing the parent prefix plus the dot '.' separator
+    String jobName =
+        parentJob
+            .map(
+                p -> {
+                  if (event.getJob().getName().startsWith(p.getName() + '.')) {
+                    return event.getJob().getName().substring(p.getName().length() + 1);
+                  } else {
```
wslulciuc, line 224: _Minor_: We may want to move this in a `DbUtils` class to handle parsing the simple name:

```java
DbUtils.simpleJobNameFor()
```

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -104,37 +113,155 @@ default UpdateLineageRow updateBaseMarquezModel(LineageEvent event, ObjectMapper
...
+                        now,
+                        namespace.getUuid(),
+                        namespace.getName(),
+                        jobName,
+                        description,
+                        jobContext.getUuid(),
+                        location,
+                        null,
+                        jobDao.toJson(toDatasetId(event.getInputs()), mapper)))
+            .orElseGet(
```
wslulciuc, line 251: Nice use of `orElseGet()` to query for the job without a parent 💯

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -190,9 +192,10 @@ SELECT run_uuid, JSON_AGG(json_build_object('namespace', namespace_name,
           + "nominal_start_time = COALESCE(EXCLUDED.nominal_start_time, runs.nominal_start_time), "
           + "nominal_end_time = COALESCE(EXCLUDED.nominal_end_time, runs.nominal_end_time), "
           + "location = EXCLUDED.location "
-          + "RETURNING *")
-  ExtendedRunRow upsert(
+          + "RETURNING uuid")
```
wslulciuc, line 195: Do we want to change the contract here to return `uuid` instead of `ExtendedRunRow`?

api/src/main/java/marquez/service/models/Job.java
```diff
@@ -42,6 +44,8 @@ public Job(
       @NonNull final JobId id,
       @NonNull final JobType type,
       @NonNull final JobName name,
+      @NonNull String simpleName,
```
wslulciuc, line 47: Do we have to update the UI to use the _simple name_ when displaying the job name?

api/src/main/java/marquez/db/JobDao.java
collado-mike: Sorry, do you mean that the `upsertJobNoParent` query would go back to `RETURNING *` instead of `RETURNING uuid`? If that's what you mean, I made this change so that the subsequent `findJobByUuidAsRow` call queries the `jobs_view` - returning the FQN rather than the simple name.

api/src/main/java/marquez/db/OpenLineageDao.java
collado-mike: Can't - its an interface and `@Slf4j` only works on classes

wslulciuc (approved):
@collado-mike left some minor comments, but otherwise great work 💯 💯 🥇 

As for keeping the `jobs_view`, here are my thoughts (also in response to our _offline_ discussion):

_Upside_:

Queries remain simple, meaning when querying the `jobs` table, the `name` column is still the simple name of the job, not the FQN. Also, the web UI should display the simple name of the job and depending on how jobs are named, parsing the FQN for displaying may result in the wrong name being used (not ideal). I think given the scope of the change, `jobs_view` allows us to avoid any unknown cases around job naming. The view can also be seen as a migration step to eventually having the `name` column in the `jobs` table be the FQN. Meaning, we can add a `simple_name` column to the `jobs` table ensuring the simple name and FQN are clearly defined and possibly dropping the view all together (or keep it arounds as there are clear benefits).

_Downside_: 

I think having the `name` and `simple_name` column in the `jobs` table would ensure the FQN or the simple name would always be referenced correctly (outside just the view). But, a deeper discussion on how much benefits this provides can be had as the REST API is how metadata should be queried for in the first place.

api/src/test/java/marquez/OpenLineageIntegrationTest.java
```diff
@@ -51,6 +73,27 @@ public static List<String> data() {
         EVENT_NAMESPACE_NAMING);
   }
 
+  @AfterEach
+  public void tearDown() {
+    Jdbi.create(POSTGRES.getJdbcUrl(), POSTGRES.getUsername(), POSTGRES.getPassword())
+        .withHandle(
+            handle -> {
+              handle.execute("DELETE FROM lineage_events");
```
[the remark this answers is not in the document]
collado-mike: yeah...

api/src/main/java/marquez/db/OpenLineageDao.java
wslulciuc: Ah you're right

Issue #1956, opened 2022-04-19
Choose non-default port for Postgres in docker/up.sh


rossturk:
I keep running into situations where I want to run Marquez alongside Airflow (or another system that also runs a Postgres database) and the two docker-compose environments both attempt to start a Postgres database. The "correct" workaround is to create a single docker-compose environment with a shared database for both systems, but this is not always ideal.

It would be nice if we could tell `docker/up.sh` to start Postgres on an alternate port.

collado-mike:
TBH, there's not a lot of reason to map the database port to any specific port on the host. The Marquez API will get routed by Docker to whatever port it's actually bound to. The only useful thing we get out of mapping to a specific port is that we can use `psql` on the command line to query the database during development. `docker port marquez-db 5432` is all we really need to find what port to connect to, though

rossturk:
It's not about accessing the DB, it's about other tools complaining that they can't start their own databases because the port is in use.

collado-mike:
Yeah, I got it. I'm saying there will be no conflicts if we bind `5432` to a _random_ port (if we change [this](https://github.com/MarquezProject/marquez/blob/main/docker-compose.yml#L37) line to just `5432` instead of `5432:5432`).

The API container will still be able to hit the database because docker magic will translate `db:5432` to whatever actual port it's bound to. Only commands outside the docker network will have problems (i.e., `psql` on the command line), but that can be addressed with the `docker port` command.

PR #1994, opened 2022-05-17
Update Getting Started Guide with note on port 5000 for MacOS

RNHTTR:
### Problem

Port 5000 is now a reserved port on MacOS

### Solution

Include note in getting started guide for how to run the API server on a custom port

### Checklist

Signed-off-by: Ryan Hatter <ryannhatter@gmail.com>


README.md
```diff
@@ -52,6 +52,8 @@ The Marquez [HTTP API](https://marquezproject.github.io/marquez/openapi.html) li
 
 > **Note:** By default, the HTTP API does not require any form of authentication or authorization.
 
+> **Note:** Port 5000 is now reserved for MacOS. If running locally on MacOS, you can run `./docker/up.sh --api-port 9000` to configure the API to listen on port 9000 instead. Keep in mind that you will need to update the URLs below with the appropriate port number.
```
wslulciuc, line 55: Thanks for adding calcification around the port! I was wondering, should we add the note closer to where the `./docker/up.sh` script is run? I.e. right under the [Quickstart](https://github.com/MarquezProject/marquez#quickstart) section?

RNHTTR: Yep, that makes way more sense! Long day :)

collado-mike:
Per my comment in https://github.com/MarquezProject/marquez/issues/1956 , there's not really a strong reason for us to map any of these ports to any specific host port. Only the web UI needs to be on a deterministic host port - the rest of the containers should probably rely on Docker to manage the port mapping behind the scenes

wslulciuc:
@collado-mike, we could allow Docker to manage the port, but there are two use-cases for starting the API container. One being the web UI, but let's say you want to `curl` the API, or use one of our clients, knowing the port upfront is helpful.

wslulciuc:
@RNHTTR Don't forget you have to [sign your commit](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#sign-your-work) ;) 

More detail here: https://github.com/MarquezProject/marquez/pull/1994/checks?check_run_id=6479779860

collado-mike:
> One being the web UI, but let's say you want to `curl` the API, or use one of our clients, knowing the port upfront is helpful.

Those are fair use cases - the former can be handled using the `docker port` command (a little clunky, but... 🤷🏽‍♂️). Or we can also add a docker-compose override file to map to a specific port, but leave the default mapping random. We could use something like a `--fixed-api-port XXXX` or something to determine when to include the override, but by default everything would work on Mac or any other system (as long as the web port isn't conflicting).

wslulciuc:
@collado-mike: In favor of making standing up Marquez locally on a Mac, and avoiding the port issue, we'll have docker assign the port (see https://github.com/MarquezProject/marquez/issues/1956). We can merge for now, but make this port conflict changes as follow up work.

wslulciuc (approved):
LGTM 👍

PR #1995, opened 2022-05-19
Fixed key value helm input

rejulius:
Signed-off-by: Julius Rentergent <julius.rentergent@thetradedesk.com>

Using `key value` pairs causes `Error: "helm upgrade" requires 2 arguments` as tested on Helm v3.8.2

Changing this to `key=value` pairs fixes the issue


wslulciuc (approved):
Thanks, @rejulius! 💯 🥇

PR #1996, opened 2022-05-19
add copyright lines to all source files

merobi-hub:
Signed-off-by: Michael Robinson <merobi@gmail.com>

### Problem

The project's source files lack a copyright, but inclusion of a copyright notice in every source file is an expectation of the LFAI. 

### Solution

This PR adds a copyright line to source files across the project (.py, .java, .sh), excluding non-source files and files where the line could create an issue.

### Checklist


mobuchowski (changes requested):
From the experience of doing the same in OpenLineage project - it would be better if this was a series of commits, each focused on particular file type only. This would allow spotting problems with particular file types.

For example, the copyright statement being top of file in `.sh` files is wrong - it makes shebang not work properly. This is not easily found when looking through 300+ files 😉

merobi-hub:
Thanks, Maciej. I separated the changes into individual commits, one per file type.

clients/python/marquez_client/__init__.py
```diff
@@ -1,3 +1,5 @@
+# Copyright 2018-2022 contributors to the Marquez project
+
 # Licensed under the Apache License, Version 2.0 (the "License");
```
wslulciuc, line 3: The `__init__.py` is still using the old license header

wslulciuc (changes requested):
Thanks for adding the copyright to the source files, @merobi-hub! I did look over the files, and have the following suggestions:

* Remove the copyright from all `*.txt` and `*.md` files
* For `*.sh` files, the copyright has been added to the bottom of the scripts, and should be in the file header
* Is there a way to combining the copyright and license in a single header? That is:

`java`

```java
/* 
 * Copyright 2018-2022 contributors to the Marquez project
 * SPDX-License-Identifier: Apache-2.0
 */
```
`bash`
```bash

   
#!/bin/bash
#
# Copyright 2018-2022 contributors to the Marquez project
# SPDX-License-Identifier: Apache-2.0
# ...
```
`py`
```python
# Copyright 2018-2022 contributors to the Marquez project
# SPDX-License-Identifier: Apache-2.0
```

merobi-hub:
Thanks for the feedback, @wslulciuc . Curious about why you want the line removed from .md and .txt files, assuming they're original to the project. Can you clarify?
@mobuchowski requested the line be added to the bottom of Bash scripts. Can we come to a consensus?
Reformatting the lines would mean starting over and debugging again. Are we okay with extending the timeline for this?
Thanks.

wslulciuc:
> Curious about why you want the line removed from .md and .txt files, assuming they're original to the project. Can you clarify?

```
The project MUST include a copyright statement in each source file
```

This is up for interpretation, but source files usually refer to code `.java`, `.py`, etc, and not docs.  But ok I see, since the copyright for the `*.md` file is hidden using `<!-- -->`, you can leave it, Initially I had assumed it broke the formatting. For new docs, we'll want to add the copyright. Have you looked into adding the copyright to the files under [`docs/`](https://github.com/MarquezProject/marquez/tree/main/docs)?

For the `*.txt` files, we can leave them but would address my most recent comment on the file `web/src/fonts/Karla/OFL.txt`

> Reformatting the lines would mean starting over and debugging again. Are we okay with extending the timeline for this?

Normally, I won't care as much but this is such a rare and infrequent change that will be copied for all new source files that I'm ok with extending the timeline to use a more compact header (like I outlined in my change request). I know, not fun and really sorry for the extra effort here. I owe you!

merobi-hub:
Hi @mobuchowski and @wslulciuc , I think this is ready to be merged

wslulciuc (approved):
Amazing work, @merobi-hub! 💯 🥇

Issue #1711, opened 2021-10-15
Proposal: Add paging support to clients


wslulciuc:
We have recently added offset-based pagination support in #1655 to Marquez. Next, we'll need to define a API contract so that our clients can support pagination when calling list APIs. Both the `java` and `python` [clients](https://github.com/MarquezProject/marquez/tree/main/clients) support listing datasets and jobs (as well as runs,  etc).

## How are datasets and jobs listed _without_ paging?

**`python`**

The list methods return a `dict` containing a list of datasets and jobs:

```python
MarquezClient.list_datasets()
MarquezClient.list_jobs()
```

**`java`**

The list methods return a `List<T>` containing a list of datasets and jobs:

```java
List<Dataset> MarquezClient.listDatasets()
List<Job> MarquezClient.listJobs()
```

## How will datasets and jobs listed _with_ paging?

To avoid any breaking changes and allow the deprecation of the methods above, I proposal we overload the list methods by accepting an `Option` class:

**`python`**

```python
MarquezClient.list_datasets(ListOption)  -> ResultsPage[Dataset]
MarquezClient.list_jobs(ListOption)  -> ResultsPage[Job]
```

**`java`**

```java
ResultsPage<Dataset> MarquezClient.listDatasets(ListOption)
ResultsPage<Job> MarquezClient.listJobs(ListOption)
```

where **`ListOption`** is defined as:

```java
/** Class for specifying list options. */
class ListOption {
  int limit;
  int offset
}
```

and the resulting paging class **`ResultsPage<T>`** defined as:

```java
/** A Page object wraps an API list method response. */
class ResultsPage<T> {
  /** Returns true if there are more pages that can be retrieved. */
  boolean hasNextPage();
  /** Retrieves the next Page object using the limit and offset. */
  Page<T> getNextPage();
  /** Returns elements in this page. */
  List<T> getValues();
}
```

PR #2018, opened 2022-06-14
Reduce UI limit for list datasets jobs in UI

wslulciuc:
### Problem

The current UI default limit to list datasets and jobs is `2000` which can be an expensive query. A more ideal approach would be to introduce paging (see https://github.com/MarquezProject/marquez/issues/1711), but for now, let's use a reasonable default `25`.

### Solution

Reduce limit for list datasets jobs from `2000` to `25`.

### Checklist


phixMe (approved):
Thanks @wslulciuc

Issue #2022, opened 2022-06-24
[Bug] Databricks namespace containing `+` being rejected.


conorbev:
I'm seeing a case where I have a namespace similar to: `databricks+connector://dbc-47b23571-1011.cloud.databricks.com` and lineage run events including that namespace are being rejected with a message like:

```
"java.lang.IllegalArgumentException: namespace 'databricks+connector://dbc-47b23571-1011.cloud.databricks.com' must contain only letters (a-z, A-Z), numbers (0-9), underscores (_), dashes (-), colons (:), equals (=), semicolons (;), slashes (/) or dots (.) with a maximum length of 1024 characters.
```

Can we add `+` to the list of allowed characters in namespace names?

PR #2023, opened 2022-06-24
namespaces: allow '+' in names

mobuchowski:
Fix rejecting databricks namespaces with `+`. 

Closes: #2022


wslulciuc (approved):
LGTM 💯 ... also, looks like we're slowly expanding namespaces to include all characters 😅

Issue #1736, opened 2021-11-01
Ability to DELETE jobs and datasets


julienledem:
When jobs or datasets are deleted (or renamed) we need a way to tell Marquez that they don't exist anymore so that they don't clutter the lineage view.
It would make sense to have `DELETE` support on `/api/v1/namespaces/{ns}/datasets/{name}` and `/api/v1/namespaces/{ns}/jobs/{name}` resources.
This is a soft delete in the internal marquez model. It creates a new `deleted` version. The resources will not show up in the "current" lineage view in `GET /api/v1/lineage`

julienledem:
also see #754

mobuchowski:
Done in https://github.com/MarquezProject/marquez/pull/2032

PR #2032, opened 2022-07-06
delete: add possibility to soft-delete datasets and jobs

mobuchowski:
First part of #1736 - ability to "soft delete" datasets and jobs - API to "hide" them which can be used from UI; the reason for adding this is process to hiding inactive datasets and jobs. This PR does not include the UI part.

The feature works by adding `is_hidden` flag to both datasets and jobs tables. Then, it changes `jobs_view` and adds `datasets_view` that hide rows where `is_hidden` flag is set to true value. This makes writing proper queries easier, since, you don't need to do this filtering manually everywhere as long as you use this view.

The soft-delete is reversed if the job or dataset is updated again - new version reverts the flag.


api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -138,51 +138,53 @@ default void setFields(Dataset ds) {
...
-          + "    AND ds ->> 'namespace' = d2.namespace_name\n"
-          + "    GROUP BY d2.uuid\n"
-          + ") f ON f.dataset_uuid = d.uuid\n"
-          + "ORDER BY d.name")
+      """
+      WITH selected_datasets AS (
+          SELECT d.*
+          FROM datasets d
+          WHERE d.namespace_name = :namespaceName
+          AND d.is_deleted is false
```
pawel-big-lebowski, line 146: Current strategy is: API returns soft deleted datasets with an a extra `is_deleted: true` flag.
Doesn't this change current behavior of the app?

pawel-big-lebowski:
This looks really great. The only inconsistency I see is that: 
 * `findBySomething` does return soft-deleted datasets, 
 *  `findAll` does not return them.
 Such a behavior definitely makes here sense but can confuse other contributors. 

What if we introduced `includeDeleted` parameter in the URL and make default values `true` and `false` for `findByName/findAll` respectively. ?

mobuchowski:
@pawel-big-lebowski changed how this works. Now it all references view.

mobuchowski:
@wslulciuc can you take a look? 

Also, can you think of any additional tests this might need?

api/src/test/java/marquez/service/LineageServiceTest.java
```diff
@@ -203,6 +210,119 @@ && jobNameEquals(n, "downstreamJob0<-outputData<-readJob0<-commonDataset"))
...
+
+    // assert that readJob is hidden
+    assertThat(lineage.getGraph())
+        .filteredOn(
+            n ->
+                n.getType().equals(NodeType.JOB)
+                    && jobNameEquals(n, "downstreamJob0<-outputData<-readJob0<-commonDataset"))
+        .isEmpty();
+  }
+
```
pawel-big-lebowski, line 325: That's a great test! 💯 🥇

pawel-big-lebowski (approved):
I fell in love with `testLineageWithDeletedDataset` test ❤️

mobuchowski:
Next time I won't change query formatting AND query itself in one PR - it makes resolving conflicts unpleasant 😞

api/src/main/java/marquez/api/DatasetResource.java
```diff
@@ -150,6 +151,28 @@ public Response list(
     return Response.ok(new ResultsPage<>("datasets", datasets, totalCount)).build();
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @DELETE
+  @Path("{dataset}")
+  @Produces(APPLICATION_JSON)
+  public Response hide(
```
julienledem, line 160: for consistency I would use "delete" everywhere. Old versions are still there but the dataset is considered deleted.

api/src/main/java/marquez/api/JobResource.java
```diff
@@ -160,6 +161,25 @@ public Response list(
     return Response.ok(new ResultsPage<>("jobs", jobs, totalCount)).build();
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @DELETE
+  @Path("/namespaces/{namespace}/jobs/{job}")
+  @Produces(APPLICATION_JSON)
+  public Response hideJob(
```
julienledem, line 170: same comment: `deleteJob`?

julienledem (commented):
I have two main comments:
1) if I agree that the sql reformatting with triple quotes is good, this should be done in a separate PR. Mixing refactoring/reformatting with same behavior **and** change of behavior in the same PR makes it difficult to review. We should avoid doing that in the future.
2) personally I would prefer we call this "delete". It is a *soft* delete since all the versions are maintained but it **is** a delete as these jobs and dataset don't exist anymore. If they do they just get undeleted the next time they run or get read/written.

api/src/main/java/marquez/api/DatasetResource.java
mobuchowski: Fixed.

api/src/main/java/marquez/api/JobResource.java
mobuchowski: Fixed.

api/src/main/java/marquez/api/DatasetResource.java
```diff
@@ -150,6 +151,28 @@ public Response list(
     return Response.ok(new ResultsPage<>("datasets", datasets, totalCount)).build();
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @DELETE
+  @Path("{dataset}")
+  @Produces(APPLICATION_JSON)
+  public Response deleteDataset(
```
pawel-big-lebowski, line 160: `DatasourceResource.delete` would be enough I think.

api/src/main/java/marquez/api/DatasetResource.java
```diff
@@ -150,6 +151,28 @@ public Response list(
...
+      @PathParam("namespace") NamespaceName namespaceName,
+      @PathParam("dataset") DatasetName datasetName) {
+    throwIfNotExists(namespaceName);
+
+    Dataset dataset =
```
pawel-big-lebowski, line 165: Not sure why do we need this. Wouldn't `datasetService.delete` be enough? We don't have to return a deleted dataset in a body.

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -131,64 +132,66 @@ default void setFields(Dataset ds) {
...
+      "SELECT d.* FROM datasets_view AS d WHERE d.name = :datasetName AND d.namespace_name = :namespaceName")
   Optional<DatasetRow> findDatasetAsRow(String namespaceName, String datasetName);
 
-  @SqlQuery("SELECT * FROM datasets WHERE name = :datasetName AND namespace_name = :namespaceName")
+  @SqlQuery(
+      "SELECT * FROM datasets_view WHERE name = :datasetName AND namespace_name = :namespaceName")
   Optional<DatasetRow> getUuid(String namespaceName, String datasetName);
 
   @SqlQuery(
-      "WITH selected_datasets AS (\n"
```
pawel-big-lebowski, line 141: Unable to find out if anything changed here. Would love to have SQL refactor in separate PR.

api/src/test/java/marquez/DatasetIntegrationTest.java
```diff
@@ -357,4 +360,84 @@ public void testApp_upsertDescription() {
...
+                });
+
+    // Ensure the event was correctly rejected and a proper response code returned.
+    assertThat(resp.join()).isEqualTo(201);
+
+    client.deleteDataset(namespace, name);
+
+    List<Dataset> datasets = client.listDatasets(namespace);
+    assertThat(datasets).hasSize(0);
+    resp =
```
pawel-big-lebowski, line 428: That's the tricky part. Aren't we losing here information that dataset was deleted in the past? 
If so, perhaps we should store it in `dataset_version`.

pawel-big-lebowski: it's not a delete of dataset but a delete action triggered from Marquez-ui. So, it does not make sense to store its history.

api/src/main/java/marquez/api/DatasetResource.java
mobuchowski: There is pattern in Marquez where it returns object on which the operation was carried out. I just decided to follow it.

api/src/main/java/marquez/db/DatasetDao.java
mobuchowski: `selected_datasets` CTE uses `datasets_view` now. 

Anyway, you're right, as indicated in multiple comments, but it's already done.

api/src/main/java/marquez/api/DatasetResource.java
mobuchowski: Changed.

pawel-big-lebowski (approved):

PR #2035, opened 2022-07-14
Add @FlywayTarget annotation to migration tests to control flyway upg…

collado-mike:
### Problem
Changes in the database schema can break the Java migration scripts if all flyway migrations are applied prior to running unit tests. This allows tests to specify fixed migration versions so that no future migrations are applied for that test. Also updated backfill tests to stop relying on sql in DAOs since they change with future migrations.

### Checklist

### DIFF
--- a/api/src/main/java/marquez/db/migrations/V44_2__BackfillAirflowParentRuns.java
+++ b/api/src/main/java/marquez/db/migrations/V44_2__BackfillAirflowParentRuns.java
@@ -64,7 +64,7 @@ public class V44_2__BackfillAirflowParentRuns implements JavaMigration {
           FROM lineage_events le
           WHERE le.run_uuid=r.uuid
           AND event->'run'->'facets'->'parent'->'run'->>'runId' IS NOT NULL
-          AND event->'run'->'facets'->'airflow_version' IS NOT NULL
+          AND event->'run'->'facets'->>'airflow_version' IS NOT NULL
           ) e ON e.run_uuid=r.uuid
       WHERE e.parent_run_id IS NOT NULL
 """;
@@ -76,8 +76,8 @@ public class V44_2__BackfillAirflowParentRuns implements JavaMigration {
       current_location
       FROM jobs
       WHERE namespace_name=:namespace AND name=:jobName
-      ON CONFLICT(name, namespace_uuid) WHERE parent_job_uuid IS NULL
-      DO UPDATE SET updated_at=now()
+      ON CONFLICT (name, namespace_uuid) WHERE parent_job_uuid IS NULL
+      DO UPDATE SET updated_at=EXCLUDED.updated_at
       RETURNING uuid
       """;
   public static final String INSERT_PARENT_RUN_QUERY =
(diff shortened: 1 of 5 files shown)


api/src/main/java/marquez/db/migrations/V44_2__BackfillAirflowParentRuns.java
```diff
@@ -76,8 +76,8 @@ INSERT INTO jobs (uuid, type, created_at, updated_at, namespace_uuid, name, desc
       current_location
       FROM jobs
       WHERE namespace_name=:namespace AND name=:jobName
-      ON CONFLICT(name, namespace_uuid) WHERE parent_job_uuid IS NULL
-      DO UPDATE SET updated_at=now()
+      ON CONFLICT (name, namespace_uuid) WHERE parent_job_uuid IS NULL
+      DO UPDATE SET updated_at=EXCLUDED.updated_at
```
wslulciuc, line 80: I would assume the `updated_at` column would set to `now()` only when the metadata has changed. Do we want always to use default to `EXCLUDED.updated_at` even when metadata has changed?

api/src/test/java/marquez/db/migrations/V44_2__BackfillAirflowParentRunsTest.java
```diff
@@ -35,6 +31,7 @@
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @ExtendWith(MarquezJdbiExternalPostgresExtension.class)
+@FlywayTarget("44")
```
wslulciuc, line 34: _Minor_: Mind adding a comment as to why the test is fixed on migration `44`?

api/src/test/java/marquez/db/migrations/V44_3_BackfillJobsWithParentsTest.java
```diff
@@ -35,6 +29,7 @@
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @ExtendWith(MarquezJdbiExternalPostgresExtension.class)
+@FlywayTarget("44") // apply migrations up to and including v44 but not beyond
```
wslulciuc, line 32: _Minor_: Similar comment on commenting on fixing the migration to `44`.

api/src/main/java/marquez/db/migrations/V44_2__BackfillAirflowParentRuns.java
collado-mike: If we don't update the record, it won't return anything :(

api/src/test/java/marquez/db/migrations/V44_2__BackfillAirflowParentRunsTest.java
collado-mike: Done

api/src/test/java/marquez/db/migrations/V44_3_BackfillJobsWithParentsTest.java
collado-mike: Done

api/src/main/java/marquez/db/migrations/V44_2__BackfillAirflowParentRuns.java
wslulciuc: Very very sad....

PR #2036, opened 2022-07-14
Avoid locking common records and remove transaction annotation

collado-mike:
### Problem
The OpenLineage API implementation is unnecessarily wrapped in a `@Transaction` annotation, which means that under heavy write load, there's a lot of unnecessary contention. For example, we lock the `Namespace` record in the `upsertNamespaceRow` method, even though we don't really update the namespace. Since the whole request is wrapped in a transaction, the namespace record is locked for the entire time we write new jobs, runs, and datasets. If many jobs for the same namespace run simultaneously, each OpenLineage request will wait until the namespace record is unlocked, forming a queue of requests, rather than handling each in parallel. This also happens for `JobContext`s and `RunArg`s, as oftentimes, many jobs will run with no arguments, no source code location, and no SQL facet. Many completely unrelated jobs will block while awaiting a lock on an empty `JobContext` and/or empty `RunArgs`. 

I updated each of these `upsert` queries to do nothing on conflict (they're really immutable records) to avoid locking the impacted record.

Moreover, I don't see a reason for this API to be wrapped in a transaction at all. These changes don't need to be made atomically. OpenLineage requests are mostly idempotent, so if only partial database updates are applied and the same request made twice, there's no duplication in the data. `Job`s, `Run`s, `Dataset`s, `JobContext`s, and `RunArg`s created in the first request will be found and used in the second. Foreign key constraints require us to always write a new record before updating an existing one to reference it, so there's no opportunity to leave the database in an inconsistent state. 

This screenshot shows the database under heavy write load - note that waits are dominated by `transactionid` ([Waiting for a transaction to finish](https://www.postgresql.org/docs/current/monitoring-stats.html#WAIT-EVENT-TABLE)) and the primary contender in this case is the `NamespaceDao`.

Screen Shot 2022-07-14 at 1 41 37 PM

### Checklist

### DIFF
--- a/api/src/main/java/marquez/db/NamespaceDao.java
+++ b/api/src/main/java/marquez/db/NamespaceDao.java
@@ -78,7 +78,27 @@ public interface NamespaceDao extends BaseDao {
   @SqlQuery("SELECT * FROM namespaces ORDER BY name LIMIT :limit OFFSET :offset")
   List<Namespace> findAll(int limit, int offset);
 
-  @SqlQuery(
+  default NamespaceRow upsertNamespaceRow(
+      UUID uuid, Instant now, String name, String currentOwnerName) {
+    doUpsertNamespaceRow(uuid, now, name, currentOwnerName);
+    return findNamespaceByName(name).orElseThrow();
+  }
+
+  /**
+   * This query is executed by the OpenLineage write path, meaning namespaces are written to a LOT.
+   * Updating the record to modify the updateAt timestamp means the same namespace is often under
+   * heavy contention unnecessarily (it's really not being updated), causing some requests to wait
+   * for a lock while other requests are finishing. If a single namespace is under heavy contention,
+   * this can cause some requests to wait a long time - i.e., minutes. This causes unacceptable
+   * latency and failures in the write path. Avoid any updates in this query to avoid unnecessary
+   * locks.
+   *
+   * @param uuid
+   * @param now
+   * @param name
+   * @param currentOwnerName
+   */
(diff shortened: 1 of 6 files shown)


api/src/main/java/marquez/db/NamespaceDao.java
```diff
@@ -78,7 +78,27 @@ default Namespace upsertNamespaceMeta(
   @SqlQuery("SELECT * FROM namespaces ORDER BY name LIMIT :limit OFFSET :offset")
   List<Namespace> findAll(int limit, int offset);
 
-  @SqlQuery(
+  default NamespaceRow upsertNamespaceRow(
+      UUID uuid, Instant now, String name, String currentOwnerName) {
+    doUpsertNamespaceRow(uuid, now, name, currentOwnerName);
+    return findNamespaceByName(name).orElseThrow();
```
wslulciuc, line 84: _Minor_: The `RETURNING *` was to avoid inserting, then querying for the namespace in two separate queries. But, also understand the cost of always updating `updated_at` to trigger the modified row to be returned. At some point, I'd love to flush out the usage of `updated_at` for all of our tables.

wslulciuc (approved):
Great analysis and write up @collado-mike on the DB performance! Who knew such "minor" changes could improve query performance so much. As for transactions, I agree. Originally, we wanted to keep the write path for Marquez in sync with metadata collected via OL. Now, I do think we'll have to wrap _some_ DB calls in a transaction, but let's add them as we encounter data inconsistencies.

Issue #1946, opened 2022-04-11
Support Job renaming and symlinking


collado-mike:
In the design for https://github.com/MarquezProject/marquez/issues/1928 , it is proposed that extra logic be added to Marquez to redirect job URLs that use the job's original simple name to instead point to the job's Fully Qualified Name (FQN). In the current design, a view is used to return the FQN of an existing job as the `name` field of `Job` records, minimizing the SQL changes needed to construct/return the FQN while avoiding the high cost of backfilling the `jobs`, `runs`, `lineage_events`, and `job_version` tables, all of which refer to a job by name.

As an alternative to using a view, we also have a need to support job renaming/symlinking. This is useful for a host of other reasons, including

* changing the namespace of a job because team ownership has changed
* adding a task group to an existing Airflow DAG, which changes the FQN of a job
* physical plan changes in Spark, which could change the generated job name

By supporting Job symlinks, a simpler backfill script for #1928 could insert new jobs with the fully qualified name defined and update the existing job to symlink to the new one. Old `runs` and `lineage_events` records can continue to point to the existing `jobs` record without a backfill and logic in Marquez would be responsible for merging the jobs. 

It's worth noting that _every_ API in Marquez that refers to a job would need to implement the logic to find and process all symlinks for a queried Job and merge all results. For example, the `/lineage` API would need to determine all of the jobs linked to every job present in the graph and to then merge all subsequent graphs into a single graph with symlinked jobs deduped. The `/runs` API would need to find the latest runs for any job symlinked to the requested one and the runs would need to be merged so that they are in time order regardless of which job name a given run points to.

Symlinks could be recursive and can be many-to-one, so that navigating symlinks would be modeled as a tree-traversal of arbitrary depth (practically probably one or two). ~~This would be modeled in the Marquez API code, as modifying all existing SQL to capture everything is likely impossible.~~ (untrue - see below comment)

collado-mike:
Practically, we can probably make the simplifying assumption that a job symlink path is linear rather than a tree. If we make the job symlink relationship 1:1 rather than many-to-one, we can modify the existing lineage query to something like the following:

```sql
          WITH RECURSIVE
              job_symlink AS (
                SELECT uuid, current_version_uuid, symlink_target
                FROM jobs j
                UNION
                SELECT j.uuid,
                COALESCE(j.current_version_uuid, js.current_version_uuid) AS current_version_uuid,
                j.symlink_target
                FROM jobs j
[...]

PR #2043, opened 2022-07-20
Fix recursive views perf

collado-mike:
### Problem
In https://github.com/MarquezProject/marquez/issues/1928, we introduced a `jobs_view`, which used recursive queries to construct a job's fully qualified name by collecting the job's ancestry and joining the names. Additionally, https://github.com/MarquezProject/marquez/issues/1946 introduced job symlinks, which also used recursive queries to find the last link in a chain of symlinks and always return the ultimate target job. Unfortunately, while initial testing showed those queries performed reasonably well on _read_ requests, we didn't sufficiently load test the OpenLineage _write_ API. Unfortunately, under sustained load in a real production environment, the write API (which repeatedly queries the `jobs_view` either directly or via the `runs_view`) puts considerable strain on our Postgres database. 

### Solution
One of the goals with using a view was to consistently construct job FQNs even when ancestors are renamed (using the symlink feature). This allows us to avoid big migration tasks updating the `runs` table with new job names when, e.g., an Airflow DAG or a Spark application is renamed, thus renaming all of its child tasks. However, constructing the FQN on every read is too computationally expensive, so this PR changes the view to compute the FQN on _write_. This continues to use the view, but now there is a `TRIGGER` on `INSERT` to compute the FQN for a job, as well as to traverse the symlink chain and write the new FQN and symlink target to a new table called `jobs_fqn`. This table also collects the past aliases of a job, so we can serve requests that still refer to the old job name, as described in the original [symlink issue](https://github.com/MarquezProject/marquez/issues/1946). 

I utilized writes to the `jobs_view` in order to capture the write by the trigger, so the code now references the view instead of the original `jobs` table. This also allowed me to return the full row from the insert query rather than just the job UUID, as it previously did. I also simplified the uniqueness constraint by adding a `parent_job_id_string` field, which just converts the parent job id to a string or an empty string if null (we couldn't use a null UUID to enforce uniqueness for jobs that have no parents because `NULL` != `NULL` in Postgres).

Given the `jobs_view` and associated write function are likely to change in the future, I made their definitions [repeatable migrations](https://flywaydb.org/documentation/concepts/migrations#repeatable-migrations) so that future changes can be captured in version control (I went ahead and did the same for the `runs_view` for the same reason, though unrelated to the purpose of this PR).

I also included an additional test for the DAG renaming use case.

The following snapshot shows performance in a load test both before and after the proposed code change. Both tests sustained a rate of 900 requests per minute to the OpenLineage write API. I used the same RDS [db.m5.large](https://aws.amazon.com/rds/instance-types/) instance for both tests (cleared the DB structure between tests).

(note that https://github.com/MarquezProject/marquez/pull/2041 is included in the test deployment)

Screen Shot 2022-07-20 at 12 08 43 PM

With the current code at HEAD, the DB's CPU utilization climbs to 95%+ within about an hour, then hovers just below 100%. The request graph shows the request rate dropping after that because pods are consistently being restarted due to failed health checks (can't access the database), so eventually only about 50% of requests are being handled.

Conversely, the DB CPU utilization never rises about ~50% with the new code changes. The request rate stays pretty steady throughout the test (there are still some OOM restarts, but that's consistent with what we see in production with an older version of Marquez). 

### Checklist


api/src/main/resources/marquez/db/migration/R__1_jobs_view_and_rewrite_function.sql
```diff
@@ -0,0 +1,127 @@
...
+       f.aliases
+FROM jobs_fqn f,
+     jobs j
+WHERE j.uuid = f.uuid;
+
+
+CREATE OR REPLACE FUNCTION rewrite_jobs_fqn_table() RETURNS TRIGGER AS
+$$
+DECLARE
+    id uuid;
```
wslulciuc, line 29: _Minor_: For readability, you could us:

```
job_uuid
job_inserted
```

for the function declaration.

api/src/main/resources/marquez/db/migration/R__1_jobs_view_and_rewrite_function.sql
```diff
@@ -0,0 +1,127 @@
+CREATE OR REPLACE VIEW jobs_view
```
wslulciuc, line 1: _Minor_: I would be more descriptive with the description for the repeatable migration:

> **R__1_add_jobs_view_and_rewrite_jobs_fqn_table_function.sql**

api/src/main/resources/marquez/db/migration/R__2_runs_view.sql
```diff
@@ -0,0 +1,24 @@
+CREATE OR REPLACE VIEW runs_view
```
wslulciuc, line 1: _Minor_: (Similar comment as above) I would be more descriptive with the description and go with:

> **R__1_add_runs_view.sql**

wslulciuc (approved):
I left a few minors comments, but overall the changes seem reasonable (I double checked the SQL 😉). Thanks for the detailed writeup (as usual), @collado-mike! Also, since the DB _write_ performance issues were observed only after deploying to production, having a not too exhaustive load test in CI with a report can help spot any DB bottlenecks sooner. I've opened https://github.com/MarquezProject/marquez/issues/2047

api/src/main/resources/marquez/db/migration/R__1_jobs_view_and_rewrite_function.sql
collado-mike: I don't know... after the first time it runs, it's not really _adding_ anything anymore. I kind of like treating the repeatable migration file just like a code file - you'd never use `CreateJobDao.java` as the file name.
Now, I think one legitimate criticism might be whether the `rewrite_function` and the `jobs_view` ought to be in the same migration file (if we're thinking of the file like code). But, IMO, they're so intrinsically connected it made sense to keep them together.

wslulciuc: I think that works, I'd like to document how we define migrations (or repeatable ones) for new contributors. So, as long as we have a consistent naming convention, and it's documented then we can leave the name of the file as is (and looks like `R__People_view.sql` is used in Flyways docs on [repeatable migrations](https://flywaydb.org/documentation/tutorials/repeatable#creating-a-repeatable-migration), which is more aligned with your naming approach)

collado-mike: Ok, I'll update the files to follow that naming convention and add a line to the `CONTRIBUTING.md` file

PR #2050, opened 2022-07-28
docs: update Airflow example for 2.3

mobuchowski:
Docs referenced 1.10 method of integration, that is usually no longer relevant.

Also, added few minor enhancements that slightly boost clarity IMO.


wslulciuc (approved):
Thanks, @mobuchowski!

PR #2055, opened 2022-08-02
Update FlywayFactory to support custom schema

collado-mike:
### Problem

There's currently no way to customize the marquez database to use a custom schema. This updates the FlywayFactory to support an argument to customize the schema programatically. This doesn't aim to support custom schemas from configuration.

### Checklist


api/src/main/java/marquez/db/FlywayFactory.java
```diff
@@ -74,6 +76,15 @@ public final class FlywayFactory {
   @Getter @Setter
   private String repeatableSqlMigrationPrefix = DEFAULT_REPEATABLE_SQL_MIGRATION_PREFIX;
 
+  @Getter private Optional<String> schema = Optional.empty();
+
+  public void setSchema(@NonNull String schema) {
+    if (StringUtils.isEmpty(schema)) {
```
wslulciuc, line 82: Under `marquez.common.base` we define:

```java
MorePreconditions.checkNotBlank()
```

I would use:

```java
checkNotBlank(schema, "schema must not be blank");
```

api/src/main/java/marquez/db/FlywayFactory.java
```diff
@@ -74,6 +76,15 @@ public final class FlywayFactory {
   @Getter @Setter
   private String repeatableSqlMigrationPrefix = DEFAULT_REPEATABLE_SQL_MIGRATION_PREFIX;
 
+  @Getter private Optional<String> schema = Optional.empty();
```
wslulciuc, line 79: In Marquez, we don't use `Optional`s for replacements of `null`s, also the `Optional` class is not serializable (see [How to use Optionals In Java](https://dzone.com/articles/optional-in-java)). What's common throughout the codebase is to define a getter method as follows:

```java
public Optional< String > getSchema() {
  return Optional.ofNullable(schema);
}
```

api/src/main/java/marquez/db/FlywayFactory.java
```diff
@@ -102,6 +113,7 @@ public Flyway build(@NonNull DataSource source) {
         .placeholderSuffix(placeholderSuffix)
         .sqlMigrationPrefix(sqlMigrationPrefix)
         .repeatableSqlMigrationPrefix(repeatableSqlMigrationPrefix)
+        .schemas(schema.map(s -> new String[] {s}).orElse(new String[] {}))
```
wslulciuc, line 116: Flyway does expose a `defaultSchema`:

> The default schema managed by Flyway. This schema name is case-sensitive. If not specified, but schemas is, Flyway uses the first schema in that list. If that is also not specified, Flyway uses the default schema for the database connection.
Consequences:
> * This schema will be the one containing the schema history table.
> * This schema will be the default for the database connection (provided the database supports this concept).

Would a better option be to set the `defaultSchema`? Right now, we default to the schema defined in the database connection.

api/src/main/java/marquez/db/FlywayFactory.java
collado-mike: I think those guidelines are sensible when dealing with service models that are serialized and contain properties that may be accessed by other parts of the code. This field, however, is entirely internal and never accessed by any other object. We don't access any other fields via public getter method (e.g., there's no `getRepeatableSqlMigrationPrefix()` in the previous line). So it doesn't seem consistent or useful to use a getter here.

I could simply default the field to `null` and replace the `Optional` call with `schema == null ? new String[]{} : new String[]{schema}`, but I think `Optional` is preferable here. 

I could use `Optional<String> schemaOption = Optional.ofNullable(schema)`, but that's really just doing `schema == null ? Optional.empty() : Optional.of(schema)`, which is not really different from the previous option. 

Of the options, I think using the `Optional.empty()` field as a default is the cleanest. It's not really different from using `Optional` as a replacement for `null` inside a method body, which we do in Marquez all the time. There's no serialization concern, no external client that needs to access the state of this object...

api/src/main/java/marquez/db/FlywayFactory.java
collado-mike: I think I avoided that because I don't like passing `null` as a method parameter, but I suppose that's what Flyway does internally anyway...

collado-mike: Though, I suppose if we're just going to pass in `null`, there's no real point in using `Optional` at all. I just find it distasteful to pass around `null` when better options exist. But, I suppose no point in making things more complicated than they need to be

api/src/main/java/marquez/db/FlywayFactory.java
wslulciuc: That's a fair distinction.

wslulciuc (approved):
LGTM 👍

PR #2067, opened 2022-08-10
Fix/rewrite jobs fqn locks

collado-mike:
### Problem

The `rewrite_jobs_fqn_table` function is inadvertently updating jobs even when no metadata about the job has changed. Under load, this causes significant locking issues, as the `jobs_fqn` table must be locked for every job update. 

### Solution

I updated the function to only update the table if the job is a new record or if the `symlink_target_uuid` is distinct from the previous value. These graphs show the number of blocked transactions and commits over time before and after the change. After applying the change (around 20:10 on these graphs), the number of commits goes way up and the number of active/blocked transactions drops.

Screen Shot 2022-08-10 at 12 08 17 PM

I also added a test assertion to validate the update of the `jobs_fqn` table.

### Checklist

### DIFF
--- a/api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql
+++ b/api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql
@@ -27,6 +27,8 @@ CREATE OR REPLACE FUNCTION rewrite_jobs_fqn_table() RETURNS TRIGGER AS
 $$
 DECLARE
     job_uuid uuid;
+    new_symlink_target_uuid uuid;
+    old_symlink_target_uuid uuid;
     inserted_job jobs_view%rowtype;
 BEGIN
     INSERT INTO jobs (uuid, type, created_at, updated_at, namespace_uuid, name, description,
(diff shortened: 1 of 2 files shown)


api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql
```diff
@@ -55,13 +57,19 @@ BEGIN
...
                       current_location         = EXCLUDED.current_location,
                       current_inputs           = EXCLUDED.current_inputs,
-                      -- update the symlink target if not null. otherwise, keep the old value
-                      symlink_target_uuid      = COALESCE(
-                              EXCLUDED.symlink_target_uuid,
-                              jobs.symlink_target_uuid)
-    RETURNING uuid INTO job_uuid;
-    IF TG_OP = 'INSERT' OR
-       (TG_OP = 'UPDATE' AND OLD.symlink_target_uuid IS DISTINCT FROM NEW.symlink_target_uuid) THEN
+                      -- update the symlink target if null. otherwise, keep the old value
```
wslulciuc, line 60: Ok, the comments here are much appreciated ❤️

wslulciuc (approved):
@collado-mike thanks for the fix, and the write up and tests that went along with the change 💯 🥇

PR #2068, opened 2022-08-10
Update lineage query to only look at jobs with inputs or outputs

collado-mike:
Signed-off-by: Michael Collado <collado.mike@gmail.com>

### Problem

In many environments a large number of jobs reporting events have no inputs or outputs - e.g., PythonOperators in an Airflow deployment. If a Marquez installation has a lot of these, the lineage query spends a lot of its time searching for overlaps with jobs that have no inputs or outputs. In one installation, we have > 200K jobs, but only ~7000 jobs that have any inputs or outputs at all.

### Solution

This changes the lineage query to query the `job_versions_io_mapping` table and _INNER_ join with the `jobs_view` so that only jobs that have inputs or outputs are present in the `jobs_io` CTE. The impact of this is that table becomes very small and the recursive join in the `lineage` CTE is very fast. 

Probably notable that the missing inputs/outputs are largely due to insufficient coverage by the OpenLineage integrations - e.g., those PythonOperators are likely reading data from somewhere. This is, at best, a short term fix until OL coverage increases, at which point, the query will have to be revisited again.

### Checklist

### DIFF
--- a/api/src/main/java/marquez/db/LineageDao.java
+++ b/api/src/main/java/marquez/db/LineageDao.java
@@ -37,35 +37,39 @@ public interface LineageDao {
    * @return
    */
   @SqlQuery(
-      // dataset_ids: all the input and output datasets of the current version of the specified jobs
-      "WITH RECURSIVE\n"
-          + "    job_io AS (\n"
-          + "        SELECT j.uuid AS job_uuid,\n"
-          + "        ARRAY_AGG(DISTINCT io.dataset_uuid) FILTER (WHERE io_type='INPUT') AS inputs,\n"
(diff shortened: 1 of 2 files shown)


wslulciuc (approved):
@collado-mike, though you consider this a quick "hack", I do view it as reasonable optimization to improve the lineage query performance. Now, as you pointed out, it is a short-term optimization change. That is, coverage for OpenLineage will improve, which, in turn, will negatively impact the lineage query. But we have spoken briefly on ways we can continue to improve  lineage query performance by introducing materialized views, caching, etc. Anyways, great work 💯 🥇

Issue #1927, opened 2022-03-29
Option to dump OpenLineage events that correspond to dataset/namespace from web


mobuchowski:
Having this feature would make debugging, and replicating errors much faster.

wslulciuc:
I would add a `lineage export` cmd to the [`marquez-cli`](https://github.com/MarquezProject/marquez/tree/main/api/src/main/java/marquez/cli) that would export the OL events from the `lineage_events` table. For example:

```
$ java -jar marquez.jar lineage export > lineage.json
```

howardyoo:
It would also be nice if Marquez UI would have a little button at the bottom that, when pressed, will reveal a console panel that outputs all the raw OL events that it received - acting sort of like a kind of debug console. Typically, users would use marquez to visualize the events (after they get stored in its backend DB) - but also may want to monitor how the OL messages are actually being received.

rossturk:
Perhaps there could be a sortable/filterable/paginated table in the UI for lineage events.

There have been many times where I just wanted to see "what got emitted" or "did my pipeline do it right?" It would be an excellent debugging tool, and could help people build OL integrations more quickly.

Does an API exist that could support this? If not, one could possibly serve both a) data export and b) debugging use cases.

mobuchowski:
>Does an API exist that could support this? If not, one could possibly serve both a) data export and b) debugging use cases.

@wslulciuc @howardyoo @rossturk Do we want to 

1. display all events, perhaps sorted by time descending to see latest events?
2. display all events, but choose some particular namespace on which we're looking at?
3. display events that look at particular job or dataset only?

howardyoo:
> > Does an API exist that could support this? If not, one could possibly serve both a) data export and b) debugging use cases.
> 
> @wslulciuc @howardyoo @rossturk Do we want to
> 
> 1. display all events, perhaps sorted by time descending to see latest events?
> 2. display all events, but choose some particular namespace on which we're looking at?
> 3. display events that look at particular job or dataset only?
> 
> I think the API would look differently depending on decision here. Of course, the first option is the simplest.

1. I would want the events sorted by time descending to see latest events - that would typically be the way users would skim through the dumps
2. filtering based on namespace could be an option, if possible, and certainly would be very helpful.
3. filtering based on particular job or dataset... it may be good to have, but I don't think it is a must have.

Rather, my opinion would be to be able to filter more on even types (like COMPLETE, FAIL, etc) that may be more useful, or based on particular time period.

[...]

Issue #1973, opened 2022-04-28
Marquez dataset & job notifications' panel


pawel-big-lebowski:
Several things can happen within a dataset or a job: 
- a new column has been added
- dataset content was truncated 3 days ago
- job is failing since column renamed

As a data engineer I can think of plenty of such scenarios. We do have in Marqeuz all the information to detect it but it is not that easy to visualize them. As a a data engineer, with a limited web UI skills, it is hard for me to make use of this information although I know how valuable it is. 

In each dataset/job view we could have something like `Notifications` where such information can be presented. This would allow other Marquez contributors to implement their own notifications or alerts.

wslulciuc:
@pawel-big-lebowski Love this idea! Added it to our roadmap. I'd be great to get a proposal written outlining how the web UI would display the notifications and any additional API changes needed to support aggregating the notification events in order to provide this level of transparency (similar to what you've outlined above). Would you be interested in writing the proposal?

PR #2070, opened 2022-08-12
add raw OpenLineage get event API

mobuchowski:
Adds an API that returns raw OpenLineage events sorted by time, and optionally filtered by namespace.
Filtering by namespace takes into account both job and dataset namespaces.

Closes: https://github.com/MarquezProject/marquez/issues/1927 


api/src/main/java/marquez/api/EventResource.java
```diff
@@ -0,0 +1,62 @@
...
+public class EventResource extends BaseResource {
+  public EventResource(@NonNull final ServiceFactory serviceFactory) {
+    super(serviceFactory);
+  }
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Path("/events")
```
wslulciuc, line 33: I would use the endpoint `GET` `/lineage_events` to be more specific, and less generic. We'll also be adding a `GET` `/lifecycle_events` endpoint that will capture events like:

* A **new** dataset / job version has created
* A dataset has been **deleted**
* A job has transited to `COMPLETED` at time **X**

See https://github.com/MarquezProject/marquez/issues/1973

mobuchowski: @wslulciuc maybe `/events/lineage` and `/events/lifecycle`? Not sure which is less confusing between `lineage` and `lifecycle`.

api/src/main/java/marquez/db/mappers/RawLineageEventMapper.java
```diff
@@ -0,0 +1,51 @@
...
+
+    List<Object> inputs = Collections.emptyList();
+    List<Object> outputs = Collections.emptyList();
+    try {
+      ObjectMapper mapper = Utils.getMapper();
+      Map<String, Object> event = mapper.readValue(rawEvent, Map.class);
+      run = (Map<String, Object>) event.getOrDefault("run", Collections.emptyMap());
+      job = (Map<String, Object>) event.getOrDefault("job", Collections.emptyMap());
+      inputs = (List<Object>) event.getOrDefault("inputs", Collections.emptyList());
+      outputs = (List<Object>) event.getOrDefault("outputs", Collections.emptyList());
```
julienledem, line 36: We could also use JSONObject and return that instead of RawLineageEvent in the REST endpoint?
I don't think we do much with the fields we extract here.

julienledem: additionally this could hide user added non-expected fields in the root JSON

julienledem (commented):
This looks good to me, I would avoid the unnecessary parsing, unparsing of json on the server side to the RawLineageEvent structure and just use Jackson's JSONObject there.

mobuchowski:
@julienledem fixed - deserializing to JsonNode. Might also just return String...

Added tests.

api/src/main/java/marquez/db/mappers/RawLineageEventMapper.java
mobuchowski: Fixed that before, forgot to notify.

api/src/main/java/marquez/api/EventResource.java
wslulciuc: @mobuchowski: actually, my initial thought was to use (and the more I think about it, we should):

```
GET /events/lineage
GET /events/lifecycle
```

I can't recall why I convinced myself `/lifecycle_events` was a better alternative. That said, I think using `/events/*` provides callers with a more intuitive API to query events by type; we could also define `GET` `/events` with a `type` param to include events of a particular type. But, the response objects differ enough where it would warrant separate endpoints as you suggested 👍

api/src/main/java/marquez/api/EventResource.java
```diff
@@ -0,0 +1,62 @@
...
+      @QueryParam("offset") @DefaultValue("0") @Min(value = 0) int offset) {
+    final List<JsonNode> events = eventService.getAll(limit, offset);
+    return Response.ok(new Events(events)).build();
+  }
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Path("/events/{namespace}")
```
wslulciuc, line 46: To get events for a given namespace, I would use:

```
GET /namespaces/{namespace}/events
```

We could also define a `namespace` query param, but having the namespace in the path aligns more with how endpoints are defined for Marquez.

api/src/main/java/marquez/api/EventResource.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+import lombok.NonNull;
+import lombok.Value;
+import marquez.service.ServiceFactory;
+
+@Path("/api/v1")
+public class EventResource extends BaseResource {
```
wslulciuc, line 24: Can we define these endpoints in [`OpenLineageResource`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/OpenLineageResource.java)?

api/src/main/java/marquez/api/EventResource.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Path("/events")
+  @Produces(APPLICATION_JSON)
+  public Response get(
+      @QueryParam("limit") @DefaultValue("100") @Min(value = 0) int limit,
```
wslulciuc, line 36: I would also allow callers

* `before`: Return events created _before_ a particular date (as `YYYY-MM-DD`)
* `after`: Return events created _after_ a particular date (as `YYYY-MM-DD`
* `sort`:  Sort events in `asc` or `desc` order by `event_time`. Default is `desc`.

api/src/main/java/marquez/api/EventResource.java
```diff
@@ -0,0 +1,62 @@
...
+      @QueryParam("offset") @DefaultValue("0") @Min(value = 0) int offset) {
+    final List<JsonNode> event = eventService.getByNamespace(namespace, limit, offset);
+    return Response.ok(new Events(event)).build();
+  }
+
+  @Value
+  static class Events {
+    @NonNull
+    @JsonProperty("events")
```
wslulciuc, line 59: ❤️

api/src/main/java/marquez/db/EventDao.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.List;
+import marquez.db.mappers.RawLineageEventMapper;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+@RegisterRowMapper(RawLineageEventMapper.class)
+public interface EventDao extends BaseDao {
```
wslulciuc, line 11: Can we query for events using [`OpenLineageDao`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/OpenLineageDao.java)?

```
OpenLineageDao.findAllEvents()
OpenLineageDao.findAllEventsByNamespace()
```

`OpenLineageDao` does need some refactoring, but given that we are interacting with lineage events, it would make sense to reuse the class vs defining a separate DAO.

wslulciuc (dismissed):
I would also update our [`CHANGELOG.md`](https://github.com/MarquezProject/marquez/blob/main/CHANGELOG.md)  with this sweet sweet endpoint!

api/src/main/java/marquez/api/EventResource.java
```diff
@@ -0,0 +1,62 @@
...
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Path("/events")
+  @Produces(APPLICATION_JSON)
+  public Response get(
+      @QueryParam("limit") @DefaultValue("100") @Min(value = 0) int limit,
+      @QueryParam("offset") @DefaultValue("0") @Min(value = 0) int offset) {
+    final List<JsonNode> events = eventService.getAll(limit, offset);
```
wslulciuc, line 38: I would just call the DAO directly here: 

```
OpenLineageDao.findAllEvents()
```

api/src/main/java/marquez/service/EventService.java
```diff
@@ -0,0 +1,10 @@
+package marquez.service;
+
+import lombok.NonNull;
+import marquez.db.BaseDao;
+
+public class EventService extends DelegatingDaos.DelegatingEventDao {
```
wslulciuc, line 6: Though the Marquez codebase uses services throughout (thanks for following our convention), we are moving away from defining services as delegating objects in favor of using the DAOs directly if no business logic is defined. You can remove this call all together to simplify things! ;)

api/src/main/java/marquez/db/EventDao.java
```diff
@@ -0,0 +1,37 @@
+package marquez.db;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.List;
+import marquez.db.mappers.RawLineageEventMapper;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+@RegisterRowMapper(RawLineageEventMapper.class)
```
wslulciuc, line 10: Have you looked into using the [`@Json`](https://jdbi.org/#_operation) instead of defining a custom mapper?

> Our PostgresPlugin provides qualified factories that will bind/map the @EncodedJson String to/from json or jsonb-typed columns.

api/src/test/java/marquez/jdbi/MarquezJdbiExternalPostgresExtension.java
```diff
@@ -34,6 +35,7 @@ public class MarquezJdbiExternalPostgresExtension extends JdbiExternalPostgresEx
     database = POSTGRES.getDatabaseName();
     plugins.add(new SqlObjectPlugin());
     plugins.add(new PostgresPlugin());
+    plugins.add(new Jackson2Plugin());
```
wslulciuc, line 38: Do we also need to add `Jackson2Plugin` to [`MarquezApp`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/MarquezApp.java#L153)?

clients/java/src/main/java/marquez/client/MarquezClient.java
```diff
@@ -87,6 +88,24 @@ public MarquezClient(final URL baseUrl, @Nullable final String apiKey) {
     this.http = http;
   }
 
+  public List<RawLineageEvent> listEvents() {
```
wslulciuc, line 91: As discussed [here](https://github.com/MarquezProject/marquez/pull/2070#discussion_r950534505), we'll want to define this method as:

```
MarquezClient.listLineageEvents()
```

clients/java/src/main/java/marquez/client/models/RawLineageEvent.java
```diff
@@ -0,0 +1,22 @@
...
+
+package marquez.client.models;
+
+import java.time.ZonedDateTime;
+import java.util.List;
+import java.util.Map;
+import lombok.Value;
+
+@Value
+public class RawLineageEvent {
```
wslulciuc, line 14: _Minor_: Maybe just `LineageEvent`?

clients/java/src/main/java/marquez/client/models/RawLineageEvent.java
```diff
@@ -0,0 +1,22 @@
...
+  String eventType;
+  ZonedDateTime eventTime;
+  Map<String, Object> run;
+  Map<String, Object> job;
+  List<Object> inputs;
+  List<Object> outputs;
+  String producer;
```
wslulciuc, line 21: Use the `URI` type since `producer` must be a valid URI?

api/src/main/java/marquez/api/EventResource.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Path("/events")
+  @Produces(APPLICATION_JSON)
+  public Response get(
+      @QueryParam("limit") @DefaultValue("100") @Min(value = 0) int limit,
+      @QueryParam("offset") @DefaultValue("0") @Min(value = 0) int offset) {
```
julienledem, line 37: What is the order here? Where does offset start from?

api/src/main/resources/marquez/db/migration/V47__add_lineage_event_indexes.sql
```diff
@@ -0,0 +1,5 @@
+CREATE INDEX lineage_events_event_time
+    on lineage_events(event_time DESC);
+
+CREATE INDEX lineage_events_namespace_event_time
+    on lineage_events(job_namespace, event_time DESC);
```
julienledem, line 5: Maybe do some perf test on a big db to measure the impact of this on POST /api/v1/lineage

spec/openapi.yml
```diff
@@ -70,6 +70,44 @@ paths:
...
+
+  /events/{namespace}:
+    parameters:
+      - $ref: '#/components/parameters/namespace'
+    get:
+      operationId: getRawEvents
+      parameters:
+        - $ref: '#/components/parameters/limit'
+        - $ref: '#/components/parameters/offset'
+      summary: List all received OpenLineage events in particular namespace.
```
julienledem, line 99: can you clarify the ordering? Offset might be expensive from a query perspective

api/src/main/java/marquez/db/EventDao.java
```diff
@@ -0,0 +1,37 @@
...
+    SELECT le.event
+    FROM lineage_events le, jsonb_array_elements(coalesce(le.event -> 'inputs', '[]'::jsonb) || coalesce(le.event -> 'outputs', '[]'::jsonb)) AS ds
+    WHERE le.job_namespace = :namespace
+    OR ds ->> 'namespace' =  :namespace
+    ORDER BY event_time DESC
+    LIMIT :limit
+    OFFSET :offset""")
+  List<JsonNode> getByNamespace(
+      @Bind("namespace") String namespace, @Bind("limit") int limit, @Bind("offset") int offset);
+}
```
julienledem, line 37: This looks like it would be expensive on a big table. I would check the perf difference of filtering on the columns you added indices for vs this.

api/src/main/java/marquez/db/EventDao.java
```diff
@@ -0,0 +1,37 @@
...
+@RegisterRowMapper(RawLineageEventMapper.class)
+public interface EventDao extends BaseDao {
+
+  @SqlQuery(
+      """
+    SELECT event
+    FROM lineage_events
+    ORDER BY event_time DESC
+    LIMIT :limit
+    OFFSET :offset""")
```
julienledem, line 19: Queries using offset will be slow for large values of offset even with small values of limit. This is because the db needs to count the rows leading to the offset you want. If you index by timestamp it might be better to use that. "timestamp > t limit 100" then to get the next 100, the client can provide the timestamp of the last event they retrieved.

julienledem (commented):
I added some comments. I think this needs to be perf tested before merging

api/src/main/java/marquez/db/EventDao.java
mobuchowski: With @wslulciuc changes described in https://github.com/MarquezProject/marquez/pull/2076 the json based filter would be replaced with fast join. 

It would be possible to [create jsonb index](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING) - however, I think it would be overkill.

api/src/main/java/marquez/service/EventService.java
mobuchowski: Done.

api/src/main/java/marquez/api/EventResource.java
mobuchowski: Done.

api/src/main/java/marquez/db/EventDao.java
mobuchowski: Replaced with `LineageEventMapper` that does what I want it to.

api/src/test/java/marquez/jdbi/MarquezJdbiExternalPostgresExtension.java
mobuchowski: Done.

clients/java/src/main/java/marquez/client/models/RawLineageEvent.java
mobuchowski: Done.

clients/java/src/main/java/marquez/client/models/RawLineageEvent.java
mobuchowski: Done.

clients/java/src/main/java/marquez/client/MarquezClient.java
mobuchowski: Done.

api/src/main/java/marquez/db/EventDao.java
mobuchowski: Done.

api/src/main/java/marquez/db/EventDao.java
mobuchowski: I've added `before` and `after` params. They should, combined with an index, provide efficient pagination.

api/src/main/java/marquez/api/EventResource.java
mobuchowski: Removed offset.

api/src/main/java/marquez/api/EventResource.java
mobuchowski: @wslulciuc `sort` - to efficiently do this, this operation would require two separate indexes.

api/src/main/resources/marquez/db/migration/V47__add_lineage_event_indexes.sql
mobuchowski: @wslulciuc @julienledem btw, we probably should do `CREATE INDEX CONCURRENTLY`

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -80,10 +91,48 @@ private int determineStatusCode(Throwable e) {
...
   }
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Path("/events/lineage")
+  @Produces(APPLICATION_JSON)
+  public Response getLineageEvents(
+      @QueryParam("before") @DefaultValue("2030-01-01T00:00:00+00:00") ZonedDateTimeParam before,
```
pawel-big-lebowski, line 108: would be great if there was a way to make it `now`

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -80,10 +91,48 @@ private int determineStatusCode(Throwable e) {
...
+    final List<LineageEvent> events =
+        openLineageDao.getAllLineageEvents(before.get(), after.get(), limit);
+    return Response.ok(new Events(events)).build();
+  }
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Path("/namespace/{namespace}/events/lineage")
```
pawel-big-lebowski, line 120: if it is the same resource, but filtered, shouldn't this look like:
```
/events/lineage?namespace=some-namespace
```

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -86,7 +88,37 @@ void createLineageEvent(
...
+  List<LineageEvent> getAllLineageEvents(ZonedDateTime before, ZonedDateTime after, int limit);
+
+  /**
+   * This is a "hack" to get inputs/outputs namespace from jsonb column: <a
+   * href="https://github.com/jdbi/jdbi/issues/1510#issuecomment-485423083">explanation</a>
+   */
+  @SqlQuery(
+      """
+  SELECT le.event
+  FROM lineage_events le, jsonb_array_elements(coalesce(le.event -> 'inputs', '[]'::jsonb) || coalesce(le.event -> 'outputs', '[]'::jsonb)) AS ds
```
pawel-big-lebowski, line 110: Is the  `jsonb` required? It kills the idea of `namespace` index and is kind of corner-case scenario.

mobuchowski: I think the idea of getting events from some dataset namespace is not a corner case scenario, but very common one. For example, getting all events that relate to particular Postgres database.

mobuchowski: Also, this would be fast in the future, after @wslulciuc split of facets. Then we could just filter by that AFAIK.

api/src/main/java/marquez/api/OpenLineageResource.java
mobuchowski: Unfortunately, I don't think this is possible: https://www.baeldung.com/java-annotation-attribute-value-restrictions

Also, `now` could potentially miss events if the Marquez server's clock is lagging - when the event time of some events is in the future.

api/src/main/java/marquez/api/OpenLineageResource.java
mobuchowski: @wslulciuc suggested that it would be not consistent with how other endpoint are defined in Marquez, which I agree with: https://github.com/MarquezProject/marquez/pull/2070#discussion_r964417166

api/src/main/java/marquez/db/OpenLineageDao.java
pawel-big-lebowski: Ok. If so, the index `lineage_events(job_namespace, event_time DESC);` will not help and can be removed, isn't it?

api/src/main/java/marquez/api/OpenLineageResource.java
pawel-big-lebowski: ok, let's stay consistent

api/src/main/java/marquez/db/OpenLineageDao.java
mobuchowski: FYI @julienledem @wslulciuc - this query won't work, since it requires full `lineage_events` table scan on every call, together with this `jsonb` unnesting.

I've talked with @pawel-big-lebowski and we tried to find faster query. The best we've come to is this:

```sql
WITH job_events AS (
    SELECT le.event
    FROM lineage_events le
    WHERE le.job_namespace = :namespace
    AND (le.event_time < :before
    AND le.event_time >= :after)
    ORDER BY le.event_time DESC
), dataset_events AS (
    SELECT le.event, le.event_time
    FROM lineage_events le
    JOIN dataset_versions dv on le.run_uuid = dv.run_uuid
    JOIN datasets_view ds on dv.dataset_uuid = ds.uuid
    JOIN namespaces n on ds.namespace_uuid = n.uuid
    WHERE n.name = :namespace
    AND (le.event_time < :before
    AND le.event_time >= :after)
    ORDER BY le.event_time DESC
)
SELECT le.event
FROM (
    SELECT * FROM dataset_events
    UNION ALL
    SELECT * FROM job_events
) le
ORDER BY le.event_time
LIMIT :limit
```

It's relatively fast - compared to previous version, which needed to do full sequential scan, it will only need to get all the events that have `dataset_versions` joined to namespace - which is a fraction compared to full table scan. It would be possible to make that faster if we could "predicate push down" filters on time. `created_at` is Marquez processing time field - we can't use that.

However, it's relatively pointless. Above query won't take into account events, where one of the input datasets has given namespace, and we don't see a fast way to get them with current database schema. 

We could either

1. get rid of the namespace filtering altogether, leaving only "get all events" endpoint
2. leave very inefficient endpoint, which will need to sequentially scan all the lineage_events table
3. use above query, which will be much faster, but won't include input datasets. 

Thoughts?

mobuchowski: I decided to remove this feature, and restore it if needed in future PR.

api/src/main/java/marquez/api/EventResource.java
mobuchowski: Removed this endpoint for now.

api/src/main/java/marquez/api/EventResource.java
mobuchowski: Done

pawel-big-lebowski (approved):

PR #2073, opened 2022-08-15
Update rewrite_jobs_fqn_table query to reduce the size of CTE and avo…

collado-mike:
### Problem

The rewrite_jobs_fqn_table function generates a CTE that contains the entire set of jobs for calculating symlinks instead of only including jobs that are actually involved in a symlink chain.

### Solution

Rewrite query to avoid the expensive CTE generation. I tested in a dev installation that was suffering from huge spikes in CPU utilization every hour. After altering the query (around 16:00 on this graph), the CPU spikes are considerably smaller.

Screen Shot 2022-08-15 at 10 29 29 AM

This query change had an impact on the search query and test, so those were adjusted accordingly.

### Checklist

### DIFF
--- a/api/src/test/java/marquez/db/SearchDaoTest.java
+++ b/api/src/test/java/marquez/db/SearchDaoTest.java
@@ -26,6 +26,7 @@ import marquez.db.models.NamespaceRow;
 import marquez.jdbi.MarquezJdbiExternalPostgresExtension;
 import marquez.service.models.JobMeta;
 import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.mapper.MapMapper;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


api/src/test/java/marquez/db/SearchDaoTest.java
```diff
@@ -109,7 +110,22 @@ public static void setUpOnce(final Jdbi jdbi) throws SQLException {
...
+        .forEach(System.out::println);
+    jdbi.withHandle(h -> h.createQuery("SELECT * FROM jobs_fqn").map(new MapMapper()).list())
+        .forEach(System.out::println);
+    jdbi.withHandle(
+            h ->
+                h.createQuery(
+                        "SELECT uuid, name, namespace_name, unnest(aliases) AS alias, symlink_target_uuid FROM jobs_view")
+                    .map(new MapMapper())
+                    .list())
+        .forEach(System.out::println);
```
wslulciuc, line 128: Are these `println` for debugging?

wslulciuc (approved):
LGTM 👍 🥇

PR #2074, opened 2022-08-15
Add ability to get open lineage events by run UUID.

fm100:
Signed-off-by: Minkyu Park <minkyu.park.200@gmail.com>

### Problem
OpenLineage events might contain lots of custom facets that marquez does not process properly, and it would be nice just
providing a way to retrieve raw events by run UUID so that the users can implement their own processing from run transition listener.

### Solution
This PR simply adds `SqlQuery` to the `OpenLineageDao` in order to provide ability to list lineage events by run UUID.

### Checklist

### DIFF
--- a/api/src/main/java/marquez/db/OpenLineageDao.java
+++ b/api/src/main/java/marquez/db/OpenLineageDao.java
@@ -86,7 +86,7 @@ public interface OpenLineageDao extends BaseDao {
       String producer);
 
   @SqlQuery("SELECT event FROM lineage_events WHERE run_uuid = :runUuid")
-  List<LineageEvent> listLineageEventsByRunUuid(UUID runUuid);
+  List<LineageEvent> findOlEventsByRunUuid(UUID runUuid);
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -81,6 +85,9 @@ void createLineageEvent(
       PGobject event,
       String producer);
 
+  @SqlQuery("SELECT event FROM lineage_events WHERE run_uuid = :runUuid")
+  List<LineageEvent> listLineageEventsByRunUuid(UUID runUuid);
```
wslulciuc, line 89: _Minor_: For DAOs, we use `findX`, so I would name this method `findOlEventsByRunUuid()`

wslulciuc (approved):
Minor suggestion on the method name for the DAO, otherwise, good stuff 💯

api/src/main/java/marquez/db/OpenLineageDao.java
fm100: Thank you for the feedback. I'll rename it.

PR #2076, opened 2022-08-18
Add proposal for ol facet tables

wslulciuc:
This PR adds the proposal: _Optimize query performance for OpenLineage facets_


proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,78 @@
...
+| job_namespace | `TEXT`       |
+| producer      | `TEXT`       |
+
+> **Note:** The table has an index on `run_uuid` and a _multicolumn_ index on `job_name`, `job_namespace`.
+
+The `event` column contains the raw OpenLineage event. When Marquez handles the event, metadata for datasets, jobs, and runs are inserted into their respective tables.
+
+OpenLineage's core model is extensible via _facets_. A `facet` is user-defined metadata and enables entity enrichment. Initially, returning dataset, job, and run facets via the REST API was not supported, but eventually added in release [`0.14.0`](https://github.com/MarquezProject/marquez/compare/0.13.1...0.14.0). The implementation was simple: when querying the `datasets`, `jobs`, or `runs` tables, also query the `lineage_events` table for facets.
+
+We knew the initial implementation would have to eventually be revisited. That is, OpenLineage events can easily exceed **>** **`10MBs`** resulting in out-of-memory (OOM) errors as facet queries require loading the raw `event` in memory, then filtering for relevant facets. This proposal outlines how we can optimize query performance for OpenLineage facets.
```
fm100, line 31: 10MB is not small for a single record, but it may not be big enough for resulting in OOM. Can you please elaborate the query example which causes an issue, which may load large number of lineage event records? Besides, we may want to address the high CPU usage problem by unnecessarily parsing big JSONB object when not needed, for example loading run facets when querying dataset.

collado-mike (commented):
Nice. Can you add specifics on how conflicts will be dealt with? Are these facet tables append-only? E.g., if two facets with the same name (I assume the `name` column in the tables refers to the facet name) but different contents are received for the same dataset version (e.g., two different GreatExpectations suites on the same data), will both records be added to the tables? If not, which one wins? First received? Last received? If both records are inserted, will they be merged at query time? Or just appended one after the other? Are there indexes on these tables? In particular, the `runs` table can get very, very large. The same will be true of the `runs_facets` table eventually.

proposals/2078-optimization-ol-facets.md
wslulciuc: > 10MB is not small for a single record, but it may not be big enough for resulting in OOM

Very true, but if you load a lot of them, it will result in OOM:

```sql
          + "    LEFT JOIN LATERAL (\n"
          + "        SELECT run_uuid, event_time, event FROM lineage_events\n"
          + "        WHERE run_uuid = dv.run_uuid\n"
          + "    ) e ON e.run_uuid = dv.run_uuid\n"
          + "    UNION\n"
          + "    SELECT d.uuid, d.name, d.namespace_name, rim.run_uuid, lifecycle_state, event_time, event\n"
          + "    FROM selected_datasets d\n"
          + "    INNER JOIN dataset_versions dv ON dv.uuid = d.current_version_uuid\n"
          + "    LEFT JOIN runs_input_mapping rim ON dv.uuid = rim.dataset_version_uuid\n"
          + "    LEFT JOIN LATERAL (\n"
          + "        SELECT run_uuid, event_time, event FROM lineage_events\n"
          + "        WHERE run_uuid = rim.run_uuid\n"
          + "    ) e ON e.run_uuid = rim.run_uuid\n"
```

The partial query above is from [`DatasetDao.findDatasetByName()`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/DatasetDao.java#L82). This specific query loads multiple OL events and is where we've seen the error. I should've added this context in the proposal (I'll do that now).

wslulciuc:
@collado-mike: All great questions that I'll elaborate on in the proposal!

proposals/2078-optimization-ol-facets.md
pawel-big-lebowski, line 31: From what I understand, we load the events to memory in Marquez and do facet filtering there. 
If so, the solution should be to offload filtering to database. The short term solution could be filtering json content in postgres as described [here](https://scalegrid.io/blog/using-jsonb-in-postgresql-how-to-effectively-store-index-json-data-in-postgresql/#7) (please note indexing json content is also possible). In this particular example we could select output dataset facets from event json within postgresql, instead of selecting whole events. 

But `lineage_events` will grow over time and querying for datasets' facets will slow down. Normalizing json facets, as described within this proposal, is a good way to go. 

The only problem with separate tables is a backfill procedure which is a heavy operation. I am not sure whether lazy migration would work. The existence of some facets in new tables does not mean we collected all the existing facets, including those from `lineage_events` table.

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+
+## Proposal
+
+To improve query performance for facets, and avoid querying the `lineage_events` table, we propose adding the following tables to group facets by how they will be accessed:
```
julienledem, line 37: I would add some fields from the `lineage_event` table to all the facets tables for the following purpose:
- `event_time`: this will allow us to expire old data more easily, we should start thinking about partitioning those tables.
- `run_uuid`  and  `event_type`: to be able to track down the event this facet came from as needed.

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Table `dataset_version_facets`
+
+| **COLUMN**           | **TYPE**  |
+|----------------------|-----------|
+| uuid                 | `UUID`    |
+| dataset_version_uuid | `UUID`    |
```
julienledem, line 44: For dataset facets we should keep track if those were `dataset facets`, `input facets` or `output facets`.
possibly through a `type`: enum('dataset', 'input', 'output') field or by having different tables.
https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#dataset-facets

julienledem (commented):
That looks great, I left some comments

proposals/2078-optimization-ol-facets.md
wslulciuc: I've added the following columns to each facet table:
* `run_uuid`
* `lineage_event_time`
* `lineage_event_type`

proposals/2078-optimization-ol-facets.md
wslulciuc: Very reasonable point, @pawel-big-lebowski. But, to get ahead of the issue, like you stated, we'll want to normalize facets.

proposals/2078-optimization-ol-facets.md
wslulciuc: Great idea. I've added `type` column.

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,117 @@
...
+
+To improve query performance for facets, and avoid querying the `lineage_events` table, we propose adding the following tables to group facets by how they will be accessed:
+
+### Table `dataset_facets`
+
+| **COLUMN**         | **TYPE**      |
+|--------------------|---------------|
+| uuid **(PK)**      | `UUID`        |
+| created_at         | `TIMESTAMPTZ` |
+| run_uuid           | `UUID`        |
```
pawel-big-lebowski, line 66: Dataset facets are related to datasets instead of runs.
I think that instead of `run_uuid`, a `dataset_version_uuid` column should be used. 
This will help later retrieving dataset information by limiting amount of joins: getting facets of a dataset will not require jumping though `runs` table.

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,117 @@
...
+
+> **Table 2:** Facets for a given job.
+
+### Table `run_facets`
+
+| **COLUMN**         | **TYPE**      |
+|--------------------|---------------|
+| uuid **(PK)**      | `UUID`        |
+| created_at         | `TIMESTAMPTZ` |
+| run_uuid           | `UUID`        |
```
pawel-big-lebowski, line 97: as above: please consider referencing `jobs` table.

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,117 @@
...
+| lineage_event_time | `TIMESTAMPTZ` |
+| lineage_event_type | `VARCHAR`     |
+| type               | `VARCHAR`     |
+| name               | `VARCHAR`     |
+| facet              | `JSONB`       |
```
pawel-big-lebowski, line 71: Do we need `uuid` column as a primary key / at all?
Would a combination of `dataset_version_uuid` and `name` be enough as a primary key?

pawel-big-lebowski (commented):
I think it's a great proposal which can have huge impact on Marquez performance 🚀 
I've put some comments which are worth considering (if not already considered).

proposals/2078-optimization-ol-facets.md
mobuchowski: Reading later part of proposal, it seems that facets will be append-only and always appended, meaning you can get the same facet multiple times - for example, on `START` and `COMPLETE` event.

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,117 @@
...
+| lineage_event_type | `VARCHAR`     |
+| name               | `VARCHAR`     |
+| facet              | `JSONB`       |
+
+> **Table 3:** Facets for a given run.
+
+Note, facet tables will be:
+
+* Append only, mirroring the current insertion pattern of the `lineage_events` table; therefore, avoiding facet conflicts
+* Merging facets will follow a _first-to-last_ received order; meaning, facet rows will be merged post query using [`MapperUtils.toFacetsOrNull()`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/mappers/MapperUtils.java#L50) mirroring the current logic (i.e. newer facets will be added or override older facet values based on when the OpenLineage event was received)
```
mobuchowski, line 108: In some cases it might make sense to accumulate content inside the facets themselves. 

For example, streaming Flink job might report how many records were processed per checkpoint - sometimes sending multiple results per OL event. The result facet should return list of those reports.

proposals/2078-optimization-ol-facets.md
wslulciuc: Originally, I had left the `jobs` table reference out, but quickly realized I needed `job_uuid` in `job_facets`!

wslulciuc: > as above: please consider referencing jobs table.

@pawel-big-lebowski, do you want to reference the `jobs` table in `run_facets`? I added the reference to `job_facets`, but in `run_facets` we'd want to reference the run for a given job.

proposals/2078-optimization-ol-facets.md
wslulciuc: I think having the `run_uuid` in the `dataset_facets` is useful (if only for debugging purposes). I've added the `dataset_uuid`, but we can't set `dataset_version_uuid` as we don't yet now the data version `UUID` for a run in the `RUNNING` state. Having `dataset_uuid` allows us to `JOIN` on `dataset_versions.dataset_uuid`.

proposals/2078-optimization-ol-facets.md
wslulciuc: Making the table append-only does simplify the implementation (at least, initially). This was also considered to avoid the complexity of managing input / output datasets facets (ex: `dataQualityMetrics`, `dataQualityAssertions`, etc). That is, a dataset version (or job version) is immutable and I'd like to avoid any mutations on those tables.

proposals/2078-optimization-ol-facets.md
wslulciuc: I agree, certain facets can be handled differently (i.e. accumulated) based on some context. I'll make a note on the proposals scope / limitations.

mobuchowski: @wslulciuc let's also create follow-up issue, will be good to go then on this issue.

proposals/2078-optimization-ol-facets.md
pawel-big-lebowski: `run_uuid` is enough. sry, if this was here from the beginning.

proposals/2078-optimization-ol-facets.md
pawel-big-lebowski: ok, thx

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Implementation
+
+The implementation requires:
+
+1. Schema changes to create the facet tables outlined above.
+2. Using the facet tables instead the `lineage_events` table to query for facets.
+3. Lazy migration, the facet tables will be queried, and if no facets are returned, then the `lineage_events` table; this approach avoids a backfill, but one will still be needed.
+
+## Migration procedure
```
pawel-big-lebowski, line 119: Please also refer to this:
https://github.com/MarquezProject/marquez/pull/2152/files?short_path=17843f7#diff-17843f7d4567ca029ee0d63f56e6b75a000b384768d9e0352badffad66eeea3c

to see migration from a user's perspective.

proposals/2078-optimization-ol-facets.md
```diff
@@ -0,0 +1,201 @@
...
+An extra single-row `migration_lock` table will be introduced:
+
+| **COLUMN** | **TYPE**      |
+|------------|---------------|
+| run_uuid   | `UUID`        |
+| created_at | `TIMESTAMPTZ` |
+
+Data migration will be run in chunks:
+```
+WITH events_chunk AS (
```
mobuchowski, line 168: I think good idea is to paste real code or remove the example.

mobuchowski (approved):

proposals/2078-optimization-ol-facets.md
pawel-big-lebowski: I tried to describe literally how the migration is done but at the end I think this SQL code speaks more and is more accurate. It is actually part of real code with some parts truncated.

Issue #2066, opened 2022-08-10
Implement symlinks in Marquez


pawel-big-lebowski:
**Problem:** 

We need an ability to store alternative dataset names. For example hive datasets can be identified by their data files` location or metastore uri with database and table.

**Solution in Spec:** 

`SymlinksDatasetFacet` -> https://github.com/OpenLineage/OpenLineage/pull/936

Implementation in Marquez: 

**Model changes:**

 * Create extra dataset_symlink table in Marquez with columns: (symlinkUid, name, namespaceUid, symlinkType)
 * Replace `name` field in `datasets` table with `symlinkUid`

**Implementation follows the proposed DB changes:**

First PR -> reflect current behaviour in modifed schema
 * provide migration SQL for existing instances
 * create a `dataset_symlink` row whenever dataset is created
 * modify SQLs in dataset_version_dao, etc. 

Second PR
 * Extract symlink facet when posting new OpenLineage event
 * fill `dataset_symlink` with multiple entries per OL event.

Issue #2047, opened 2022-07-26
Add API load test in CI


wslulciuc:
As we improve the Marquez DB model, having an API load test in CI can help spot DB bottlenecks sooner (ex: [`2043`](https://github.com/MarquezProject/marquez/pull/2043)). I recommend we us a load testing tool like [`k6`](https://k6.io). Here's a great writeup: [How to Perform Load Testing with k6 using GitHub Actions](https://k6.io/blog/load-testing-using-github-actions/)

PR #2087, opened 2022-08-25
dataset symlinks provided

pawel-big-lebowski:

### Problem

A `SymlinkDatasetFacet` has been introduced in spec recently (https://github.com/OpenLineage/OpenLineage/pull/936) and it allows a dataset to be identified by multiple `(name, namespace)` tuples. We need to modify Marquez to handle it, as currently many joins to dataset table are done directly based on `name` and `namespace` value. 

Part of: #2066

### Solution

This PR contains:
* a refactor of current Marquez database model to allow alternative dataset names by creating an extra `dataset_symlinks` table with dataset name. 
* removal of dataset's name from dataset table.
* implementation of `SymlinkDatasetFacet` logic. 

### Checklist

### DIFF
--- a/api/src/main/java/marquez/db/Columns.java
+++ b/api/src/main/java/marquez/db/Columns.java
@@ -80,6 +80,9 @@ public final class Columns {
   public static final String FIELD_UUIDS = "field_uuids";
   public static final String LIFECYCLE_STATE = "lifecycle_state";
 
+  /* DATASET SYMLINK ROW COLUMNS */
+  public static final String IS_PRIMARY = "is_primary";
+
   /* STREAM VERSION ROW COLUMNS */
   public static final String SCHEMA_LOCATION = "schema_location";
 
--- a/api/src/main/java/marquez/db/DatasetDao.java
+++ b/api/src/main/java/marquez/db/DatasetDao.java
@@ -23,6 +23,7 @@ import marquez.common.models.TagName;
 import marquez.db.mappers.DatasetMapper;
 import marquez.db.mappers.DatasetRowMapper;
 import marquez.db.models.DatasetRow;
+import marquez.db.models.DatasetSymlinkRow;
 import marquez.db.models.DatasetVersionRow;
 import marquez.db.models.NamespaceRow;
 import marquez.db.models.SourceRow;
@@ -73,8 +74,7 @@ public interface DatasetDao extends BaseDao {
       WITH selected_datasets AS (
           SELECT d.*
           FROM datasets_view d
-          WHERE d.namespace_name = :namespaceName
-          AND d.name = :datasetName
+          WHERE CAST((:namespaceName, :datasetName) AS DATASET_NAME) = ANY(d.dataset_symlinks)
       ), dataset_runs AS (
           SELECT d.uuid, d.name, d.namespace_name, dv.run_uuid, dv.lifecycle_state, event_time, event
           FROM selected_datasets d
@@ -229,7 +229,7 @@ public interface DatasetDao extends BaseDao {
             :description,
             :isDeleted,
             false
-          ) ON CONFLICT (namespace_uuid, name)
+          ) ON CONFLICT (uuid)
           DO UPDATE SET
           type = EXCLUDED.type,
           updated_at = EXCLUDED.updated_at,
@@ -275,7 +275,7 @@ public interface DatasetDao extends BaseDao {
           + ":sourceName, "
           + ":name, "
           + ":physicalName) "
-          + "ON CONFLICT (namespace_uuid, name) "
+          + "ON CONFLICT (uuid) "
           + "DO UPDATE SET "
           + "type = EXCLUDED.type, "
           + "updated_at = EXCLUDED.updated_at, "
@@ -296,8 +296,10 @@ public interface DatasetDao extends BaseDao {
       """
   UPDATE datasets
   SET is_hidden = true
-  WHERE namespace_name = :namespaceName
-  AND name = :name
+  FROM dataset_symlinks, namespaces
+  WHERE dataset_symlinks.dataset_uuid = datasets.uuid
+  AND namespaces.uuid = dataset_symlinks.namespace_uuid
+  AND namespaces.name=:namespaceName AND dataset_symlinks.name=:name
   RETURNING *
   """)
   Optional<DatasetRow> delete(String namespaceName, String name);
@@ -310,6 +312,10 @@ public interface DatasetDao extends BaseDao {
         createNamespaceDao()
             .upsertNamespaceRow(
                 UUID.randomUUID(), now, namespaceName.getValue(), DEFAULT_NAMESPACE_OWNER);
+    DatasetSymlinkRow symlinkRow =
+        createDatasetSymlinkDao()
+            .upsertDatasetSymlinkRow(
+                UUID.randomUUID(), datasetName.getValue(), namespaceRow.getUuid(), true, null, now);
     SourceRow sourceRow =
         createSourceDao()
             .upsertOrDefault(
@@ -318,13 +324,12 @@ public interface DatasetDao extends BaseDao {
                 now,
                 datasetMeta.getSourceName().getValue(),
                 "");
-    UUID newDatasetUuid = UUID.randomUUID();
     DatasetRow datasetRow;
 
     if (datasetMeta.getDescription().isPresent()) {
       datasetRow =
           upsert(
-              newDatasetUuid,
+              symlinkRow.getUuid(),
               datasetMeta.getType(),
               now,
               namespaceRow.getUuid(),
@@ -338,7 +343,7 @@ public interface DatasetDao extends BaseDao {
     } else {
       datasetRow =
           upsert(
-              newDatasetUuid,
+              symlinkRow.getUuid(),
               datasetMeta.getType(),
               now,
               namespaceRow.getUuid(),
@@ -349,7 +354,8 @@ public interface DatasetDao extends BaseDao {
               datasetMeta.getPhysicalName().getValue());
     }
 
-    updateDatasetMetric(namespaceName, datasetMeta.getType(), newDatasetUuid, datasetRow.getUuid());
+    updateDatasetMetric(
+        namespaceName, datasetMeta.getType(), symlinkRow.getUuid(), datasetRow.getUuid());
 
     TagDao tagDao = createTagDao();
     List<DatasetTagMapping> datasetTagMappings = new ArrayList<>();
--- a/api/src/main/java/marquez/db/DatasetSymlinkDao.java
+++ b/api/src/main/java/marquez/db/DatasetSymlinkDao.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2018-2022 contributors to the Marquez project
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package marquez.db;
+
+import java.time.Instant;
+import java.util.Optional;
+import java.util.UUID;
+import marquez.db.mappers.DatasetSymlinksRowMapper;
+import marquez.db.models.DatasetSymlinkRow;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+@RegisterRowMapper(DatasetSymlinksRowMapper.class)
+public interface DatasetSymlinkDao extends BaseDao {
+
+  default DatasetSymlinkRow upsertDatasetSymlinkRow(
+      UUID uuid, String name, UUID namespaceUuid, boolean isPrimary, String type, Instant now) {
+    doUpsertDatasetSymlinkRow(uuid, name, namespaceUuid, isPrimary, type, now);
+    return findDatasetSymlinkByNamespaceUuidAndName(namespaceUuid, name).orElseThrow();
+  }
+
+  @SqlQuery("SELECT * FROM dataset_symlinks WHERE namespace_uuid = :namespaceUuid and name = :name")
+  Optional<DatasetSymlinkRow> findDatasetSymlinkByNamespaceUuidAndName(
+      UUID namespaceUuid, String name);
+
+  @SqlUpdate(
+      """
+          INSERT INTO dataset_symlinks (
+          dataset_uuid,
+          name,
+          namespace_uuid,
+          is_primary,
+          type,
+          created_at,
+          updated_at
+          ) VALUES (
+          :uuid,
+          :name,
+          :namespaceUuid,
+          :isPrimary,
+          :type,
+          :now,
+          :now)
+          ON CONFLICT (name, namespace_uuid) DO NOTHING""")
+  void doUpsertDatasetSymlinkRow(
+      UUID uuid, String name, UUID namespaceUuid, boolean isPrimary, String type, Instant now);
+}
--- a/api/src/main/java/marquez/db/OpenLineageDao.java
+++ b/api/src/main/java/marquez/db/OpenLineageDao.java
@@ -33,6 +33,7 @@ import marquez.db.JobVersionDao.BagOfJobVersionInfo;
 import marquez.db.mappers.LineageEventMapper;
 import marquez.db.models.DatasetFieldRow;
 import marquez.db.models.DatasetRow;
+import marquez.db.models.DatasetSymlinkRow;
 import marquez.db.models.DatasetVersionRow;
 import marquez.db.models.JobContextRow;
 import marquez.db.models.JobRow;
@@ -120,6 +121,7 @@ public interface OpenLineageDao extends BaseDao {
 
   default UpdateLineageRow updateBaseMarquezModel(LineageEvent event, ObjectMapper mapper) {
     NamespaceDao namespaceDao = createNamespaceDao();
+    DatasetSymlinkDao datasetSymlinkDao = createDatasetSymlinkDao();
     DatasetDao datasetDao = createDatasetDao();
     SourceDao sourceDao = createSourceDao();
     JobDao jobDao = createJobDao();
@@ -316,6 +318,7 @@ public interface OpenLineageDao extends BaseDao {
                 runUuid,
                 true,
                 namespaceDao,
+                datasetSymlinkDao,
                 sourceDao,
                 datasetDao,
                 datasetVersionDao,
@@ -337,6 +340,7 @@ public interface OpenLineageDao extends BaseDao {
                 runUuid,
                 false,
                 namespaceDao,
+                datasetSymlinkDao,
                 sourceDao,
                 datasetDao,
                 datasetVersionDao,
@@ -532,6 +536,7 @@ public interface OpenLineageDao extends BaseDao {
       UUID runUuid,
       boolean isInput,
       NamespaceDao namespaceDao,
+      DatasetSymlinkDao datasetSymlinkDao,
       SourceDao sourceDao,
       DatasetDao datasetDao,
       DatasetVersionDao datasetVersionDao,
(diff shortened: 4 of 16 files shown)


mobuchowski:
>removal of dataset's name from dataset table.

Why?

mobuchowski:
@pawel-big-lebowski let me try to summarize changes to model: after this change, internally `datasets` table don't represent "full" dataset, since name is dropped - getting `dataset identifier` requires joining with symlinks table. `datasets` table still contains `namespace_name` - it's the same namespace name that is used by `primary` symlink.

Primary symlink is the one that is being send in regular dataset name, rather the ones in `SymlinksDatasetFacet`. That suggest question: what happens if two jobs produce the same dataset, and they have the identifiers in different order; that is one has namespace X and name A as primary, and second one has namespace Y and name B? Does the first one "win" and dataset is forever known as (X, A) pair?

pawel-big-lebowski:
> @pawel-big-lebowski let me try to summarize changes to model: after this change, internally `datasets` table don't represent "full" dataset, since name is dropped - getting `dataset identifier` requires joining with symlinks table. `datasets` table still contains `namespace_name` - it's the same namespace name that is used by `primary` symlink.
> 

Yes, you're right. The reason behind droping a `name` column  is to make sure that no-one now or later tries to retrieve dataset by filtering `name` column in `dataset` table. Whenever this is required, `dataset_symlinks` table should be joined to make sure we look for alternative names. 

The `namespace_name` column existed before together with `namespace_uuid` field. In my opinion, it should be removed but I found out it's out of scope for this PR. 

> Primary symlink is the one that is being send in regular dataset name, rather the ones in `SymlinksDatasetFacet`. That suggest question: what happens if two jobs produce the same dataset, and they have the identifiers in different order; that is one has namespace X and name A as primary, and second one has namespace Y and name B? Does the first one "win" and dataset is forever known as (X, A) pair?

Yes, the first (namespace/name) becomes primary and I think it's the best we can do.

julienledem (commented):
Overall this looks good to me. Is it going to conflict with the delete feature @mobuchowski is working on?

collado-mike (commented):
Given the problem the dataset symlinking aims to solve is to resolve lineage for jobs that may be accessing datasets by different names, I'm wondering if we can scope the symlinks functionality to _just_ the lineage API. As is, its impact is pervasive and, as we saw with the job symlinks, that can have unintended consequences. I think we can start by scoping this to just the lineage API, then adjust other APIs as needed.

pawel-big-lebowski:
> Given the problem the dataset symlinking aims to solve is to resolve lineage for jobs that may be accessing datasets by different names, **I'm wondering if we can scope the symlinks functionality to _just_ the lineage API.** As is, its impact is pervasive and, as we saw with the job symlinks, that can have unintended consequences. I think we can start by scoping this to just the lineage API, then adjust other APIs as needed.

In order to confirm that, one should know all the Marquez features to make sure datasets are not joined by name whenever symlinks should be applied. This creates an assumption which is not stated explicitly, so is error prone in future. That's why I am in favour of an assumption that `dataset_symlinks` should be always use to retrieve a dataset by name. If a dataset has two different names, it gets returned to matter how we request it. 

Performance wise, this should not affect postgresql. It's just an extra join of index column to retrieve extra column (not used for filtering) and an uncorrelated subquery to get possible symlinks run before existing queries. Postgres is smart and joins on foreign keys, which are just applied on result rows, should be fast. Full-scans digging jsonb columns are slow, we know that and there is proposal to fix it.

To sum up, there is a tradeoff "(1) not clean assumption in DB modelling" vs "(2) DB performance risk that should not happen but cannot be confirmed due to lack of perf tests". One one hand, I would go with (2) as it gives greater chance to make it clean in future. On the other hand, I don't mind going with (1) as I haven't worked that much with Marquez DB perf issues.

api/src/main/resources/marquez/db/migration/V46__dataset_symlinks.sql
```diff
@@ -0,0 +1,18 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+CREATE TABLE dataset_symlinks (
+  dataset_uuid      UUID,
```
pawel-big-lebowski, line 4: TODO: index on `dataset_uuid` missing

mobuchowski:
@julienledem great point. I talked with @pawel-big-lebowski and we aligned to use `datasets_view` to provide joined data from `datasets` and `datasets_symlinks`. This way, most of the SQL busywork behind this change will be opaque to queries.

With that, only issue is that "namespaces" aren't unique anymore - methods like 
```
List<Dataset> findAll(String namespaceName, int limit, int offset)
```
still need to filter by `is_primary` field to not return duplicates.

api/src/main/java/marquez/db/Columns.java
```diff
@@ -80,6 +80,9 @@ private Columns() {}
   public static final String FIELD_UUIDS = "field_uuids";
   public static final String LIFECYCLE_STATE = "lifecycle_state";
 
+  /* DATASET SYMLINK ROW COLUMNS */
+  public static final String IS_PRIMARY = "is_primary";
```
wslulciuc, line 84: Is the `is_primary` column used to indicated that it's the _primary_ name of the dataset?

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -42,9 +43,14 @@
 @RegisterRowMapper(DatasetMapper.class)
 public interface DatasetDao extends BaseDao {
   @SqlQuery(
-      "SELECT EXISTS ("
+      "WITH symlink AS (\n"
```
wslulciuc, line 46: _Minor_: `symlink` -> `symlinks`

api/src/main/java/marquez/db/Columns.java
wslulciuc: Ahh it totally is 👍

api/src/main/java/marquez/db/DatasetSymlinkDao.java
```diff
@@ -0,0 +1,48 @@
...
+public interface DatasetSymlinkDao extends BaseDao {
+
+  default DatasetSymlinkRow upsertDatasetSymlinkRow(
+      UUID uuid, String name, UUID namespaceUuid, boolean isPrimary, Instant now) {
+    doUpsertDatasetSymlinkRow(uuid, name, namespaceUuid, isPrimary, now);
+    return findDatasetSymlinkByNamespaceUidAndName(namespaceUuid, name).orElseThrow();
+  }
+
+  @SqlQuery("SELECT * FROM dataset_symlinks WHERE namespace_uuid = :namespaceUuid and name = :name")
+  Optional<DatasetSymlinkRow> findDatasetSymlinkByNamespaceUidAndName(
```
wslulciuc, line 27: _Typo_: `findDatasetSymlinkByNamespaceUidAndName()` -> `findDatasetSymlinkByNamespaceUuidAndName()`

api/src/main/java/marquez/db/DatasetSymlinkDao.java
```diff
@@ -0,0 +1,48 @@
...
+      UUID namespaceUuid, String name);
+
+  @SqlUpdate(
+      "INSERT INTO dataset_symlinks ("
```
wslulciuc, line 31: In [`V46__dataset_symlinks.sql`](https://github.com/MarquezProject/marquez/blob/feature/dataset-symlinks/api/src/main/resources/marquez/db/migration/V46__dataset_symlinks.sql) a `type` column is defined, but never set here. Do we need the column?

api/src/main/resources/marquez/db/migration/V46__dataset_symlinks.sql
```diff
@@ -0,0 +1,18 @@
...
+  name              VARCHAR(255) NOT NULL,
+  namespace_uuid    UUID REFERENCES namespaces(uuid),
+  type              VARCHAR(64),
```
wslulciuc, line 7: Do we need `type` defined in `dataset_symlinks`? Or, would it be helpful to know the type of symlink?

api/src/main/resources/marquez/db/migration/V46__dataset_symlinks.sql
```diff
@@ -0,0 +1,18 @@
...
+  created_at        TIMESTAMP NOT NULL,
+  updated_at        TIMESTAMP NOT NULL,
+  UNIQUE (namespace_uuid, name)
+);
+
+INSERT INTO dataset_symlinks (dataset_uuid, name, namespace_uuid, is_primary, created_at, updated_at)
+SELECT d.uuid, d.name, d.namespace_uuid, TRUE, d.created_at, d.updated_at FROM datasets d;
+
+ALTER TABLE datasets DROP CONSTRAINT datasets_namespace_uuid_name_key;
+ALTER TABLE datasets DROP COLUMN name;
```
wslulciuc, line 18: Mind if we split these changes into two separate migrations?

* `V46__add_dataset_symlinks.sql`
* `V46.1__drop_name_column_in_datasets.sql`

wslulciuc: Do we need to drop the `name` column from `datasets`? I'm wondering, can `datasets.name` be viewed as the primary name? Or, will the `name` column in `datasets_view` (introduced in https://github.com/MarquezProject/marquez/pull/2032) be considered the dataset name (=_primary_)?

api/src/main/java/marquez/db/DatasetSymlinkDao.java
```diff
@@ -0,0 +1,48 @@
...
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+@RegisterRowMapper(DatasetSymlinksRowMapper.class)
+public interface DatasetSymlinkDao extends BaseDao {
+
+  default DatasetSymlinkRow upsertDatasetSymlinkRow(
+      UUID uuid, String name, UUID namespaceUuid, boolean isPrimary, Instant now) {
+    doUpsertDatasetSymlinkRow(uuid, name, namespaceUuid, isPrimary, now);
+    return findDatasetSymlinkByNamespaceUidAndName(namespaceUuid, name).orElseThrow();
```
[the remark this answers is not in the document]
wslulciuc: `RETURNING` is great, but only if the row as changed. To get around this, most of our queries update the `updated_at` column even though no changes have occurred; which, to me, isn't ideal and something I'd like to move away from.

wslulciuc:
> To sum up, there is a tradeoff "(1) not clean assumption in DB modelling" vs "(2) DB performance risk that should not happen but cannot be confirmed due to lack of perf tests". On one hand, I would go with (2) as it gives great chance to make it clean in future. On the other hand, I don't mind going with (1) as I haven't worked that much with Marquez DB perf issues.

@pawel-big-lebowski I think @collado-mike's comment: "_As is, its impact is pervasive and, as we saw with the job symlinks, that can have unintended consequences_" wasn't necessarily performance related (though there was follow up work on job symlinks needed to improve performance and implementation), but rather unexpected behavior given the number of queries modified. 

We're hoping to add [load testing](https://github.com/MarquezProject/marquez/blob/main/docs/load-testing.md) to CI at some point, see #2047

wslulciuc (changes requested):
I would also update our [`CHANGELOG.md`](https://github.com/MarquezProject/marquez/blob/main/CHANGELOG.md) with this sweet sweet feature!

wslulciuc:
> The namespace_name column existed before together with namespace_uuid field. In my opinion, it should be removed but I found out it's out of scope for this PR.

@pawel-big-lebowski mind opening an issue to capture this change as follow up work?

api/src/main/resources/marquez/db/migration/V46__dataset_symlinks.sql
wslulciuc: Don't forget to add this index ;)

api/src/main/java/marquez/db/DatasetSymlinkDao.java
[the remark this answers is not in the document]
pawel-big-lebowski: I saw that pattern in other pieces of Marquez and decided to reuse it.

api/src/main/java/marquez/db/DatasetDao.java
pawel-big-lebowski: code got removed when switching to `dataset_view`

api/src/main/resources/marquez/db/migration/V46__dataset_symlinks.sql
pawel-big-lebowski: Based on the feedback, `name` column won't be dropped to mitigate the risk of migration issues and limit unexpected behavior changes.

api/src/main/java/marquez/db/DatasetSymlinkDao.java
pawel-big-lebowski: This column is part of a spec to contain some description of an identifier for example: `table` may be a type for symlink names containing table name.

I forgot to make use of the column. Thanks for spotting this. 

I added some code to save a `type` and included it in unit test `marquez.db.OpenLineageDaoTest#testUpdateMarquezModelDatasetWithSymlinks` to make sure the type from OL event is written to Marquez DB.

pawel-big-lebowski:
@collado-mike , @pawel-big-lebowski  and @wslulciuc, thank you for great reviews and feedback.

Changes applied according to that: 
 * will reuse a  `dataset_view` introduced by Maciej,
 * won't drop `name` column in `datasets` table to avoid risky migrations and limit amount of possible unexpected behavior changes.

api/src/main/java/marquez/db/DatasetSymlinkDao.java
```diff
@@ -0,0 +1,50 @@
...
+          + "updated_at"
+          + ") VALUES ( "
+          + ":uuid, "
+          + ":name, "
+          + ":namespaceUuid, "
+          + ":isPrimary, "
+          + ":type, "
+          + ":now, "
+          + ":now) "
+          + "ON CONFLICT (name, namespace_uuid) DO NOTHING")
```
mobuchowski, line 47: If we're adding new SQL blocks, please use Java 17's text blocks feature.

api/src/main/resources/marquez/db/migration/V47__dataset_symlinks.sql
```diff
@@ -0,0 +1,24 @@
...
+  UNIQUE (namespace_uuid, name)
+);
+
+CREATE INDEX dataset_symlinks_dataset_uuid on dataset_symlinks (dataset_uuid);
+
+INSERT INTO dataset_symlinks (dataset_uuid, name, namespace_uuid, is_primary, created_at, updated_at)
+SELECT d.uuid, d.name, d.namespace_uuid, TRUE, d.created_at, d.updated_at FROM datasets d;
+
+DROP TYPE IF EXISTS DATASET_NAME;
+CREATE TYPE DATASET_NAME AS (
```
mobuchowski, line 20: Why do we need it?

api/src/main/resources/marquez/db/migration/V47__dataset_symlinks.sql
```diff
@@ -0,0 +1,24 @@
+/* SPDX-License-Identifier: Apache-2.0 */
```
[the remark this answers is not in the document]
pawel-big-lebowski: sure, will do that once the review is over ;-)

api/src/main/resources/marquez/db/migration/V47__dataset_symlinks.sql
pawel-big-lebowski: because of this:
```
WHERE CAST((:namespaceName, :datasetName) AS DATASET_NAME) = ANY(d.dataset_symlinks)
```
https://github.com/MarquezProject/marquez/pull/2087/files#diff-c687ca258e023847591c9b8c044adddaae1e7c1f6f9483928bf2871050cf6a63R77

I thought it's necessary in postgresql to check if pair of varchars is contained in array of pairs of varchars.

mobuchowski: Ok, so you can do `WHERE namespace_name IN X AND name in Y` because it can match namespace, and match name, but those belong to different symlinks, right?

pawel-big-lebowski: > Ok, so you can do `WHERE namespace_name IN X AND name in Y` because it can match namespace, and match name, but those belong to different symlinks, right?

Exactly, @mobuchowski

api/src/main/java/marquez/db/DatasetSymlinkDao.java
wslulciuc: +1

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -310,6 +312,10 @@ default Dataset upsertDatasetMeta(
         createNamespaceDao()
             .upsertNamespaceRow(
                 UUID.randomUUID(), now, namespaceName.getValue(), DEFAULT_NAMESPACE_OWNER);
+    DatasetSymlinkRow symlinkRow =
+        createDatasetSymlinkDao()
+            .upsertDatasetSymlinkRow(
+                UUID.randomUUID(), datasetName.getValue(), namespaceRow.getUuid(), true, "", now);
```
wslulciuc, line 318: Can we default the symlink type to `UNKNOWN` or `null` if we don't know the symlink type rather than an empty string?

api/src/main/java/marquez/db/DatasetSymlinkDao.java
wslulciuc: Thanks! ❤️

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -540,6 +545,35 @@ default DatasetRecord upsertLineageDataset(
             formatNamespaceName(ds.getNamespace()),
             DEFAULT_NAMESPACE_OWNER);
 
+    DatasetSymlinkRow symlink =
+        datasetSymlinkDao.upsertDatasetSymlinkRow(
+            UUID.randomUUID(),
+            formatDatasetName(ds.getName()),
+            dsNamespace.getUuid(),
+            true,
+            "",
```
wslulciuc, line 554: Similar comment as above, I would avoid using an empty string for the symlink type and default to `UNKNOWN` or `null`.

api/src/main/java/marquez/service/models/DatasetSymlink.java
```diff
@@ -0,0 +1,28 @@
...
+import java.time.Instant;
+import java.util.Optional;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+public class DatasetSymlink {
```
wslulciuc, line 17: Is class `DatasetSymlink` used?

wslulciuc (approved):
Great work, @pawel-big-lebowski! I left some additional comments, but looks good to merge overall once they are resolved 💯 🥇

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -42,9 +43,14 @@
 @RegisterRowMapper(DatasetMapper.class)
 public interface DatasetDao extends BaseDao {
   @SqlQuery(
-      "SELECT EXISTS ("
+      "WITH symlink AS (\n"
+          + "    SELECT dataset_uuid\n"
+          + "    FROM dataset_symlinks\n"
+          + "    JOIN namespaces ON namespaces.uuid = dataset_symlinks.namespace_uuid\n"
+          + "    WHERE namespaces.name=:namespaceName and dataset_symlinks.name=:datasetName\n"
```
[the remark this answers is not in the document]
pawel-big-lebowski: i am using a view defined in Maciej's PR.

api/src/main/java/marquez/db/OpenLineageDao.java
pawel-big-lebowski: let's go with `null`

api/src/main/java/marquez/db/DatasetDao.java
pawel-big-lebowski: `null` is fine

PR #2121, opened 2022-09-13
document in code that JobDao.upsertJob executes trigger

mobuchowski:
Just a comment for future reader that JobDao.upsertJob does not actually execute SQL in `SqlQuery` annotation.

### DIFF
--- a/api/src/main/java/marquez/db/JobDao.java
+++ b/api/src/main/java/marquez/db/JobDao.java
@@ -231,6 +231,10 @@ public interface JobDao extends BaseDao {
     }
   }
 
+  /*
+   Note: following SQL never executes. There is database trigger on `jobs_view` that replaces following SQL
+   with rewrite_jobs_fqn_table plpgsql function. Code of that function is at R__1 migration file.
+  */
   @SqlQuery(
       """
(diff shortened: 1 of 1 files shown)


wslulciuc (approved):
LGTM 👍

api/src/main/java/marquez/db/JobDao.java
```diff
@@ -231,6 +231,10 @@ default PGobject toJson(Set<DatasetId> dataset, ObjectMapper mapper) {
     }
   }
 
+  /*
+   Note: following SQL never executes. There is database trigger on `jobs_view` that replaces following SQL
```
collado-mike, line 235: I might leave off that first sentence - it's a bit inaccurate. As far as the Jdbi client is concerned, this is the SQL that is executed. I might phrase it as 

> Inserts to the `jobs_view` are supported by the `rewrite_jobs_fqn_table` plpgsql function. Code of that function is at R__1 migration file.

wslulciuc: Ah sorry @collado-mike! I enabled auto merge

PR #2130, opened 2022-09-21
Add parentRun alias to LineageEvent RunFacet to support older OpenLin…

collado-mike:
### Problem

Prior to https://github.com/OpenLineage/OpenLineage/pull/1037 , all Airflow OpenLineage events reported their parent runs as `parentRun` rather than the spec-defined `parent` facet. This change adds an alias to support those older events.

Unfortunately, jackson seems to prefer the alias field name rather than the target (or maybe it's ordered on some condition I don't understand), so if an event defines both, the `parentRun` facet will be used. This is confirmed in a unit test. The OpenLineage integration does use the same value for both `parentRun` and `parent` facets, so this should not cause any issues.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased](https://github.com/MarquezProject/marquez/compare/0.26.0...HEAD)
+### Fixed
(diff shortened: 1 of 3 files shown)


fm100 (approved):
LGTM!

wslulciuc (approved):

PR #2144, opened 2022-09-26
Update insert job function to avoid joining on symlinks for jobs that have no symlinks

collado-mike:
### Problem

Typical marquez installations don't have a large number of new jobs being created on a regular basis. However, in some small number of installations, there can be a large number of new jobs being created, which executes the `rewrite_jobs_fqn_table` function each time, putting stress on the backing database. Most of the compute cost of this function is in computing the symlinks and aliases for jobs - even when the inserted job has no symlink.

Closes: #ISSUE-NUMBER

### Solution

Adding a check for the symlink field and offering a lower cost query in cases when no symlink is present (the norm) radically reduces the database compute load in Marquez installations that frequently create a large number of new jobs.
The following graph shows query count and latency and database CPU utilization under a test load of many new jobs being created. The test load was several days of real production OpenLineage events being replayed on a dev instance. To verify results, I ran the same test twice for both the old query and the new. Under heavy load, the job creation query causes database CPU utilization to climb to 100% and query latency climbs to as high as 2 seconds. Under the same load (I renamed all of the jobs in the database, so the same load shows up as new jobs that invoke the job creation query), the new query drives CPU utilization to around 30% and query latency is around 300 microseconds.

Note that the query latency in this graph is shown at log scale (right axis). Otherwise, the latency for the new query would be indistinguishable from 0.
Screen Shot 2022-09-26 at 8 41 11 AM

### Checklist

### DIFF
--- a/api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql
+++ b/api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql
@@ -21,13 +21,14 @@ SELECT f.uuid,
 FROM jobs_fqn f,
      jobs j
 WHERE j.uuid = f.uuid
-AND j.is_hidden IS FALSE;
+  AND j.is_hidden IS FALSE;
 
 
 CREATE OR REPLACE FUNCTION rewrite_jobs_fqn_table() RETURNS TRIGGER AS
 $$
 DECLARE
     job_uuid uuid;
+    job_updated_at timestamp with time zone;
     new_symlink_target_uuid uuid;
     old_symlink_target_uuid uuid;
     inserted_job jobs_view%rowtype;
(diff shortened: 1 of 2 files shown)


api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql
```diff
@@ -53,7 +54,7 @@ BEGIN
            COALESCE(NEW.parent_job_uuid::char(36), ''),
            false
     ON CONFLICT (name, namespace_uuid, parent_job_uuid_string)
-        DO UPDATE SET updated_at               = EXCLUDED.updated_at,
+        DO UPDATE SET updated_at               = now(),
```
wslulciuc, line 57: Using `now()` better represents the action on the row 👍

api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql
```diff
@@ -64,14 +65,47 @@ BEGIN
...
+
+    -- update the jobs_fqn table when inserting a new record
+    -- (NEW.uuid will equal the job_uuid when inserting a new record)
+    -- AND if the symlink target is null
+    -- Avoid constructing the symlinks and aliases, as that is expensive
+    IF TG_OP='INSERT'
+        AND NEW.uuid = job_uuid
+        AND NEW.symlink_target_uuid IS NULL
+        AND NEW.updated_at=job_updated_at THEN
+        RAISE LOG 'Inserting into jobs_fqn for new job % (%)', NEW.name, job_uuid;
```
wslulciuc, line 84: Can change to log level to `DEBUG`?

wslulciuc (approved):
Left some minor comments, otherwise thanks for the amazing write up and analysis accompanying the fix @collado-mike 💯 🥇

PR #2146, opened 2022-09-27
Display current run state for job node in lineage graph

wslulciuc:
### Problem

To view the current run state of a job, you have to select the job node in the lineage graph, then navigate the run info tab.

### Solution

This PR fills job nodes in the lineage graph with the latest run state; this PR also makes some minor changes to column names used to display dataset and job metadata.

Screen Shot 2022-09-28 at 3 44 45 PM

### Checklist

### DIFF
--- a/web/src/components/lineage/components/node/Node.tsx
+++ b/web/src/components/lineage/components/node/Node.tsx
@@ -43,10 +43,11 @@ type NodeProps = DispatchProps & OwnProps
 
 function runStateToNodeColor(run: Nullable<Run>) {
   switch (run?.state) {
-    case 'RUNNING': return theme.palette.primary.main;
+    case 'NEW': return theme.palette.secondary.main;
+    case 'RUNNING': return theme.palette.info.main;
     case 'COMPLETED': return theme.palette.primary.main;
     case 'FAILED': return theme.palette.error.main;
-    case 'ABORTED': return theme.palette.primary.main;
+    case 'ABORTED': return theme.palette.warning.main;
     default:
       break;
   }
--- a/web/src/components/lineage/components/node/Node.tsx
+++ b/web/src/components/lineage/components/node/Node.tsx
@@ -42,12 +42,17 @@ interface OwnProps {
 type NodeProps = DispatchProps & OwnProps
 
 function runStateToNodeColor(run: Nullable<Run>) {
-  switch (run?.state) {
-    case 'NEW': return theme.palette.secondary.main
-    case 'RUNNING': return theme.palette.info.main
-    case 'COMPLETED': return theme.palette.secondary.main
-    case 'FAILED': return theme.palette.error.main
-    case 'ABORTED': return theme.palette.warning.main
+  switch (run && run.state) {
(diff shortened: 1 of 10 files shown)


web/src/components/lineage/components/node/Node.tsx
```diff
@@ -62,11 +76,11 @@ class Node extends React.Component<NodeProps> {
...
               style={{ cursor: 'pointer' }}
               r={RADIUS}
-              fill={theme.palette.common.white}
+              fill={runStateToNodeColor(job.latestRun)}
               stroke={
                 selectedNode === node.label
-                  ? theme.palette.primary.main
-                  : theme.palette.secondary.main
+                  ? theme.palette.common.white
+                  : runStateToNodeColor(job.latestRun)
```
phixMe, line 83: You've got the case where you're selected and failing, you can just move all this logic up to your function above.

web/src/helpers/theme.ts
```diff
@@ -32,6 +32,12 @@ export const theme = createTheme({
     error: {
       main: '#ee7b7b'
     },
+    warning: {
+      main: '#7D7D7D'
+    },
```
phixMe, line 37: Typically, this would be an orange or yellow in a design system.

web/src/components/lineage/components/node/Node.tsx
phixMe: I think mixing both a ternary and a switch is too confusing.

web/src/helpers/theme.ts
wslulciuc: I was referencing the colors defined in [`helpers/runs.ts`](https://github.com/MarquezProject/marquez/blob/main/web/src/helpers/runs.ts). Also, maybe I should just use `runColorMap` defined in `runs.ts`?

web/src/components/lineage/components/node/Node.tsx
wslulciuc: Ok, good point. I'll clean things up 👍

mobuchowski:
Thanks for working on UI @wslulciuc ❤️

web/src/helpers/theme.ts
phixMe: Yeah, let's go with that.

web/src/components/lineage/components/node/Node.tsx
```diff
@@ -39,6 +41,18 @@ interface OwnProps {
...
 
+function runStateToNodeColor(run: Nullable<Run>) {
+  switch (run?.state) {
+    case 'NEW': return theme.palette.secondary.main;
+    case 'RUNNING': return theme.palette.info.main;
+    case 'COMPLETED': return theme.palette.primary.main;
+    case 'FAILED': return theme.palette.error.main;
+    case 'ABORTED': return theme.palette.warning.main;
+    default:
+      break;
```
phixMe, line 52: You'll probably want to return a color for a null run here.

phixMe (approved):
LGTM

PR #2147, opened 2022-09-28
cli: add commend to generate graph-based runs

mobuchowski:
This PR adds graph generation capabilities working with algorithm described below:

```
Generate DAGs in a "layered" structure. First, it split graph nodes into randomly
sized layers of monotonically growing node numbers. That means that node N will always be in
lower layer or the same as every node with number greater than N. Next, create jobs for each
node element that are in layers higher than 1. Job consists of one output node and up to
maxJobInputCount input nodes. The input nodes are chosen by randomly selecting numbers in layers
lower than current layer.
```

Load testing tool uses that structure later to generate job runs using `k6` load testing tool. 

### DIFF
--- a/api/src/main/java/marquez/MarquezApp.java
+++ b/api/src/main/java/marquez/MarquezApp.java
@@ -27,6 +27,7 @@ import javax.sql.DataSource;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import marquez.api.filter.JobRedirectFilter;
+import marquez.cli.GraphCommand;
 import marquez.cli.MetadataCommand;
 import marquez.cli.SeedCommand;
 import marquez.common.Utils;
@@ -79,6 +80,7 @@ public final class MarquezApp extends Application<MarquezConfig> {
 
     // Add CLI commands
     bootstrap.addCommand(new MetadataCommand());
+    bootstrap.addCommand(new GraphCommand());
     bootstrap.addCommand(new SeedCommand());
 
     bootstrap.getObjectMapper().disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
--- a/api/src/main/java/marquez/cli/GraphCommand.java
+++ b/api/src/main/java/marquez/cli/GraphCommand.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2018-2022 contributors to the Marquez project
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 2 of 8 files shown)


wslulciuc:
@mobuchowski, this is really cool! My initial thought was to expand on the [`metadata`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/cli/MetadataCommand.java) cmd to support generating lineage graphs. Have you considered extending the `metadata` cmd to also support graphs? My view is that we'd have a single cmd to generate run metadata that would by extension also generate lineage metadata (i.e. a graph as defined by the `graph` cmd being proposed). Currently, the `metadata` only produces runs with no dependencies, therefore you have `N` graphs with a single job node and `X` input / output datasets (which is that exciting).

For example, the usage for `metadata` is as follows:

```
$ java -jar marquez-api.jar metadata --help
       metadata [--runs RUNS] [--bytes-per-event BYTES-PER-EVENT] [-o OUTPUT] [-h]

generate random metadata using the OpenLineage standard

named arguments:
  --runs RUNS            limits OL runs up to N (default: 25)
  --bytes-per-event BYTES-PER-EVENT
                         size (in bytes) per OL event (default: 33404)
  -o OUTPUT, --output OUTPUT
                         the output metadata file (default: metadata.json)
```

with the following additional options:

```
  --inputs-per-event INPUTS-PER-EVENT
                         limits inputs per event to N (default: 4)
  --bytes-per-input BYTES-PER-INPUT
                         size (in bytes) per input (default: 4096)
  --outputs-per-event OUTPUTS-PER-EVENT
                         limits outputs per event to N (default: 2)
  --bytes-per-output BYTES-PER-OUTPUT
                         size (in bytes) per output (default: 4096)
```

The above options would cover most of the `graph` cmd features. To support graphs in `metadata`, we can add:

```
  --graph-levels GRAPH-LEVELS
                         limits the graph to N levels
  --graph-first-level  GRAPH-FIRST-LEVEL
                         limits the first level of the graph to N
```

This has the added benefit of working with a single cmd to generate metadata. Let me know what your thoughts!

wslulciuc:
@mobuchowski, any additional comments / commits? Wondering if we can move this PR out of `draft` into `review`?

mobuchowski:
hey @wslulciuc I hope I'll have time to pick this task back up in close future. I think I'd try different approach - generating this kind of data using Java seems to be very cumbersome.

wslulciuc:
Ok, sounds great!

PR #2152, opened 2022-09-30
Add OL facet tables

wslulciuc:
### Problem

To access OL facets, we have to run direct queries against OL events table, which has significant performance concerns.

Closes: #2076 

### Solution

Add tables for dataset, job, and run facets.

### Checklist


pawel-big-lebowski (commented):
This is going to be huge improvement in terms of performance. 
It will also help developing some features like getting raw events by namespace which @mobuchowski coped with some time ago (https://github.com/MarquezProject/marquez/pull/2070)

My concerns are: 
 * If my understanding is correct, current behavior leads users to losing dataset history (facets written the old way won't be read the new way). If this acceptable, at least we should put some doc notes. Otherwise, providing some migration (even the optional migration) would be great.
 * We're storing the same facets twice now: in `lineage_events` and new facets' tables. Is this an expected behavior?
 *  Wouldn't we like to have a test that verifies after posting lineage event that introduced tables are filled? I couldn't spot anything similar in the PR.  
 * Flyway DB migration version files numeration need to be adjusted (v46 seem to be outdated)

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -75,22 +75,22 @@ void updateLastModifiedAt(
...
+          + "    SELECT d.uuid, d.name, d.namespace_name, dv.run_uuid, dv.lifecycle_state, lineage_event_time, facet\n"
           + "    FROM selected_datasets d\n"
-          + "    INNER JOIN dataset_versions dv ON dv.uuid = d.current_version_uuid\n"
+          + "    INNER JOIN dataset_versions AS dv ON dv.uuid = d.current_version_uuid\n"
           + "    LEFT JOIN LATERAL (\n"
-          + "        SELECT run_uuid, event_time, event FROM lineage_events\n"
+          + "        SELECT run_uuid, lineage_event_time, facet FROM dataset_facets\n"
           + "        WHERE run_uuid = dv.run_uuid\n"
-          + "    ) e ON e.run_uuid = dv.run_uuid\n"
+          + "    ) df ON df.run_uuid = dv.run_uuid\n"
```
pawel-big-lebowski, line 84: this seems to be missing `dataset_uuid = dv.dataset_uuid`

api/src/test/java/marquez/db/migrations/V55_5__BackfillFacetsTest.java
```diff
@@ -0,0 +1,502 @@
...
+            new LineageEvent.JobLink("namespace", "name")),
+        ImmutableMap.of()
+            .of(
+                "custom-run-facet", "some-run-facet",
+                "spark.logicalPlan", "{some-spark-logical-plan:true}",
+                "errorMessage", "{some-error-message-facet:true}",
+                "spark_unknown", "unknown-facet-to-be-skipped-in-import"));
+  }
+
+  record JobFacet(
```
mobuchowski, line 473: `record` 👍

mobuchowski:
`metadata.json` - do we need this file? It shows one line changed, but also `1 addition, 0 deletions not shown because the diff is too large. Please use a local Git client to view these changes.`

api/src/main/java/marquez/db/migrations/V55_5__BackfillFacets.java
```diff
@@ -0,0 +1,282 @@
...
+              ==================================================
+              MARQUEZ INSTANCE TOO BIG TO RUN AUTO UPGRADE.
+              YOU NEED TO RUN v55_migrate COMMAND MANUALLY.
+              FOR MORE DETAILS, PLEASE REFER TO:
+              https://github.com/MarquezProject/marquez/blob/main/api/src/main/resources/marquez/db/migration/V55__readme.md
+              ==================================================
+              ==================================================
+              ==================================================
+              """);
+      // We end migration successfully although no data has been migrated to facet tables
```
mobuchowski, line 220: This will leave a lot of database calls that utilize new tables in a bad state. I think better option is to fail the migration - if you won't run the migration, using new version with this successful migration won't work.

Another way, if we really want to suceed this is to remove usage of new tables from db queries, and do them in a future release.

One more question - will running this migration manually even work if we succeed here? I think flyway will see the migration as completed.

pawel-big-lebowski: If you fail flyway, you will not have the empty tables created. A manual run calls a `migrate` method from `V55_5__BackfillFacets.java` class while not starting flyway context, so it is transparent to `flyway`. 

The other issue you raise is splitting the PR into two:
 * first create new tables and give users time to migrate,
 * make use of new tables once the migration is over.

It's add a bit more overhead to a process but may be useful for some users.

pawel-big-lebowski: @mobuchowski your concerns are valid but splitting operation into multiple releases is not a solution as users may run multiple upgrades at once at any point. 

The solution here would be to make code aware of asynchronous migration being finished or not, so that it returns data:
 * from `run_facets`, `job_facets`, `dataset_facets` when migration is over,
 * from `lineage_events` table when migrations is still in progress.

mobuchowski: That solution sounds good 👍

api/src/main/java/marquez/db/RunFacetsDao.java
```diff
@@ -0,0 +1,215 @@
...
+                    lineageEventTime,
+                    lineageEventType,
+                    Facet.PARENT.getName(),
+                    toPgObject(Facet.PARENT.asJson(parent))));
+
+    // Add ..
+    Optional.ofNullable(runFacet.getAdditionalFacets())
+        .ifPresent(
+            additional ->
+                additional.forEach(
```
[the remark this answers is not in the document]
pawel-big-lebowski: there are two inner lambdas as the code construct uses `ifPresentOrElse`.
Filtering above would not skip code execution as expected but let it execute in `orElse` lambda.

pawel-big-lebowski:
Database migration in a nutshell: 
 * Users with up to 100K `lineage_events` should not bother about this migration. 
 * Users with more than 100K `lineage_events` will need to run additional migration command. 
 * DB upgrade is millisecond long. This is because migration procedure creates views with job, run and dataset facets. When target facet tables are loaded, the views are pointing to them. Until that moment views are defined over `lineage_events` table. 
 * After the change facets are written into two locations: `lineage_events` table and `job`/`run`/`dataset` facets tables. 
 * Perf tests were run to verify data migration performance. See [results](https://github.com/MarquezProject/marquez/blob/feature/ol-facet-tables/api/src/main/resources/marquez/db/migration/V55__readme.md).

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -80,9 +80,9 @@ public interface RunDao extends BaseDao {
           + "FROM runs_view AS r\n"
           + "LEFT OUTER JOIN\n"
           + "(\n"
-          + "    SELECT le.run_uuid, JSON_AGG(event->'run'->'facets') AS facets\n"
-          + "    FROM lineage_events le\n"
-          + "    GROUP BY le.run_uuid\n"
+          + "    SELECT rf.run_uuid, JSON_AGG(rf.facet ORDER BY rf.lineage_event_time ASC) AS facets\n"
```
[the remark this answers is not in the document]
wslulciuc: Thanks @mobuchowski for the pointing this out (which, I agree with).

@pawel-big-lebowski I know this PR / optimization has been dragging (mostly my fault, so happy to take the blame), but splitting the changes into separate PRs / releases would reduce any challenges with rollbacks, but also provide users the option to decide when to apply the migration / upgrade. So, that said, what I propose is:

1. Open a PR to just introduce the dataset, job, and run facets tables (a write only mode), meaning the facet tables won't be referenced in anyway in our queries, but will be written to so that features can take advantage of the optimization; lineage events will continue to be written to the `lineage_events` table as well. This can be released in `0.30.0`.
2. As a follow up PR, we can update all references to `lineage_events` in our queries to use the individual facet tables. Within this PR, the migration script (and documentation) to backfill facets from the `lineage_events` table will also be introduced (and be **required** to run for users with larger `lineage_events` tables). This can be released in `0.31.0`.

PR #2152 can be **step 2** and marked ias blocked, and I can cherry pick any changes into a separate PR for **step 1** (also option to other options). Thoughts?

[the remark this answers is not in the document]
pawel-big-lebowski: It may be difficult to split the change into separate PRs. The initial PR was missing tests. I added them within my commits and it detected some issues with previous commits like:
 * `runFacetExists` method in `api/src/main/java/marquez/db/RunFacetsDao.java`
 * `fromName` method in `api/src/main/java/marquez/db/DatasetFacetsDao.java` 
 (see my latest commit)

So merging some of the commits shouldn't be a solution. Additionally, one cannot take benefit from facet tables unless they are filled with data. Otherwise, users will experience incomplete results. 

This PR does not clean or remove data from `lineage_events` table so we always have the ability to prepare a fix that restores previous mechanics if we want to. 

@wslulciuc Although splitting this PR into multiple PRs could improve readability, I don't see any advantage of splitting this into separate releases (`0.30.0` and `0.31.0`).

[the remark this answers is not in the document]
collado-mike: The combination of a manual migration process and the fact that the code stops reading from the `lineage_events` table introduces a real problem for users. When a user deploys, the tables will be created and all new records will be written into the new facet tables. However, all old data still lives only in the `lineage_events` table and won't be accessible until after the user executes the migration command. This means that the service will effectively be broken until users choose to execute this migration.

I think splitting the PR so that writes and reads are committed and can be deployed separately makes sense. This gives users the opportunity to push multiple deployments in a way that guarantees no downtime. It also makes this PR muuuch easier to review - readability is not just good for its own sake; it makes it easier to catch issues before they go to production. The migration script itself warrants its own review, I think.

pawel-big-lebowski:
Replying to @collado-mike 

> The combination of a manual migration process and the fact that the code stops reading from the lineage_events table introduces a real problem for users. When a user deploys, the tables will be created and all new records will be written into the new facet tables. However, all old data still lives only in the lineage_events table and won't be accessible until after the user executes the migration command.

I think there may be some kind of misunderstanding here. For the users with more than 100K lineage_events:
 * `run_facets`, `job_facets` and `dataset_facets` tables are created during the flyway migration, 
 *  Views `run_facets_view`, `job_facets_view` and `dataset_facets_view` are created during flyway migration and their definition is created on the top of `lineage_events` tables. Application code makes use of those views. 
 * Filling `run_facets`, `job_facets` and `dataset_facets` with data has to be done as manual asynchronous step but the application is still up and running before the command is called and executed. 
 * Once manual migration is finished, views `run_facets_view`, `job_facets_view` and `dataset_facets_view` are recreated and their definition is based on `run_facets`, `job_facets` and `dataset_facets` tables. 

>  This means that the service will effectively be broken until users choose to execute this migration.

No, it's not and that's why `run_facets_view`, `job_facets_view` and `dataset_facets_view` are created. They provide no-downtime experience at the time of manual migration procedure. 

> 
> I think splitting the PR so that writes and reads are committed and can be deployed separately makes sense. This gives users the opportunity to push multiple deployments in a way that guarantees no downtime. It also makes this PR muuuch easier to review - readability is not just good for its own sake; it makes it easier to catch issues before they go to production. The migration script itself warrants its own review, I think.

I agree that splitting PR into multiple PRs makes sense:
 * PR I - create `run_facets`, `job_facets` and `dataset_facets` tables, write to them for new OL events. don't read data from new tables. 
 * PR II - create `run_facets_view`, `job_facets_view` and `dataset_facets_view` on the top of `lineage_events` table, switch application logic to making use of the views. 
 * PR III -  migration command to fill `run_facets`, `job_facets` and `dataset_facets` tables. Replace the views from PR II to point to `run_facets`, `job_facets` and `dataset_facets` tables once migration is over. 

Releasing PR I, PR II and PR III in one go does not imply any downtime.

api/src/main/java/marquez/cli/V55MigrationCommand.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jdbi.v3.sqlobject.SqlObjectPlugin;
+
+/**
+ * A command to manually run V55 database migration. This migration requires a heavy DB operation
+ * which can be done asynchronously (with limited API downtime) due to separate migration command.
+ *
+ * <p>Please refer to @link marquez/db/migration/V55__readme.md for more details.
+ */
+@Slf4j
+public class V55MigrationCommand<MarquezConfig> extends EnvironmentCommand<marquez.MarquezConfig> {
```
wslulciuc, line 31: Is `V55MigrationCommand` too specific (but also vague)? We'll certainly have more migrations in the future and one-off commands for each will be hard to manage and not the best user experience. I suggest we name the class `DbMigrationsCommand` and define `v55_migrate` as a subcommand. This [tutorial](https://riptutorial.com/dropwizard/example/28864/subcommands) is a pretty good reference on how to define subcommands.

api/src/main/java/marquez/db/JobFacetsDao.java
```diff
@@ -0,0 +1,189 @@
...
+                insertJobFacet(
+                    UUID.randomUUID(),
+                    now,
+                    jobUuid,
+                    runUuid,
+                    lineageEventTime,
+                    lineageEventType,
+                    DatasetFacetsDao.Facet.DOCUMENTATION.getName(),
+                    toPgObject(DatasetFacetsDao.Facet.DOCUMENTATION.asJson(documentation))));
+
```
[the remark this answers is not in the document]
wslulciuc: I agree, it's a bit ugly. Do you mean defining an interface for each enum to implement?

api/src/main/java/marquez/db/migrations/V55_5__BackfillFacets.java
```diff
@@ -0,0 +1,346 @@
...
+          ORDER BY
+              COALESCE(events_chunk.created_at, events_chunk.event_time) ASC,
+              events_chunk.run_uuid ASC
+          LIMIT 1
+          RETURNING created_at, run_uuid;
+      """;
+
+  @Setter private Integer chunkSize = null;
+
+  @Setter private boolean triggeredByCommand = false;
```
wslulciuc, line 190: _Minor_: I'd use the flag name `manual`.

api/src/main/java/marquez/db/migrations/V55_5__BackfillFacets.java
```diff
@@ -0,0 +1,346 @@
...
+      return;
+    }
+    Optional<MigrationLock> lastExpectedLock = getLock(GET_FINISHING_LOCK_SQL);
+
+    // create facet views
+    execute(CREATE_RUN_FACETS_VIEW);
+    execute(CREATE_DATASET_FACETS_VIEW);
+    execute(CREATE_JOB_FACETS_VIEW);
+
+    if (!triggeredByCommand && countLineageEvents() >= BASIC_MIGRATION_LIMIT) {
```
wslulciuc, line 255: It's great that we inform the user on how to handle very large tables 💯

api/src/main/java/marquez/db/migrations/V55_5__BackfillFacets.java
```diff
@@ -0,0 +1,346 @@
...
+  public void migrate(Context context) throws Exception {
+    if (context != null) {
+      jdbi = Jdbi.create(context.getConnection());
+    }
+
+    /**
+     * Workaround to register uuid_generate_v4 function to generate uuids. gen_random_uuid() is
+     * available since Postgres 13
+     */
+    jdbi.withHandle(h -> h.createCall("CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\"").invoke());
```
wslulciuc, line 238: It might be possible that the migration might not be applied by a superuser? Can we the uuid using `java.util.UUID`, then inject the value in the inserts?

api/src/main/resources/marquez/db/migration/V55__readme.md
```diff
@@ -0,0 +1,54 @@
...
+Default `chunkSize` was used.
+
+Table below shows migration time for different amount of `lineage_events` generated the same way.
+
+| `lineage_events` | `job_facets` | `run_facets` | `dataset_facets` | time taken |
+|------------------|--------------|--------------|------------------|------------|
+| 10K events       | 50K rows     | 50K rows     | 150K rows        | 10sec      |
+| 100K events      | 500K rows    | 500K rows    | 150K rows        | 106sec     |
+| 500K events      | 2.5M rows    | 2.5M rows    | 7.5M rows        | 612sec     |
+| 1M events        | 5M rows      | 5M rows      | 15M rows         | 1473sec    |
```
wslulciuc, line 54: Thanks for this detailed and well throughout migration plan @pawel-big-lebowski 💯

wslulciuc:
> The combination of a manual migration process and the fact that the code stops reading from the lineage_events table introduces a real problem for users.

In **step 1** that I [outlined](https://github.com/MarquezProject/marquez/pull/2152#discussion_r1061327750) above, Marquez would continue to write to the `lineage_events` table (and simultaneously to the new facet tables) would  address this concern, which is a significant operational advantage. At Astronomer, we'd want this type of control to limit any downtime during the database migration. But, after doing another pass on the PR (and additional context from @pawel-big-lebowski [comment](https://github.com/MarquezProject/marquez/pull/2152#issuecomment-1371881993) above), _views_ are used to avoid such downtime and effectively provide the safe guard for users when (manually, or automatically) applying the migration.

@pawel-big-lebowski has made some great additions and has helped push this PR to it's current state (and close to being merged!). Based on your feedback @collado-mike and @pawel-big-lebowski's input, there's are enough safe guards to ensure the user has control of how the migration is applied (as outlined in [`V55__readme.md`](https://github.com/MarquezProject/marquez/blob/feature/ol-facet-tables/api/src/main/resources/marquez/db/migration/V55__readme.md)) while also ensuring Marquez can continue to handle writes / reads. To not completely stop progress, I feel addressing the comments in the PR and cleaning up a few sections of the code would avoid splitting this PR.

api/src/main/java/marquez/db/migrations/V55_5__BackfillFacets.java
pawel-big-lebowski: That's a hard one. So, we don't want Java to touch the data. It's good for performance when everything is done in SQLs and no data transfers between backend DB and Java application are required. 

Adding this extension is a workaround for PostgreSQL 12, that we still support. If we stopped supporting PostgreSQL 12, we could use built-in uuid function (https://www.postgresql.org/docs/current/functions-uuid.html) available since version 13. 

Otherwise we can add extra condition here to include extra extension only for PostgreSQL 12. Anyway, I think PostgreSQL 12 users will need a superuser.

wslulciuc: Ok, that's reasonable. Mind making a note in the migration doc on the superuser requirement?

api/src/main/java/marquez/db/JobFacetsDao.java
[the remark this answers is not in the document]
pawel-big-lebowski: I created separate PR #2350 which is a part of this one responsible for creating new tables, writing new events to new tables, while not reading data from new tables. 

I addressed this feedback remark in #2350.

api/src/main/java/marquez/db/DatasetFacetsDao.java
```diff
@@ -0,0 +1,284 @@
...
+                insertDatasetFacet(
+                    UUID.randomUUID(),
+                    now,
+                    datasetUuid,
+                    runUuid,
+                    lineageEventTime,
+                    lineageEventType,
+                    Facet.DOCUMENTATION.getType(),
+                    Facet.DOCUMENTATION.getName(),
+                    toPgObject(Facet.DOCUMENTATION.asJson(documentation))));
```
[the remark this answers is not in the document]
pawel-big-lebowski: I created separate PR https://github.com/MarquezProject/marquez/pull/2350 which is a part of this one responsible for creating new tables, writing new events to new tables, while not reading data from new tables. 

I got rid of the repeated code block in #2350

api/src/main/java/marquez/MarquezApp.java
```diff
@@ -149,6 +150,12 @@ public void registerResources(
     }
   }
 
+  @Override
+  protected void addDefaultCommands(Bootstrap<MarquezConfig> bootstrap) {
```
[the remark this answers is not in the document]
pawel-big-lebowski: The existing commands are kind of no-arg commands. Migration commands needs to be able to access database connection. That's why I included it here so that it can get `Application` as a constructor param. Within `initialize` commands are added although application is not created yet.

pawel-big-lebowski:
The content of this PR has been split to:
 * https://github.com/MarquezProject/marquez/pull/2350
 * https://github.com/MarquezProject/marquez/pull/2355
 * https://github.com/MarquezProject/marquez/pull/2359

Review comments provided in this PR have been applied there.

Issue #2071, opened 2022-08-12
Proposal: Update DatasetVersion versioning


RNHTTR:
There has been some discussion (mostly in #1977) about reworking the versioning system for `DatasetVersion`. 

### Motivation
The current `DatasetVersion` versioning system leads to confusion (e.g. #1883). `DatasetVersion` has a `uuid` field (of type `UUID`) and a `version` field (also of type `UUID`). In a practical sense, I think these fields are redundant.

Additionally, external data systems might already support dataset versioning (e.g. delta, iceberg). It'd make sense for Marquez to support these.

### Proposal
I propose that a `Version`'s `uuid` field should assume the functionality currently provided by `Version`'s `version` field, and add an additional field `external_version` to support dataset versions provided by external applications. This would have a downstream impact on `JobVersion`.

#### Work required
1. Update [Version.getValue()](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/common/models/Version.java#L27) to be of type `String`
2. Drop `DatasetVersion`'s [`version` field](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/models/ExtendedDatasetVersionRow.java#L26)
3. Add a field to `DatasetVersion`: `external_version` (`String`)
4. Drop `JobVersion`'s [`version` field](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/models/ExtendedJobVersionRow.java#L32)
5. Add a field to `JobVersion`: `external_version` (`String`). 
    1. I'm not sure if this is currently necessary, but it seems reasonable to assume that data applications might support job versions tied to code in the future if they don't already.
7. Use OpenLineage's [`DatasetVersionDatasetFacet` facet](https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/DatasetVersionDatasetFacet.json) to support external dataset versions.
8. Upstream/downstream code changes to support 1-6 (e.g. updating queries to use `dv.uuid` instead of `dv.version`)
9. Database migrations

If this proposal is accepted, I'll open an [official proposal](https://github.com/MarquezProject/marquez/tree/main/proposals#submitting-a-proposal-for-review).

collado-mike:
What's the reasoning behind steps 2 and 3? I can see the usefulness of the `external_version` column to note specifically that the dataset version is assigned by some other system. But what column do we use when we compute the dataset version? Will we still write to the `external_version` column?

[...]

PR #2153, opened 2022-09-30
Proposal/2071 update version versioning

RNHTTR:
### Overview

This PR formally introduces [Proposal: Update DatasetVersion versioning #2071](https://github.com/MarquezProject/marquez/issues/2071)

### Checklist


RNHTTR:
I could use some help with a couple things here:

1. Confirming the proposed db changes are backward compatible
2. Added details for implementing the facet. I assume it'll be part of the DatasetVersionDAO, but I'm not entirely sure.

RNHTTR:
FYI @collado-mike @wslulciuc

proposals/2071-update-versioning.md
```diff
@@ -0,0 +1,42 @@
...
+
+Additionally, external data systems might already support dataset versioning (e.g. delta, iceberg). It'd make sense for Marquez to be able support these external versions.
+
+## Proposal
+This proposal aims to simplify the Marquez versioning system and enable new features to support dataset versions supplied by external data systems.
+
+Moving forward, `DatasetVersionRow` and its cousin [`ExtendedDatasetVersionRow`](https://github.com/MarquezProject/marquez/blob/0.26.0/api/src/main/java/marquez/db/models/ExtendedDatasetVersionRow.java#L26) will no longer support a `version` field; rather, this will be dropped and replaced by a nullable `externalVersion`. This `externalVersion` field will be populated if provided, using [OpenLineage's DatasetVersionDatasetFacet facet](https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/DatasetVersionDatasetFacet.json) to support external dataset versions. `DatasetVersionRow`'s `uuid` field will serve as the version's sole identifier.
```
collado-mike, line 21: Can you comment on how the _public_ API is impacted (if at all)?

RNHTTR: Not from a user's perspective as far as I can tell, although there is another `JobVersion` in the public API that will need the `version` field removed. I'll also need to update the various versions in the clients. I'll add all of these to the list of TODOs in the implementation section.

RNHTTR: Actually, I guess, while the public API will be backward compatible, we'll to expose `externalVersion` to the public API as well.

collado-mike: I was thinking specifically about how the URLs for fetching DatasetVersions will work and how we return references to the DatasetVersion. E.g., [this structure](https://github.com/MarquezProject/marquez/blob/main/spec/openapi.yml#L1186-L1202) in the spec references a UUID version, but I don't know if that is going to change or if it will just reference the DatasetVersion's primary key. So what is returned in the [`Run` spec](https://github.com/MarquezProject/marquez/blob/main/spec/openapi.yml#L1655-L1689)'s `inputDatasetVersions` field?
Additionally, the [`getDatasetVersion` spec](https://github.com/MarquezProject/marquez/blob/main/spec/openapi.yml#L201-L218) accepts a `version` parameter. Is that the UUID? or the `externalVersion`?
Personally, I prefer the `externalVersion`, if present, as it introduces compatibility between systems - e.g., a Snowflake dashboard can just _know_ how to construct a link to the Marquez UI for a given dataset@version because the standard is known and consistent.

wslulciuc: @collado-mike: very good points on backwards compatibility. To ensure we don't introduce any breaking changes, Marquez will continue to reference the `uuid` of `dataset_versions` and set that value accordingly. That is, in the case of `Datasets.currentVersion`, the value will continue to be of type `uuid`, but with the addition of `Datasets.currentExternalVersion` (we can think of this as an alias). More specifically, for `DatasetVersionId`, we will add `Datasets.externalVersion` which can be `null`, but `Datasets.version` will always be set.

I think retrieving a dataset version is less obvious, but we can update the endpoint:

```
GET /namespaces/{namespace}/datasets/{dataset}/versions/{version}
```

where `{version}` can either be a `uuid` or a string type. The parsing on the backend really isn't that complicated and be easy enough to do.

Another option would be we expose a new endpoint, but that seems unnecessary and would prefer being flexible on `{version}` type:

```
GET /namespaces/{namespace}/datasets/{dataset}/versions/{version}/external
```

wslulciuc (approved):
Great proposal, @RNHTTR! I'm going to merge as this PR has been around for a while (really my fault here 😅). I've responded with my thoughts on backwards compatibility, we can follow up, but feel what you outlined for versioning is in a state where someone can pick up the work.

Issue #2095, opened 2022-08-30
api: add possibility to delete NAMESPACES


mobuchowski:
Building on top of possibility added in https://github.com/MarquezProject/marquez/pull/2032 - add endpoint to soft delete everything in particular dataset - both jobs and datasets.

mobuchowski:
Few thoughts about deleting namespaces:

Deleting namespace can be thought as a two operations from UI side. 

- All jobs and datasets belonging to the namespace are deleted.
- After delete we can’t see the deleted namespace in Marquez namespace dropdown.

That leaves us with few challenges if we want to have both.

- How would deleted namespace interact with receiving new event for the same namespace?
    - Should the namespace be back in namespace list?
        - This would require event to check and clear potential flag on namespaces.
    - Should we “undelete” the namespace with all the resources?
        - Naive `is_hidden` flag would work exactly like that. Maybe better idea is having all `is_hidden` flags - on jobs, datasets and namespaces set. If event clears namespace flag, it does not clear all flags for jobs and namespaces.
        - Alternative to flags could be logic that shows on the UI only namespaces with some existing jobs or datasets - checking count on every `GET /namespace` call. That is not consistent with `default` namespace though.
    - Should we only show the jobs/datasets that were created by events received after deletion?
        - In this case, should there be any undelete logic for the whole namespace?
- Should behavior of deleted namespace, and namespace from which all the datasets and jobs were deleted be different?
- Theoretically simplest approach, hard delete namespaces would require multiple cascades - everything references namespaces.

howardyoo:
My `two-cents`:
- Should the namespace be back in namespace list? --> Yes
- Should we “undelete” the namespace with all the resources --> Yes
- Should we only show the jobs/datasets that were created by events received after deletion? --> Yes

mobuchowski:
>Should we “undelete” the namespace with all the resources --> Yes
>Should we only show the jobs/datasets that were created by events received after deletion? --> Yes

But those conflict 😉

julienledem:
I agree that the namespace should be marked soft deleted as well. It does not show up in the list of existing namespaces but we can retrieve it.

I think we should adopt the following behavior: Anywhere we would have created a resource (Job, Dataset, Namespace) that didn't exist yet and there was previously an existing but (soft) deleted one we should just undelete it and update it.
Example: 
- when receiving an OL event for a job or dataset that don't exist yet.
[...]

PR #2172, opened 2022-10-06
delete: add delete namespace endpoint

mobuchowski:
Add option to hide/soft delete namespace. 

When namespace is hidden, all datasets and jobs contained by this namespace are also hidden. Namespace stops being shown on UI - frontend filters it by `isHidden` attribute, instead of filtering on backend. Reason for doing it this way is huge potential complexity of changing namespace contract - for example, a lot of methods check whether namespace exists before performing.

Deleted namespace is being undeleted when relevant OpenLineage event is received. This does not automatically undelete all the datasets and jobs in the namespace, only those that are received in this event.

Closes: #2095


api/src/test/java/marquez/DatasetIntegrationTest.java
```diff
@@ -408,36 +383,118 @@ public void testApp_showsDeletedDatasetAfterReceivingNewVersion() throws IOExcep
...
+            "the_producer");
+
+    LineageEvent secondEvent =
+        new LineageEvent(
+            "COMPLETE",
+            Instant.now().atZone(ZoneId.systemDefault()),
+            new LineageEvent.Run(UUID.randomUUID().toString(), null),
+            new LineageEvent.Job(namespaceName, "second_job_name", null),
+            List.of(
+                new LineageEvent.Dataset(namespaceName, name, LineageTestUtils.newDatasetFacet())),
```
pawel-big-lebowski, line 446: Isn't it the same dataset as before?
If so, test cannot return two datasets regardless of delete namespace behavior.

api/src/test/java/marquez/db/DatasetDaoTest.java
```diff
@@ -415,6 +415,35 @@ public void testGetDatasets() {
...
+            new Dataset(
+                NAMESPACE,
+                DATASET + "2",
+                LineageEvent.DatasetFacets.builder()
+                    .lifecycleStateChange(
+                        new LineageEvent.LifecycleStateChangeFacet(
+                            PRODUCER_URL, SCHEMA_URL, "DROP"))
+                    .build())));
+
+    datasetDao.deleteByNamespaceName(NAMESPACE);
```
pawel-big-lebowski, line 444: Assertion is missing, right?

Issue #2183, opened 2022-10-12
ERROR: cannot drop columns from view from marquez/db/migration/R__3_Datasets_view.sql


Esaych:
# Issue
With existing local database instance, starting up my lineage backend service throws this error. Work around for me is to delete my database and recreate it. Ideally this shouldn't need to happen.

```
! org.postgresql.util.PSQLException: ERROR: cannot drop columns from view
! at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2676)
! at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2366)
! at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:356)
! at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:496)
! at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:413)
! at org.postgresql.jdbc.PgStatement.executeWithFlags(PgStatement.java:333)
! at org.postgresql.jdbc.PgStatement.executeCachedSql(PgStatement.java:319)
! at org.postgresql.jdbc.PgStatement.executeWithFlags(PgStatement.java:295)
! at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:290)
! at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
! at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
! at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
! at java.base/java.lang.reflect.Method.invoke(Method.java:568)
! at org.apache.tomcat.jdbc.pool.StatementFacade$StatementProxy.invoke(StatementFacade.java:118)
! at jdk.proxy2/jdk.proxy2.$Proxy46.execute(Unknown Source)
! at org.flywaydb.core.internal.jdbc.JdbcTemplate.executeStatement(JdbcTemplate.java:201)
! at org.flywaydb.core.internal.sqlscript.ParsedSqlStatement.execute(ParsedSqlStatement.java:95)
! at org.flywaydb.core.internal.sqlscript.DefaultSqlScriptExecutor.executeStatement(DefaultSqlScriptExecutor.java:210)
! ... 28 common frames omitted
[...]

PR #2184, opened 2022-10-13
drop and create datasets_view

pawel-big-lebowski:

### Problem

Change within `datasets_view` definition breaks upgrades.

Closes: #2183

### Solution

Drop view each time migration is run

### Checklist

- [] Your changes are accompanied by tests (_if relevant_)

### DIFF
--- a/api/src/main/resources/marquez/db/migration/R__3_Datasets_view.sql
+++ b/api/src/main/resources/marquez/db/migration/R__3_Datasets_view.sql
@@ -1,4 +1,5 @@
-CREATE OR REPLACE VIEW datasets_view
+DROP VIEW IF EXISTS datasets_view;
+CREATE VIEW datasets_view
 AS
 SELECT d.uuid,
        d.type,


mobuchowski (approved):

collado-mike:
It should be a development principle that we don't make backward-incompatible changes. Dropping/renaming columns is fine in a single instance of an app where upgrades mean taking down the application and restarting it, but in a distributed environment, upgrading should not require downtime. As is, a rolling upgrade will cause all other running hosts to begin failing requests as they will be looking for a column that no longer exists in the central database.

Dropping/renaming columns should always be a two-step process
1. Remove/update all references to the column in the code
2. Remove/rename the column in the database

These two steps should be done in different releases so that users have an opportunity to upgrade in a controller fashion without downtime.

pawel-big-lebowski:
@collado-mike Not clear to me how the problem you described applies for this PR.

The issue is related to a logical view modification and adding new column `dataset_symlinks` to it ([former change](https://github.com/MarquezProject/marquez/pull/2087/files#diff-5b333ecf98961ba5fda4781ce267c917c1eac46df405597d9225ceb163dc2ec9)). Adding a new column into a view should be backward compatible and old-code readers should be able to read data properly. 

It turned out that `CREATE OR REPLACE` is not able to add a column to existing view and solution to this is to drop the view and re-create afterwards. Theoretically, this introduces millisecond time window when a view does not exist and other readers can fail (the moment after drop view and before create view) but that shouldn't be a big deal within distributed environment where single failures are allowed to happen.

collado-mike:
Ah, sorry. I thought the previous view change was dropping a column and that this was attempting to band-aid over that by dropping the view and recreating it. Yeah, I'm not sure why postgres doesn't detect that the previous columns are still there - order should be irrelevant to the view definition. I think adding the new column to the end of the view definition would fix the issue.

> but that shouldn't be a big deal within distributed environment where single failures are allowed to happen.

I think our assumptions of what happens in distributed systems differ 🤔. Distributed systems should be _resilient_ to failures. So, if the view is missing and the code retries the query and succeeds, that is resilient to single failures. But that's not what Marquez does right now. Currently, if the query fails, we'll fail the whole request. If it happens to fail during an OpenLineage event processing, we'll lose the whole event. The OL clients don't have any retry built-in, so the data is simply lost. Of course, these issues can be addressed with retries on the client or persistent queues on the backend, but the simpler fix is to make sure view definition changes are backward compatible whenever possible.

pawel-big-lebowski:
Thanks for providing an explanation on how Marquez is operating right now. It's really helpful. According to the [docs](https://flywaydb.org/documentation/concepts/migrations#transactions):

> By default, Flyway always wraps the execution of an entire migration within a single transaction.

So there should be no moment of a missing view visible to other readers.
 
I added `dataset_symlinks` column in the middle of view definition to locate it just after existing `name` column. I can change it to make it the last, however it's unclear to me if this can be any problem, as Marquez makes use of util methods like `marquez.db.Columns#stringOrNull` which do `results.getObject(column)`. These methods rely on column name instead of column ordering. 

The root cause issue was not according to column ordering but because we were trying to replace the view with a view containing different amount of columns which is not allowed in PostgreSQL.

Issue #2117, opened 2022-09-13
Point in time lineage proposal


pawel-big-lebowski:
Extend column lineage endpoint to be able to show how the graph looked like for a specified timestamp or dataset version. 

See proposal for more details:
https://github.com/MarquezProject/marquez/blob/main/proposals/2045-column-lineage-endpoint.md

howardyoo:
Do we currently have this similar feature for dataset also? Just checking...

pawel-big-lebowski:
@howardyoo I think we don't but we wanted to add one. Perhaps this deserves another `lineage over time` epic.

PR #2193, opened 2022-10-17
point in time proposal

pawel-big-lebowski:

### Problem

Proposal on extending Marquez API to be able to request historical data.

Part of : #2117

### Solution

### Checklist

### DIFF
--- a/proposals/2117-marquez-over-time.md
+++ b/proposals/2117-marquez-over-time.md
@@ -1,0 +1,39 @@
+# Proposal: Run-level Lineage in Marquez API
+
+
+Created: 2022-11-22
+
+## Problem
+
+Marquez data model in PostgresSQL allows extracting a snapshot of any lineage content from the past. This may be extremely
+useful to track how job, dataset etc. evolved over time. We do already support job and dataset versioning that return requested versions. We want to extend this feature to support `lineage` and `column-lineage` versioning which we call Run-Level Lineage. The purpose of this doc is to agree on API implementation approach. 
+
+## Solution
+
+Point in time can be described by a datetime, a dataset version, a job version or  run identifier. Versions are the only reliable parameter as datetime is ambiguous. For example: having different START and COMPLETE datetimes for each run event. We assume the ability to scan over time can be implemented within UI. 
+
+We will extend defintion of `nodeId` to contain an optional version of the node: 
+
+| **ID**       | **`dataset:{namespace}:{dataset}@{version}`**                     |
+|:-------------|:------------------------------------------------------------------|
+| **Example**  | **`dataset:food_delivery:public.top_delivery_times@947c0388..`**  |
+
+| **ID**       | **`job:{namespace}:{job}@{version}`**                             |
+|:-------------|:------------------------------------------------------------------|
+| **Example**  | **`job:food_delivery:orders_popular_day_of_week@947c0388..`**     |
+
+| **ID**       | **`run:{id}`**                                                    |
+|:-------------|:------------------------------------------------------------------|
+| **Example**  | **`run:a03422cf..`**    
+
+Run-level lineage will be implemented for the two lineage endpoints:
+ * `/api/v1/namespaces/some-namespace/lineage?nodeId=nodeId=dataset:food_delivery:public.delivery_7_days&@5ca3b37e-4e18-11ed-bdc3-0242ac120002`
+ * `/api/v1/namespaces/some-namespace/column-lineage?nodeId=nodeId=datasetField:db1:table1:a@5ca3b37e-4e18-11ed-bdc3-0242ac120002`
+
+
+Existing API endpoint will use extended `nodeId`. In case of `nodeId` having a version specified, returned graph nodes will also have it. If `nodeId` in query parameter is missing version, `nodeId`s of returned graph will NOT contain version so that no change is done to current behaviour of currently available features. 
+
+Future steps
+ 
+ * Implementing Run-Level Lineage may require database model changes in Marquez and this deserves a separate proposal or design document which is out of scope of this doc. 
+ * Run-Level column lineage can be implemented within issue #2262.


proposals/2117-marquez-over-time.md
```diff
@@ -0,0 +1,34 @@
...
+
+* *Shall a response contain a point in time marker?* This would have been helpful but can be difficult to incorporate
+into existing endpoints, especially if the root node of the JSON response is an array. 
+
+* *How to pass point in time parameter in URL?* Parameter is an url encoded string representation of `ZonedDateTime.
+
+* *Can we rely on `createdAt` columns of foreign tables?* I think we should be able to do so but it's worth checking 
+if this is the case now. In other words, `createdAt` column for a dataset field should be the same as `createdAt` for a dataset 
+and run tables. It cannot be evaluated as `Instant.now()` as this can lead to errors: returning dataset at snapshot
+can miss dataset field, as it considers it be younger.
```
mobuchowski, line 34: I think `createdAt` should be just a shortcut to not have to provide perfect UUID for initial dataset or job that we're looking at. The lineage then should be retrieved by explicit foreign key relations instead of relying on time correlation.

proposals/2117-marquez-over-time.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Problem
+
+Marquez data model in PostgresSQL allows extracting a snapshot of any lineage content from the past. This may be extremely
+useful to track how job, dataset etc. evolved over time. The feature can be implemented within each Marquez endpoint: like
+retrieving datasets or jobs. The purpose of this doc is to agree on a consistent API implementation approach. 
+
+## Solution
+
+Point in time can be described by a datetime, a dataset version or a job version. In case of dataset version or job version, an API should find out
```
collado-mike, line 11: IMO, versions are the only reliable parameter for a point in time lineage API. I think a good UI might give users the option to scan versions by some time condition (e.g., I want to diagnose the job run that executed yesterday at 12pm - tell me what job version was executed and what dataset version was read). Datetime as a parameter is ambiguous. If a datetime falls between two dataset versions, which one should be picked? Does the parameter have to match the createdAt field exactly? If so, why not just use the version?

proposals/2117-marquez-over-time.md
```diff
@@ -0,0 +1,34 @@
...
+on the entries older that. 
+
+Examples:
+
+ * `/api/v1/namespaces/some-namespace/datasets/some-dataset?snapshotAt=datetime:2011-12-03T10%3A15%3A30%2B01%3A00` which is an url encoded version of `ZonedDateTime` `2011-12-03T10:15:30+01:00`
+ * `/api/v1/namespaces/some-namespace/datasets/some-dataset?snapshotAt=dataset_version:5ca3b37e-4e18-11ed-bdc3-0242ac120002`
+ * `/api/v1/namespaces/some-namespace/jobs/some-job?snapshotAt=job_version:5ca3b37e-4e18-11ed-bdc3-0242ac120002`
+ 
+Please mind that requested version id does not have to be an id of the requested dataset. A user debugging issue 
+with `dataset_x` at version `ver_x`, can request `dataset_y` snapshot at `ver_x`. The request should be still valid.
```
collado-mike, line 22: This only works if the dataset version is a strict UUID. In https://github.com/MarquezProject/marquez/pull/2153 , we're proposing using external version identifiers, which may not be globally unique.

proposals/2117-marquez-over-time.md
collado-mike: A runId is also a good input parameter for point-in-time lineage, as a run is always tied to a specific job version and dataset version (at least, there's only ever one version for a given output dataset in a run).

mobuchowski: >If a datetime falls between two dataset versions, which one should be picked? 

I think here the choice is for version that exists - meaning older one. However, I agree that having support from UI side for this is superior choice.

pawel-big-lebowski: ok, let's restrict the API to version parameters.

proposals/2117-marquez-over-time.md
pawel-big-lebowski: ok, let's modify the assumption and remove such feature.

proposals/2117-marquez-over-time.md
pawel-big-lebowski: ok @mobuchowski , your opinion aligns with @collado-mike. Let's go that way.

proposals/2117-marquez-over-time.md
```diff
@@ -0,0 +1,29 @@
...
+only reliable parameter for a point in time lineage API. Datetime can be ambiguous, for instance by having START and COMPLETE
+datetime for each job.
+
+We assume the ability to scan over time can be implemented within UI. 
+
+We introduce `snapshotAt` URL parametere and provide below some example values of it:
+
+ * `/api/v1/namespaces/some-namespace/datasets/some-dataset?snapshotAt=dataset_version:5ca3b37e-4e18-11ed-bdc3-0242ac120002`
+ * `/api/v1/namespaces/some-namespace/jobs/some-job?snapshotAt=job_version:5ca3b37e-4e18-11ed-bdc3-0242ac120002`
+ * `/api/v1/namespaces/some-namespace/jobs/some-job?snapshotAt=run_id:5ca3b37e-4e18-11ed-bdc3-0242ac120002`
```
collado-mike, line 21: How are these different from the `/versions` variant of the APIs? 

https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/JobResource.java#L111-L131
https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/DatasetResource.java#L101-L121

proposals/2117-marquez-over-time.md
```diff
@@ -0,0 +1,29 @@
+# Proposal: Marquez API to retrieve historical data
+
+## Problem
+
+Marquez data model in PostgresSQL allows extracting a snapshot of any lineage content from the past. This may be extremely
```
collado-mike, line 5: Is the intention here to return _lineage_ for a given point in time? Or just the dataset/job definitions as they were at a particular version? If the latter, don't we already have that? If the former, do you intend to store the lineage information differently? The current storage model is sufficient to return point-in-time lineage, but it's really slow if looking at anything but the latest version.

julienledem (approved):

julienledem (commented):
even though I have approved this, you should resolve @collado-mike ' s comments first.  (sorry mike I submitted before seeing your comments)

proposals/2117-marquez-over-time.md
pawel-big-lebowski: The intention of this PR is to make consensus on how do we expose point in time endpoints, including lineage endpoints as well. Storage model, and performance related it, is specific to each endpoint and it's outside the scope of this PR. You're right that it may require some remodeling but I think this should be discussed one by one when implementing specific endpoint for point in time. 

In other words: this PR shows how API for lineage over time should look like, but not how to implement it.

proposals/2117-marquez-over-time.md
pawel-big-lebowski: `versions` variant of the APIs are implemented for `Dataset` and `Job` and this approach does not seem to be extendable to `lineage` or `column-lineage` endpoints. It makes sense to ask for lineage at specific `run_id` or lineage of a specific `dataset_version`. Lineage can be versioned by multiple params like tracking how the lineage looked like for different dataset versions or how did it look like for certain run_ids.

collado-mike: Ok, then if this is particular to the `/lineage` and `/column-lineage` APIs, let's update the URLs here to point to those APIs rather than the `/datasets` and `/jobs` APIs. I don't think we need to change the existing job/dataset APIs, as they're already being used.

Currently, the `/lineage` and `/column-lineage` APIs accept a `NodeId` as their argument. If asking for a `run`, we shouldn't need any extra parameters, as the point-in-time parameter is inferred. For the `job` and `dataset` nodes, can we simply pass in a `version` parameter? Whether by modifying the node id (e.g., `job:abc@version` or something similar) or by passing in a query parameter.

proposals/2117-marquez-over-time.md
collado-mike: I'm fine with agreeing on the API changes before tackling the storage, but I do think there ought to be a design doc tackling the storage issue up front. We're really only talking about the `/lineage` and `/column-lineage` APIs here, so it's really a question of how to store point-in-time lineage in a scalable way.

wslulciuc: I agree with @collado-mike, to get a snapshot of a lineage graph the caller _should_ invoke `GET` `/lineage` with a `nodeId` (see proposed `nodeId`s below). A lineage graph is modified within the context of a run. We can call these incremental changes to the graph at a given point in time via a some run `R` as a  **run-level** graph (=lineage graph snapshot). A run-level graph represents the relationships between dataset, job, and run metadata at a given point in time.

### Graph Data Model

**A run-level graph consists of the following nodes:**

* **Dataset Version:** A read-only _immutable_ version of a dataset.
* **Job Version:** A read-only _immutable_ version of a job, with a unique referenceable link to code preserving the reproducibility of builds from source.
* **Run:** A discrete instantiation of a job version, with a unique run ID used to update each stage of execution.

### Nodes

| **ID**       | **`dataset:{namespace}:{dataset}#{version}`**                     |
|:-------------|:------------------------------------------------------------------|
| **Example**  | **`dataset:food_delivery:public.top_delivery_times#947c0388..`**  |

| **ID**       | **`job:{namespace}:{job}#{version}`**                             |
|:-------------|:------------------------------------------------------------------|
| **Example**  | **`job:food_delivery:orders_popular_day_of_week#947c0388..`**     |

| **ID**       | **`run:{id}`**                                                    |
|:-------------|:------------------------------------------------------------------|
| **Example**  | **`run:a03422cf..`**

pawel-big-lebowski:
@wslulciuc @collado-mike I think you're right. Thank you. 

To sum up: 
 * the proposal refers only to modifying `lineage` and `column-lineage` endpoints,
 * `nodeId` will be extended to contain `version`.

mobuchowski:
Looks good to me @pawel-big-lebowski.

mobuchowski (approved):

wslulciuc (approved):
LGTM 👍

PR #2211, opened 2022-10-24
optimizing current runs query for lieage api 

prachim-collab:
### Problem

Introduce a simpler alternate getCurrentRuns query that gets only simple runs from DB without the additional data from tables such as run_args, job_context, facets, input/output versions etc which required the extra table joins in the old getCurrentRuns query. This new getCurrentRuns DAO is NOT being used in Marquez as of now.

Closes: [#4425](https://github.com/astronomer/astro/issues/4425)

### Solution

- The old `getCurrentRuns` DAO is renamed to [`getCurrentRunsWithFacets`](https://github.com/MarquezProject/marquez/pull/2211/files#diff-956145b0ebae24e2767b33a19fc3d24fc56b0ff6c4c3c1bc9428c50c3ccbc03eR132) without any change to the sql query .
- This new light weight query is call `getCurrentRuns` and is NOT called from /lineage api as of now so NO change is required to the /lineage api response spec.
- A  flag `withRunFacets` is also introduced as parameter to lineage api, which is always set to `true` to call `getCurrentRunsWithFacets`, so that /lineage api and hence all the tests still call the old DAO .

### Checklist


api/src/test/java/marquez/service/LineageServiceTest.java
```diff
@@ -268,7 +268,7 @@ public void testLineageWithDeletedDataset() {
         .hasSize(0);
     runAssert
         .extracting(Run::getOutputVersions, InstanceOfAssertFactories.list(DatasetVersionId.class))
-        .hasSize(1);
+        .hasSize(0);
```
phixMe, line 271: I think this is a little misleading since we're not asking the database for this information anymore. Should we just remove this assertion or change the type definition?

prachim-collab: I agree and had initially removed this completely but then added it back again, because getInputVersions also had an assertion. I guess i should remove both of them now.

api/src/test/java/marquez/service/LineageServiceTest.java
```diff
@@ -158,9 +158,6 @@ public void testLineage() {
     runAssert
         .extracting(Run::getInputVersions, InstanceOfAssertFactories.list(DatasetVersionId.class))
         .hasSize(0);
-    runAssert
-        .extracting(Run::getOutputVersions, InstanceOfAssertFactories.list(DatasetVersionId.class))
-        .hasSize(1);
```
fm100, line 163: We are asserting on the `Lineage` object returned from `lineageService.lineage` call, and it is the response payload of GET lineage API. Does this mean that API response is being changed?

collado-mike: Yeah, I don't like all the information being returned here, but breaking API compatibility is not good. If we want a lighter-weight version of the lineage API, I think it's better to either include an optional parameter to exclude the superfluous data or to create a new API and deprecate the old one

prachim-collab: I have updated the PR and added a `withRunFacets` flag to get runs with all the superfluous data and this flag always true in Marquez apis, so no affect on the API compatibility.

api/src/main/java/marquez/service/LineageService.java
```diff
@@ -46,7 +46,7 @@ public LineageService(LineageDao delegate, JobDao jobDao) {
     this.jobDao = jobDao;
   }
 
-  public Lineage lineage(NodeId nodeId, int depth) {
+  public Lineage lineage(NodeId nodeId, int depth, boolean withRunFacets) {
```
collado-mike, line 49: This is a good change, but I worry that we'll want to add more options to this method (e.g., include job facets? dataset facets? exclude runs altogether?). I don't think we should take this on now, but let's add a TODO to make the input parameters here more easily extendable so that we can add those other options later one.

collado-mike (approved):

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -129,5 +129,14 @@ WHERE ds.uuid IN (<dsUuids>)""")
...
           + ") ro ON ro.run_uuid=r.uuid")
+  List<Run> getCurrentRunsWithFacets(@BindList Collection<UUID> jobUuid);
+
+  @SqlQuery(
+      "SELECT DISTINCT on(r.job_name, r.namespace_name) r.*, jv.version as job_version\n"
+          + "    FROM runs_view r\n"
+          + "    INNER JOIN job_versions jv ON jv.uuid=r.job_version_uuid\n"
+          + "    INNER JOIN jobs_view j ON j.uuid=jv.job_uuid\n"
+          + "    WHERE j.uuid in (<jobUuid>) OR j.symlink_target_uuid IN (<jobUuid>)\n"
+          + "    ORDER BY r.job_name, r.namespace_name, created_at DESC\n")
```
phixMe, line 140: I think we should use the more readable syntax variant as we update our queries.
```java
"""
SELECT DISTINCT on(r.job_name, r.namespace_name) r.*, jv.version as job_version
          FROM runs_view
          INNER JOIN job_versions jv ON jv.uuid=r.job_version_uuid
          INNER JOIN jobs_view j ON j.uuid=jv.job_uuid
          WHERE j.uuid in (<jobUuid>) OR j.symlink_target_uuid IN (<jobUuid>)
          ORDER BY r.job_name, r.namespace_name, created_at DESC
"""
```

phixMe (approved):
Thanks so much!

api/src/main/java/marquez/service/LineageService.java
prachim-collab: Couldn't agree more. I had small aversion to adding a flag to make this work, but there was no other better option. I also thought in future if more changes like this come along that alter api significantly, we could add these as options to api query parameters, or create more broken apis to get specific data.

prachim-collab: added TODO

api/src/main/java/marquez/db/LineageDao.java
prachim-collab: updated the syntax as you asked

Issue #2216, opened 2022-10-26
Error: value too long for type character varying(255)


prachim-collab:
Error in Db migration script.

```
Message    : ERROR: value too long for type character varying(255)
Location   : marquez/db/migration/V48__dataset_symlinks.sql (/usr/src/app/file:/usr/src/app/datakin-api.jar!/marquez/db/migration/V48__dataset_symlinks.sql)
Line       : 16
Statement  : INSERT INTO dataset_symlinks (dataset_uuid, name, namespace_uuid, is_primary, created_at, updated_at)
SELECT d.uuid, d.name, d.namespace_uuid, TRUE, d.created_at, d.updated_at FROM datasets d
```

V48__dataset_symlinks.sql Db migration script is throwing error while it inserts data in  dataset_symlinks table

pawel-big-lebowski:
@prachim-collab You're right 😢  This happens because of:
https://github.com/MarquezProject/marquez/blob/db92c4f4cd53e43b965ad02f1d4055e9ef1cbd4d/api/src/main/resources/marquez/db/migration/V40__alter_tables_resize_namespace_and_connection_url.sql#L3

The solution to this would be:
 * modify `V48__dataset_symlinks.sql` and change column type to varchar for those who failed on migration, 
 * create extra migration step to alter column for those who succeeded migration so that they can have longer names in future.

prachim-collab:
> @prachim-collab You're right 😢 This happens because of:
> 
> https://github.com/MarquezProject/marquez/blob/db92c4f4cd53e43b965ad02f1d4055e9ef1cbd4d/api/src/main/resources/marquez/db/migration/V40__alter_tables_resize_namespace_and_connection_url.sql#L3
> 
> The solution to this would be:
> 
> * modify `V48__dataset_symlinks.sql` and change column type to varchar for those who failed on migration,

The change to varchar should take effect for all old and new rows automatically. 

> * create extra migration step to alter column for those who succeeded migration so that they can have longer names in future.

The insert statement in V48__dataset_symlinks.sql would have inserted partial data successfully so we also need to modify the insert to fail gracefully on duplicate keys and continue inserting missing data.

PR #2217, opened 2022-10-27
fix symlink table column length

pawel-big-lebowski:

### Problem

Dataset's name column in `dataset_symlinks` table is shorter than column in `datasets` table. 

Closes: #2216

### Solution

 * Change the existing V48 migration script to allow proper migration for users who did not upgrade yet,
 * Add extra migration script to extend column length for users who did upgrade but did not experience the issues.
 
### Checklist

### DIFF
--- a/api/src/main/resources/marquez/db/migration/V48__dataset_symlinks.sql
+++ b/api/src/main/resources/marquez/db/migration/V48__dataset_symlinks.sql
@@ -2,7 +2,7 @@
 
 CREATE TABLE dataset_symlinks (
   dataset_uuid      UUID,
-  name              VARCHAR(255) NOT NULL,
+  name              VARCHAR NOT NULL,
   namespace_uuid    UUID REFERENCES namespaces(uuid),
   type              VARCHAR(64),
   is_primary        BOOLEAN DEFAULT FALSE,
--- a/api/src/main/resources/marquez/db/migration/V52__alter_dataset_symlinks.sql
+++ b/api/src/main/resources/marquez/db/migration/V52__alter_dataset_symlinks.sql
@@ -1,0 +1,2 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+ALTER TABLE dataset_symlinks ALTER COLUMN name TYPE VARCHAR;


wslulciuc (approved):
~LGTM 💯 🥇~ Sorry, approved too soon!

api/src/main/resources/marquez/db/migration/V48__dataset_symlinks.sql
```diff
@@ -2,7 +2,7 @@
 
 CREATE TABLE dataset_symlinks (
   dataset_uuid      UUID,
-  name              VARCHAR(255) NOT NULL,
+  name              VARCHAR NOT NULL,
   namespace_uuid    UUID REFERENCES namespaces(uuid),
   type              VARCHAR(64),
   is_primary        BOOLEAN DEFAULT FALSE,
```
prachim-collab, line 8: The insert statement in V48__dataset_symlinks.sql line#16 would have inserted partial data successfully so we also need to modify the insert to fail gracefully on duplicate keys and continue inserting missing data.

collado-mike: I think the insert job would fail atomically and rollback, so users that face this problem shouldn't have any partial data. However, we should add an insert statement in `V52` and just add an `ON CONFLICT DO NOTHING` directive to ensure that everything in the `datasets` table has a value in the `dataset_symlinks`.

pawel-big-lebowski: My understanding of Flyway scripts (based on [docs](https://flywaydb.org/documentation/learnmore/faq#rollback):

> Flyway runs each migration in a separate transaction. In case of failure this transaction is rolled back. 

 * If a migration fails, nothing has been changed in DB, you just need to run migration script again with a fixed code.
 * If a migration succeeded (which means one did not have long names), we add extra migration to extend field length and don't need to insert anything.

collado-mike: Ok, let's leave it as is and trust Flyway to behave as expected. I've seen unexpected behaviors in production deployments, so I don't entirely trust our setup, but I can't pinpoint a specific issue. This should work for most users, and I don't want to unnecessarily burden all users with a potentially costly INSERT migration without a specific issue to point to.

PR #2244, opened 2022-11-14
delete: add endpoint for deleting namespaces

mobuchowski:
Add option to hide/soft delete namespace.

When namespace is hidden, all datasets and jobs contained by this namespace are also hidden. Namespace stops being shown on UI - frontend filters it by isHidden attribute, instead of filtering on backend. Reason for doing it this way is huge potential complexity of changing namespace contract - for example, a lot of methods check whether namespace exists before performing.

Deleted namespace is being undeleted when relevant OpenLineage event is received. This does not automatically undelete all the datasets and jobs in the namespace, only those that are received in this event.

Also, fixes bug where deleted child job (from an event with `ParentRunFacet`) wasn't getting deleted due to internal job name handling.

Closes: https://github.com/MarquezProject/marquez/issues/2095

Signed-off-by: Maciej Obuchowski [obuchowski.maciej@gmail.com](mailto:obuchowski.maciej@gmail.com)


api/src/main/java/marquez/common/Utils.java
```diff
@@ -108,7 +108,8 @@ static final class ZonedDateTimeMixin {}
 
   public static String toJson(@NonNull final Object value) {
     try {
-      return MAPPER.writeValueAsString(value);
+      String x = MAPPER.writeValueAsString(value);
```
pawel-big-lebowski, line 111: Look's like some debugging leftover.

api/src/test/java/marquez/DatasetIntegrationTest.java
```diff
@@ -495,4 +452,109 @@ public void testApp_getDatasetContainsColumnLineage() {
...
+              assertThat(namespace.getIsHidden()).isFalse();
+              assertThat(namespace.getName()).isEqualTo(namespaceName);
+            });
+
+    datasets = client.listDatasets(namespaceName);
+    assertThat(datasets).hasSize(1);
+
+    jobs = client.listJobs(namespaceName);
+    assertThat(jobs).hasSize(1);
+  }
```
pawel-big-lebowski, line 559: Could we add some `undelete` scenario?

api/src/test/java/marquez/db/migrations/V44_2__BackfillAirflowParentRunsTest.java
```diff
@@ -1,108 +0,0 @@
-/*
- * Copyright 2018-2022 contributors to the Marquez project
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package marquez.db.migrations;
```
pawel-big-lebowski, line 6: Please provide some comment in PR for reviewer about why are these files deleted

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -292,15 +292,27 @@ DatasetRow upsert(
...
+  @SqlUpdate(
+      """
+    UPDATE datasets d
+    SET is_hidden = true
+    FROM dataset_symlinks ds
+    JOIN namespaces n
+    ON n.uuid=ds.namespace_uuid
+    WHERE ds.dataset_uuid = d.uuid
+    AND n.name=:namespaceName
+  """)
```
pawel-big-lebowski, line 304: Shouldn't we also consider namespace name of the symlink?

api/src/main/java/marquez/db/JobDao.java
```diff
@@ -88,6 +88,16 @@ SELECT run_uuid, JSON_AGG(e.facets) AS facets
...
+  @SqlUpdate(
+      """
+  UPDATE jobs
+  SET is_hidden = true
+  FROM namespaces n
+  WHERE jobs.namespace_uuid = n.uuid
+  AND n.name = :namespaceName
+  """)
+  void deleteByNamespaceName(String namespaceName);
+
```
pawel-big-lebowski, line 100: Shouldn't we also consider job symlinks?

api/src/main/java/marquez/db/NamespaceDao.java
```diff
@@ -78,10 +78,20 @@ default Namespace upsertNamespaceMeta(
...
+
   default NamespaceRow upsertNamespaceRow(
       UUID uuid, Instant now, String name, String currentOwnerName) {
     doUpsertNamespaceRow(uuid, now, name, currentOwnerName);
-    return findNamespaceByName(name).orElseThrow();
+    NamespaceRow namespaceRow = findNamespaceByName(name).orElseThrow();
+    if (namespaceRow.getIsHidden()) {
+      undelete(namespaceRow.getName());
+    }
+    return namespaceRow;
```
pawel-big-lebowski, line 94: Aren't we returning a namespace row that is still hidden?

api/src/test/java/marquez/db/migrations/V44_2__BackfillAirflowParentRunsTest.java
mobuchowski: Java dto changed (`NamespaceRow`), and they don't match the state of db during those tests. Solution would be to either provide separate java classes or delete those tests.

Since the migration is not from few versions ago and we did not hear about any bugs during the migration, the decision was to delete the tests.

api/src/main/java/marquez/db/JobDao.java
mobuchowski: Discussed offline. No - because the query captures other symlinks directly, and those are symlinks, not hard links - we don't want to hide them by design.

api/src/main/java/marquez/db/DatasetDao.java
mobuchowski: Discussed offline. No - because the query captures other symlinks directly, and those are symlinks, not hard links - we don't want to hide them by design.

api/src/main/java/marquez/common/Utils.java
mobuchowski: Fixed.

api/src/main/java/marquez/db/NamespaceDao.java
pawel-big-lebowski: I may be wrong but my understanding is:
 * we get `namespaceRow` which has `is_hidden=true`,
 * we perform `udelete` in database which sets `is_hidden=false`,
 * de method returns unmodified `namespaceRow` which still has wrong `is_hidden=true`

api/src/test/java/marquez/db/migrations/V44_2__BackfillAirflowParentRunsTest.java
pawel-big-lebowski: thanks

api/src/main/java/marquez/db/NamespaceDao.java
wslulciuc: Right, but the reference `namespaceRow` will return `namespaceRow.getIsHidden() = true` unless you query for the namespace again. Why not always set `is_hidden=true` when calling `doUpsertNamespaceRow()`? There's also `upsertNamespaceRow()` which seems to implement this reset logic? Wondering why we have both.

codecov.yml
```diff
@@ -1,4 +1,11 @@
 codecov:
   branch: main
+
+coverage:
   status:
     patch: off
+
+ignore:
+  - "api/src/main/java/marquez/db/migrations/V44_1__UpdateRunsWithJobUUID.java"
```
wslulciuc, line 9: OOh, cool! Had no idea this was even an option 💯

clients/java/src/main/java/marquez/client/MarquezClient.java
```diff
@@ -230,6 +230,10 @@ public Dataset deleteDataset(@NonNull String namespaceName, @NonNull String data
     return Dataset.fromJson(bodyAsJson);
   }
 
+  public void deleteNamespace(@NonNull String namespaceName) {
```
wslulciuc, line 233: _Minor_: I'd be explicit and use `Client.softDeleteNamespace()`?

api/src/test/java/marquez/DatasetIntegrationTest.java
wslulciuc: The test does check the number of datasets after the namespace is undeleted. But, looks like that's a new dataset added to the namespace and not the original datasets?

api/src/main/java/marquez/db/NamespaceDao.java
mobuchowski: The basic idea was to follow @collado-mike's comment

```

Comment (mobuchowski) [reply api/src/main/java/marquez/db/NamespaceDao.java:94]: Fixed the original issue by assigning to `namespaceRow` on `undelete`. If we want to change this logic substantially, I don't think this PR is a good place to do that.

File: api/src/main/java/marquez/api/NamespaceResource.java [hunk 9]
```diff
@@ -77,6 +78,23 @@ public Response list(
     return Response.ok(new Namespaces(namespaces)).build();
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @DELETE
+  @Path("/namespaces/{namespace}")
+  @Produces(APPLICATION_JSON)
+  public Response delete(@PathParam("namespace") NamespaceName name) {
```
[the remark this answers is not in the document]
mobuchowski: Should we do that for every delete - `Dataset`, `Job` one?? 

My previous idea was that we'll later want to add query param that could specify hard deletes.

api/src/main/java/marquez/api/NamespaceResource.java
```diff
@@ -77,6 +78,23 @@ public Response list(
...
+    final Namespace namespace =
+        namespaceService
+            .findBy(name.getValue())
+            .orElseThrow(() -> new NamespaceNotFoundException(name));
+    datasetService.deleteByNamespaceName(namespace.getName().getValue());
+    jobService.deleteByNamespaceName(namespace.getName().getValue());
```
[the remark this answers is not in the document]
mobuchowski: Only thing would be that visually empty namespace would appear on the UI, as this is filtered on the frontend.

api/src/test/java/marquez/DatasetIntegrationTest.java
mobuchowski: Yes, the idea here is not to automatically undelete everything in the namespace.

wslulciuc (approved):
A much much requested feature! 💯 🥇

api/src/main/java/marquez/api/NamespaceResource.java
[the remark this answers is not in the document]
wslulciuc: Great point! Yeah, no need to rename as we will _eventually_ support _hard_ deletes.

api/src/main/java/marquez/api/NamespaceResource.java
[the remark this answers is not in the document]
wslulciuc: I _think_ it's fine as is. Perhaps I would have had `NamespaceService.delete()` soft delete the dataset and job within a transaction.

PR #2253, opened 2022-11-18
Ensure job data in lineage query is not null or empty

wslulciuc:
### Problem

An unknown edge case in our **lineageAPI** exists when querying for a job `uuid` that has no lineage when calling [`LineageDao.getLineage()`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/LineageDao.java#L73), yet is associated with a given dataset `D`. That is, given a dataset `D`, a reverse job lookup is performed using [`LineageDao.getJobFromInputOrOutput()`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/LineageDao.java#L92) (which returns the job `uuid` that produced dataset `D`) but when querying for the lineage data for the returned job an empty set is returned, therefore resulting in a backend exception (=`5xx` status code) when invoking `LineageDao.getJobFromInputOrOutput()`.

### Solution

This PR ensures the lineage data for a job is not empty (an empty lineage graph is returned instead) and introduces logs to better understand the scenario in which this unknown edge case occurs for debugging purposes. **This PR also adds a API check to ensure the query param `nodeID` provided in the lineageAPI call exists.** 

### Checklist


api/src/main/java/marquez/service/LineageService.java
```diff
@@ -48,14 +49,27 @@ public LineageService(LineageDao delegate, JobDao jobDao) {
...
+    // Ensure job data is not empty, an empty set cannot be passed to LineageDao.getCurrentRuns() or
+    // LineageDao.getCurrentRunsWithFacets().
+    if (jobData.isEmpty()) {
+      // Log warning, then return an empty lineage graph; a graph must contain at most one
+      // job->dataset relationship.
+      log.warn(
+          "Failed to get lineage for job '{}' associated with node '{}', returning empty graph...",
+          job,
+          nodeId.getValue());
+      return toLineage(Sets.newHashSet(), Sets.newHashSet());
```
collado-mike, line 70: Isn't the graph in the UI re-rendered based on the contents of this response? If a dataset is selected, rather than a graph with a single node, we'll show an error, right? We should make sure that the node is a valid node (e.g., a real dataset), then respond with at least that node in the graph.

collado-mike (changes requested):

api/src/main/java/marquez/service/LineageService.java
```diff
@@ -85,19 +99,23 @@ public Lineage lineage(NodeId nodeId, int depth, boolean withRunFacets) {
...
       log.warn(
-          "Found jobs {} which no longer share lineage with dataset {} - discarding",
-          jobData.stream().map(JobData::getId).toList());
-      DatasetId datasetId = nodeId.asDatasetId();
-      DatasetData datasetData =
-          getDatasetData(datasetId.getNamespace().getValue(), datasetId.getName().getValue());
-      return new Lineage(
-          ImmutableSortedSet.of(
-              Node.dataset().data(datasetData).id(NodeId.of(datasetData.getId())).build()));
+          "Found jobs '{}' which no longer share lineage with dataset '{}' - discarding",
```
collado-mike, line 102: The jobs will be a list, surrounded by array brackets - I don't see a point in quoting them.

api/src/main/java/marquez/service/LineageService.java
```diff
@@ -49,14 +50,27 @@ public LineageService(LineageDao delegate, JobDao jobDao) {
 
   // TODO make input parameters easily extendable if adding more options like 'withJobFacets'
   public Lineage lineage(NodeId nodeId, int depth, boolean withRunFacets) {
+    log.debug("Attempting to get lineage for node '{}' with depth '{}'", nodeId.getValue(), depth);
     Optional<UUID> optionalUUID = getJobUuid(nodeId);
     if (optionalUUID.isEmpty()) {
-      throw new NodeIdNotFoundException("Could not find node");
+      throw new NodeIdNotFoundException(String.format("Node '%s' not found!", nodeId.getValue()));
     }
```
collado-mike, line 57: Ok, minor change, but can we just update this block to work like the block you just added - if there is no job data, just return lineage for the dataset. This could be considered an API change, but it's not specified in the contract and, personally, I think it's just broken.

collado-mike (changes requested):

api/src/main/java/marquez/service/LineageService.java
wslulciuc: Yeah, I think that's a reasonable default at least until we revisit the behavior on missing relationships.

>  I think it's just broken

Are you referring to the `LineageAPI` or a broken relationship between nodes?

collado-mike: The fact that we throw an exception if we don't find a job associated with the dataset. In the code you just added, and in the PR I merged, if no job is _currently_ associated with the dataset, we return a lineage graph with a single node. But in this block, if no job was _ever_ associated with the dataset, we throw an exception. Why not just return the dataset node as we do below?

wslulciuc: Yeah, I was going to return a single node, but was just curious (and wanted clarification) on what you meant by "just broken" :p

collado-mike: Got it - I misread your original statement to suggest that the _current_ behavior was correct 🤦🏽‍♂️
By "just broken", I mean to say it's a poor user experience. It works for jobs, it should work for datasets, too, IMO.

wslulciuc: Agreed! and no worries ;)

collado-mike (approved):
Nice! 🥳

collado-mike (approved):

PR #2254, opened 2022-11-18
add i18next and config for i18n of UI

merobi-hub:
Signed-off-by: Michael Robinson <merobi@gmail.com>

### Problem

Internationalization is strongly recommended by the LFAI (in the CII Silver Badge). The UI is not currently internationalized.

Closes: #ISSUE-NUMBER

### Solution

This change adds i18next, configures it, and edits the text to make use of the tool's translation capabilities.

### Checklist


package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "@types/react": "^18.0.25",
+    "i18next": "^22.0.6",
+    "i18next-browser-languagedetector": "^7.0.1",
+    "react-i18next": "^12.0.0"
+  }
+}
```
phixMe, line 8: This should be inside the existing `package.json` file inside the 'web` directory.

Just add these to the file:
```
    "i18next": "^22.0.6",
    "i18next-browser-languagedetector": "^7.0.1",
    "react-i18next": "^12.0.0"
```
and run `npm i`

web/src/i18n.js
```diff
@@ -0,0 +1,46 @@
...
+          header: 'API Docs',
+          JobTab1: 'ÚLTIMA EJECUCIÓN',
+          JobTab2: 'HISTORIAL DE EJECUCIONES'
+        }
+      }
+
+    }
+  });
+
+export default i18n;
```
phixMe, line 46: You can apply formatting by running `npm run eslint-fix` that will cleanup newline issues like this.

web/src/i18n.js
```diff
@@ -0,0 +1,46 @@
+import i18n from 'i18next';
```
phixMe, line 1: Can you make this a TypeScript file? End it with the `*.ts` instead of the JavaScript extension.

package.json
phixMe: This file along with the lock file should then be deleted.

web/src/i18n.js
merobi-hub: ran eslint with npx, think I resolved this

merobi-hub:
@phixMe reverting to `i18n.js` temporarily because the .ts version was causing the build to fail. Here's what it looks like when I set my system language to French:
Screen Shot 2022-11-18 at 22 18 04

merobi-hub:
When the translation is longer than the original, the effect is not great, unfortunately. But there's no reason to assume these "translations" (from Google) are worth keeping. @julienledem @mobuchowski would you please review?
Screen Shot 2022-11-19 at 16 17 30

web/src/i18n.js
```diff
@@ -0,0 +1,107 @@
...
+            location: 'LOKALIZACJA',
+            empty_title: 'Brak dostępnych informacji o przebiegu',
+            empty_body: 'Spróbuj dodać kilka przebiegów dla tego zadania.'
+          },
+          search: {
+            search: 'Wyszukaj Oferty',
+            jobs: 'Pracy',
+            and: 'i',
+            datasets: 'Zestawy Danych'
+          }
```
mobuchowski, line 100: I think we need to agree on run, job, dataset translations. IMO

```
RUN: wykonanie
JOB: zadanie
DATASET: zbiór danych
```

I think only `wykonanie` might be controversial, but I don't have better idea. `Bieg` or `przebieg` make sense only as the sport IMO.

then: 
```
docs_link: Dokumentacja API
latest_tab: OSTATNIE WYKONANIE
history tab: HISTORIA WYKONAŃ
location: LOKALIZACJA
empty_title: Brak dostępnych informacji o wykonaniu
empty_body: Spróbuj dodać kilka wykonań dla tego zadania

search: Wyszukiwanie
jobs: Zadania
and: i
datasets: Zbiory Danych
```

Still, translations of "computer concepts" will always sound awkward in Polish for me.

mobuchowski:
Does it add an option to switch language or does it force a translation based on something like browser language?

I think it would be great to have language switcher if we do something like it.

merobi-hub:
@mobuchowski i18next does support a language-switcher setup, but that route is a bit more labor-intensive.

merobi-hub:
@phixMe To translate the text in web/src/components/lineage/Lineage.tsx I did a workaround that required changing the tsconfig.json to set 'noImplicitAny' to 'false'. The initial problem was that there doesn't seem to be any way to use the `useTranslation` function (`const {t} = useTranslation()`), as in the other pages. I don't intend to leave this change to tsconfig. Would you please take a look at Lineage.tsx when you have time?

web/package.json
```diff
@@ -59,7 +59,10 @@
     "redux": "^4.0.1",
     "redux-logger": "^3.0.6",
     "redux-saga": "^1.0.2",
-    "webpack-merge": "^4.2.1"
+    "webpack-merge": "^4.2.1",
+    "i18next": "^22.0.6",
+    "i18next-browser-languagedetector": "^7.0.1",
+    "react-i18next": "^12.0.0"
```
phixMe, line 65: There should be an update to the lock file in this directory.

web/src/components/lineage/Lineage.tsx
```diff
@@ -23,6 +23,8 @@ import MqEmpty from '../core/empty/MqEmpty'
 import MqText from '../core/text/MqText'
 import Node from './components/node/Node'
 import ParentSize from '@visx/responsive/lib/components/ParentSize'
+import { useTranslation } from 'react-i18next'
+import i18next from '../../i18n'
```
phixMe, line 27: @merobi-hub I think you don't want to be importing this here. This should be your type issue for `no-implicit-any`

web/src/i18n.js
merobi-hub: thank you, @mobuchowski . @pawel-big-lebowski @JDarDagran @tnazarew WDYT?

JDarDagran: I'd say `wykonanie` sounds more natural to me. We shouldn't be so strict about translating this one to one.  Job and dataset are just right.

mobuchowski:
@merobi-hub I understand it would require more work, but I personally would be unhappy to be forced to use some language, especially other than english. 

The internalization makes sense for UI purposes, but then if you want to write some code, you still have to work with english API and other code parts.

merobi-hub:
@mobuchowski That's a good point. For some reason, I've been in a rush to complete this, but there's actually no particular urgency. I'll look at the docs about the button approach again.

merobi-hub:
I've redone this so it is properly compatible with Typescript (hopefully). The (updated) translations have been moved and are now in `web/src/i18n/config.ts`.

web/src/i18n.js
merobi-hub: Thanks, @mobuchowski and @JDarDagran . I've revised the translations.

merobi-hub:
Although it still needs some cosmetic work, the language switcher is functional now
Screen Shot 2022-11-28 at 21 04 27

web/src/components/sidenav/Sidenav.tsx
```diff
@@ -95,6 +113,28 @@ class Sidenav extends React.Component<SidenavProps> {
...
+                {'es'}
+              </MenuItem>
+              <MenuItem key={'fr'} value={'fr'}>
+                {'fr'}
+              </MenuItem>
+              <MenuItem key={'pl'} value={'pl'}>
+                {'pl'}
+              </MenuItem>
+            </Select>
+          </FormControl>
```
phixMe, line 137: How is this looking now? Can you add a screenshot to the PR?

web/tsconfig.json
```diff
@@ -11,7 +11,8 @@
     "lib": ["es2016", "es2017", "dom"],
     "skipLibCheck": true,
     "esModuleInterop": true,
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "resolveJsonModule": true
   },
   "include": ["./src/**/*"],
   "exclude": ["./src/__tests__/*"]
```
phixMe, line 18: Can we do without these config changes now? Implicit any in TypeScript is not the best rule to have disabled.

web/webpack.dev.js
```diff
@@ -22,7 +22,7 @@ const webpackDev = {
     },
     proxy: {
       '/api': {
-        target: `http://${process.env.MARQUEZ_HOST || 'localhost'}:${process.env.MARQUEZ_PORT || 8080}/`,
+        target: `http://${process.env.MARQUEZ_HOST || 'localhost'}:${process.env.MARQUEZ_PORT || 5000}/`,
```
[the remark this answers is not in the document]
merobi-hub: where's the facepalm emoji? (reverted this)

merobi-hub:
Screen Shot 2022-12-06 at 13 42 03
Screen Shot 2022-12-06 at 13 42 39
Screen Shot 2022-12-06 at 13 43 07
Screen Shot 2022-12-06 at 13 43 36
Screen Shot 2022-12-06 at 13 44 09

merobi-hub:
@phixMe This is RFR. Added a reload after a language change and fixed reloading issue so the new language setting persists.

merobi-hub:
OK, now this is actually RFR @phixMe

phixMe (approved):
Looks great @merobi-hub. Thanks for the contribution.

web/src/components/sidenav/Sidenav.tsx
```diff
@@ -74,13 +92,10 @@ class Sidenav extends React.Component<SidenavProps> {
...
                 id={'eventsButton'}
-                title={'EVENTS'}
+                title={i18next.t('sidenav.events')}
                 active={this.props.location.pathname === '/events'}
               >
-                <SVG
-                  src="https://raw.githubusercontent.com/MarquezProject/marquez/main/web/src/img/iconSearchArrow.svg"
-                  width={'30px'}
-                />
+                <SVG src='../../img/iconSearchArrow.svg' width={'30px'} />
```
wslulciuc, line 98: @merobi-hub, this reverts the change for https://github.com/MarquezProject/marquez/pull/2280

Issue #2262, opened 2022-11-21
Point in time column lineage endpoint - implementation


pawel-big-lebowski:
Implement point in time column-level lineage endpoint according to the proposal in #2117 2193

wslulciuc:
Related issues https://github.com/MarquezProject/marquez/issues/1519

pawel-big-lebowski:
todo list

 * implement point-in-time endpoint -> PR #2265
 * support point-in-time together with `withDownstream` feature,
 * implement Java client -> #2269
 * implement Python client.

PR #2265, opened 2022-11-25
point-in-time endpoint for column-level lineage

pawel-big-lebowski:

### Problem

Point in time column lineage endpoint - implementation.

Closes: #2262

### Solution

 * Dataset or job version from a request is used to determine date time when a version was created. 
 * `column_lineage` table is filtered for rows earlier than a determined date time. 
 * Column-lineage rows, dataset rows and dataset_version rows make use of the same date time. 
 * The combination of `point-in-time` and `withDownstream` features does not suit well and deserves a separate PR if needed. 

 
### Checklist


api/src/main/java/marquez/common/models/JobVersionId.java
```diff
@@ -19,4 +19,9 @@ public class JobVersionId {
   @NonNull NamespaceName namespace;
   @NonNull JobName name;
   @NonNull UUID version;
+
+  public static JobVersionId of(
```
mobuchowski, line 23: why not `new` if we already have that constructor?

api/src/main/java/marquez/service/ColumnLineageService.java
```diff
@@ -180,4 +242,6 @@ public void enrichWithColumnLineage(List<Dataset> datasets) {
         .filter(dataset -> datasetLineage.containsKey(dataset))
         .forEach(dataset -> dataset.setColumnLineage(datasetLineage.get(dataset)));
   }
+
+  private record ColumnNodes(Instant createdAtUntil, List<UUID> nodeIds) {}
```
mobuchowski, line 246: `record` 👍

mobuchowski (approved):

api/src/main/java/marquez/common/models/JobVersionId.java
pawel-big-lebowski: We do have such a convention for model classes and I kept constant with it.

wslulciuc: Yeah, it's a factory method we have for our models as convention throughout the code base. Thanks @pawel-big-lebowski for keeping things constant ❤️

Issue #2298, opened 2022-12-08
V52 database migration breaks existing Marquez installations


collado-mike:
The migration introduced in https://github.com/MarquezProject/marquez/pull/2217 is not backward compatible with existing marquez installations. The following exception is thrown during the flyway migration.
Database migrations can't change the types of columns in existing views in order to change the type, it is necessary to drop the existing view, then alter the columns, and recreate the view.

Marquez is missing a way of testing these migrations against existing installations. We should figure out a way to validate these scripts before merging to main.

```
{"exception":"org.postgresql.util.PSQLException: ERROR: cannot alter type of a column used by a view or rule
  Detail: rule _RETURN on view datasets_view depends on column \"name\"
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2676)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2366)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:356)
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:496)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:413)
	at org.postgresql.jdbc.PgStatement.executeWithFlags(PgStatement.java:333)
	at org.postgresql.jdbc.PgStatement.executeCachedSql(PgStatement.java:319)
	at org.postgresql.jdbc.PgStatement.executeWithFlags(PgStatement.java:295)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:290)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.apache.tomcat.jdbc.pool.StatementFacade$StatementProxy.invoke(StatementFacade.java:118)
	at jdk.proxy2/jdk.proxy2.$Proxy47.execute(Unknown Source)
	at org.flywaydb.core.internal.jdbc.JdbcTemplate.executeStatement(JdbcTemplate.java:201)
[...]

PR #2308, opened 2022-12-09
Update migration query to make it work with existing view

fm100:
Signed-off-by: Minkyu Park <minkyu.park.200@gmail.com>

### Problem
V52 migration fails if the `datasets_view` is already created by the previous migration, because postgresql does not allow altering the type of the column when it is being used by views.

Closes: #2298 

### Solution
This PR changes the V52 migration query to drop the view before `ALTER`. Because repeatable migration runs only when its checksum changes, we have to get the view definition first, and then drop and recreate it as before.

### Checklist

### DIFF
--- a/api/src/main/resources/marquez/db/migration/V52__alter_dataset_symlinks.sql
+++ b/api/src/main/resources/marquez/db/migration/V52__alter_dataset_symlinks.sql
@@ -1,2 +1,28 @@
 /* SPDX-License-Identifier: Apache-2.0 */
-ALTER TABLE dataset_symlinks ALTER COLUMN name TYPE VARCHAR;
+
+DO
+$$
+DECLARE
+    datasets_view_exists boolean;
+    datasets_view_definition text;
+BEGIN
+    SELECT EXISTS (
+        SELECT * FROM information_schema.views
+        WHERE table_schema='public' AND table_name='datasets_view'
+    ) INTO datasets_view_exists;
+
+    IF datasets_view_exists THEN
+        -- Altering is not allowed when the column is being used from views. So here,
+        -- we temporarily drop the view before altering and recreate it.
+        SELECT view_definition FROM information_schema.views
+        WHERE table_schema='public' AND table_name='datasets_view'
+        INTO datasets_view_definition;
+
+        DROP VIEW datasets_view;
+        ALTER TABLE dataset_symlinks ALTER COLUMN name TYPE VARCHAR;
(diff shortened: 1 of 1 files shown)


api/src/main/resources/marquez/db/migration/V52__alter_dataset_symlinks.sql
```diff
@@ -1,2 +1,28 @@
...
+        SELECT view_definition FROM information_schema.views
+        WHERE table_schema='public' AND table_name='datasets_view'
+        INTO datasets_view_definition;
+
+        DROP VIEW datasets_view;
+        ALTER TABLE dataset_symlinks ALTER COLUMN name TYPE VARCHAR;
+        EXECUTE format('CREATE VIEW datasets_view AS %s', datasets_view_definition);
+    ELSE
+        ALTER TABLE dataset_symlinks ALTER COLUMN name TYPE VARCHAR;
+    END IF;
```
collado-mike, line 26: 😳

collado-mike (approved):

api/src/main/resources/marquez/db/migration/V52__alter_dataset_symlinks.sql
pawel-big-lebowski: I think it's an overkill, as 
```
DROP VIEW IF EXISTS datasets_view;
```
at the beginning of the script should be enough as in PR: 
https://github.com/MarquezProject/marquez/pull/2313/files

pawel-big-lebowski (changes requested):
I think the view will recreate on its own.

pawel-big-lebowski (commented):
@fm100 You're right that dropping the view wouldn't be sufficient to go. Thanks for explanation. 

Why is this better than modifying `R__3_Datasets_view.sql` file to get another checksum or making it run each time as described here -> https://flywaydb.org/blog/flyway-timestampsandrepeatables ?

pawel-big-lebowski (approved):
+1 not to block the progress on this.

fm100:
> @fm100 You're right that dropping the view wouldn't be sufficient to go. Thanks for explanation.
> 
> Why is this better than modifying `R__3_Datasets_view.sql` file to get another checksum or making it run each time as described here -> https://flywaydb.org/blog/flyway-timestampsandrepeatables ?

Interesting. I don't think that this is better than making repeatable migration run. I did try adding a new line but it didn't change the checksum and this is what I ended up doing. It seems like that white spaces don't change the checksum but comments do. Anyway, unblocking this is pretty important and let's merge this.

PR #2313, opened 2022-12-12
drop view before modifying table fields

pawel-big-lebowski:

### Problem

👋 Thanks for opening a [pull request](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#submitting-a-pull-request)! Please include a brief summary of the problem your change is trying to solve, or bug fix. If your change fixes a bug or you'd like to provide context on why you're making the change, please [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) as follows:

Closes: #ISSUE-NUMBER

### Solution

### Checklist


api/src/main/resources/marquez/db/migration/V52__alter_dataset_symlinks.sql
```diff
@@ -1,2 +1,3 @@
 /* SPDX-License-Identifier: Apache-2.0 */
+DROP VIEW IF EXISTS datasets_view;
```
pawel-big-lebowski, line 2: https://flywaydb.org/documentation/concepts/migrations.html#repeatable-migrations

> Repeatable migrations have a description and a checksum, but no version. Instead of being run just once, they are (re-)applied every time their checksum changes.

> Within a single migration run, repeatable migrations are always applied last, after all pending versioned migrations have been executed. Repeatable migrations are applied in the order of their description.

fm100: As the doc says, `they are (re-)applied every time their checksum changes.` Just dropping the view will not create a view again and the application will break.

fm100 (changes requested):
https://github.com/MarquezProject/marquez/pull/2308

We should drop and recreate a view after altering like above PR.

PR #2326, opened 2022-12-15
Add CI workflow for testing db migrations

wslulciuc:
### Problem

We use [flyway](https://flywaydb.org/) to manage schema changes to Marquez's model, but no validation is performed in CI to ensure schema changes are (backwards) compatible when applied to an existing database (with metadata).

### Solution

Add CI job `migrate-db` to apply:
1. migrations on the latest Marquez release (seeds the database with metadata and creates a backup)
2. migrations on latest Marquez build using backup

Screen Shot 2023-01-04 at 12 50 02 AM

You can see the output logs of the CI job `migrate-db` [here](https://app.circleci.com/pipelines/github/MarquezProject/marquez/8841/workflows/fc03e0ca-dd2e-4500-917c-3b403d11fc68/jobs/64466/parallel-runs/0/steps/0-103)!

**Note:** To reuse `docker/up.sh` in testing db migrations changes, flags `--no-web` and `--no-volumes` have been added along with `--args` (as well as reorganizations of the docker compose files):

```
$ ./docker/up.sh --help
usage: ./up.sh [FLAGS] [ARG...]
A script used to run Marquez via Docker

EXAMPLES:
  # Build image from source
  $ ./up.sh --build

  # Build image from source, then seed HTTP API server with metadata
  $ ./up.sh --build --seed

  # Use tagged image
  ./up.sh --tag X.Y.X

  # Use tagged image, then seed HTTP API server with metadata
  ./up.sh --tag X.Y.X --seed

  # Set HTTP API server port
  ./up.sh --api-port 9000

ARGUMENTS:
  -a, --api-port int          api port (default: 5000)
  -m, --api-admin-port int    api admin port (default: 5001)
  -w, --web-port int          web port (default: 3000)
  -t, --tag string            docker image tag (default: 0.29.0)
  --args string               docker arguments

FLAGS:
  -b, --build           build images from source
  -s, --seed            seed HTTP API server with metadata
  -d, --detach          run in the background
  --no-web              don't start the web UI
  --no-volumes          don't create volumes
  -h, --help            show help for script
```

### Checklist


pawel-big-lebowski (approved):
I created a dummy PR (https://github.com/MarquezProject/marquez/pull/2344) based on the errors we had in past. It modified an existing view in a way that: should succeed on empty clean install but should fail migration on existing latest version.

Circle CI has failed:
https://app.circleci.com/pipelines/github/MarquezProject/marquez/8842/workflows/6ef5aa9a-a0fb-4ce4-962a-33eafb4fdd16/jobs/64474

which is five-star ⭐ ⭐ ⭐ ⭐ ⭐ 

Would it be possible to contain logs/output from the command exited with -1. That's the only missing part -> no info on how did it fail .

wslulciuc:
Thanks for giving the db migration test a try, @pawel-big-lebowski! Good point on preserving the logs, that should be easy enough.

.circleci/db-migration.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+#
+# Copyright 2018-2022 contributors to the Marquez project
+# SPDX-License-Identifier: Apache-2.0
+#
+# Usage: $ ./db-migration.sh
```
collado-mike, line 6: A descriptive comment on this file will tell readers the intention of this script - it's a bit hard to parse by just reading the bash

.circleci/db-migration.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# (2) Apply db migrations on latest Marquez build using backup
+log "start db using backup:"
+if ! ./docker/up.sh \
+  --args "--exit-code-from seed_marquez" \
+  --no-web \
+  --no-volumes \
+  --build \
+  --seed; then
+  exit 1
```
collado-mike, line 62: `echo` to tell us where in the script we failed

.circleci/db-migration.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+#
+# Copyright 2018-2022 contributors to the Marquez project
+# SPDX-License-Identifier: Apache-2.0
+#
+# Usage: $ ./db-migration.sh
+
+readonly POSTGRES_VERSION="12.1"
+readonly MARQUEZ_VERSION="latest"
```
collado-mike, line 9: Anybody who builds marquez regularly is going to have this image tag all the time. It'll still run in CI, which is fine, but I wonder whether there's any way to fetch the latest released version so that people can catch mistakes locally rather than waiting for the CI build to fail.

collado-mike (commented):
Has this been tested with a non-backward-compatible migration? I'm curious to know if this will catch some of the prior issues we've seen.

.circleci/db-migration.sh
wslulciuc: Very doable. We running [`new-version.sh`](https://github.com/MarquezProject/marquez/blob/main/new-version.sh), we can also bump `MARQUEZ_VERSION `. Which is better than just `latest` and assuming the version. Also helps with debugging.

wslulciuc:
> Has this been tested with a non-backward-compatible migration? I'm curious to know if this will catch some of the prior issues we've seen.

@collado-mike you can see the verification done by @pawel-big-lebowski in https://github.com/MarquezProject/marquez/pull/2344

.circleci/db-migration.sh
```diff
@@ -0,0 +1,68 @@
...
+query_db_migration() {
+  # Start db using backup
+  [[ $(docker ps -f "name=${DB_MIGRATION_BACKUP}" --format '{{.Names}}') == "${DB_MIGRATION_BACKUP}" ]] || \
+    docker run -d --name "${DB_MIGRATION_BACKUP}" \
+      -v "${DB_MIGRATION_VOLUME}:/var/lib/postgresql/data" \
+      "postgres:${POSTGRES_VERSION}"
+  # Query applied db migrations
+  log "latest migration applied to db:"
+  docker exec "${DB_MIGRATION_BACKUP}" \
+    psql -U marquez -c "${DB_MIGRATION_QUERY}"
```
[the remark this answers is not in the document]
wslulciuc: We aren't parsing the response, just displaying it to the console. But, we could do more with the resulting query as follow up.

collado-mike:
> @collado-mike you can see the verification done by @pawel-big-lebowski in #2344

Ah, nice! I looked at the modification to output the container logs - Is it possible to only output the container logs if the migration fails? No biggie if it's a major ordeal (I didn't see where the stdout configuration was done), but it's hard to find the output between the first container run and the second (i.e., to verify what the flyway version was and what it is now) because of all the logging.

wslulciuc:
> Ah, nice! I looked at the modification to output the container logs - Is it possible to only output the container logs if the migration fails?

Yep! Docker compose preserves container logs for us, so we can just print them out anytime. Here's example based on recent changes:

Screen Shot 2023-01-06 at 2 44 25 PM

collado-mike (approved):

Issue #2167, opened 2022-10-05
UI: Soft Delete


howardyoo:
Display option to delete jobs and datasets using soft delete button. Also, feature to search and ‘undelete’ soft deletes if necessary. Might be required to delete ‘group’ of datasets or jobs based on their namespaces - so multi-deletion should be available. Ability to undelete may come later, as deletion should be easier to implement it first.

# Solution
![image](https://user-images.githubusercontent.com/32691630/197675054-8bb131de-38aa-43a4-bc5c-6ed07249a011.png)
Per each dataset and job selected, if the job/dataset's status is 'not deleted,' have a button to 'delete' them (See the `RED` button in the UI) , which will then invoke api calls:
- https://marquezproject.github.io/marquez/openapi.html#operation/deleteDataset
- https://marquezproject.github.io/marquez/openapi.html#operation/deleteJob
To `soft-delete` the resources.
Before deleting, make sure to ask users to confirm, by displaying and asking `Are you sure?` message.
After the deletion is finished, you should 'refresh' the current Graph so that the job/dataset deleted will no longer be visible.

PR #2343, opened 2023-01-03
Soft delete of jobs and datasets. Style dialog component

tito12:
Signed-off-by: tito12 <vladyslav.sedenko@gmail.com>

### Problem

According to the issue https://github.com/MarquezProject/marquez/issues/2167 we need option to soft delete records, it's already possible from API but not from UI. 

### Solution

Added option for jobs and datasets to be able to soft delete record with dialog component and double confirmation from UI side. 

image

### Checklist


wslulciuc:
A much needed addition to the UI! Looks great, we should also added deletion of namespaces as a follow. I'll let @phixMe provide a more thorough review.

web/src/components/Dialog.tsx
```diff
@@ -26,16 +27,28 @@ export default function AlertDialog(props: IProps) {
     <div>
```
phixMe, line 27: Do we need to have this encapsulating tag?

web/src/components/datasets/DatasetDetailPage.tsx
```diff
@@ -147,9 +170,31 @@ const DatasetDetailPage: FunctionComponent<IProps> = props => {
...
+                onClick={() => {
+                  props.dialogToggle('')
+                }}
+              >
+                {i18next.t('jobs.delete')}
+              </Button>
+              <Dialog
+                dialogIsOpen={display.dialogIsOpen}
+                dialogToggle={dialogToggle}
+                title={'Are you sure?'} // i18next.t('jobs.dialogTitleConfirm')
```
phixMe, line 187: Don't we wish have this in the internationalization configuration?

web/src/components/jobs/JobDetailPage.tsx
```diff
@@ -30,23 +33,35 @@ const styles = ({ spacing }: ITheme) => {
   return createStyles({
     root: {
       padding: spacing(2)
+    },
+    buttonDelete: {
+      backgroundColor: theme.palette.error.main,
+      color: 'white',
```
phixMe, line 39: I think I like doing the `theme.palette.common.white` for these in case we ever wish to support light mode.

web/src/components/datasets/DatasetDetailPage.tsx
```diff
@@ -147,9 +170,31 @@ const DatasetDetailPage: FunctionComponent<IProps> = props => {
...
+              <Button
+                color='primary'
+                className={classes.buttonDelete}
+                onClick={() => {
+                  props.dialogToggle('')
+                }}
+              >
+                {i18next.t('jobs.delete')}
+              </Button>
+              <Dialog
```
phixMe, line 184: There are two instances of this dialog... Can the dialog itself just be a reusable component.

web/src/components/datasets/DatasetDetailPage.tsx
```diff
@@ -147,9 +170,31 @@ const DatasetDetailPage: FunctionComponent<IProps> = props => {
               />
             </Tabs>
           </Box>
-          <IconButton onClick={() => history.push('/datasets')}>
-            <CloseIcon />
-          </IconButton>
+          <Box display={'flex'} alignItems={'center'}>
+            <Box mr={1}>
+              <Button
+                color='primary'
```
phixMe, line 176: Can we make this a `ghost` variant... I think it's a little too strong on the page.

web/src/components/Dialog.tsx
```diff
@@ -26,16 +27,28 @@ export default function AlertDialog(props: IProps) {
...
             Cancel
           </Button>
+          <Button
+            className='dialogButton'
+            color='primary'
+            variant='outlined'
+            onClick={props.ignoreWarning}
+          >
+            Continue
+          </Button>
```
phixMe, line 51: Can we make these small buttons inside the dialog?

web/src/components/Dialog.tsx
tito12: This component exist previously with encapsulating tag. Moved it to functional component as rest part components.

web/src/components/datasets/DatasetDetailPage.tsx
tito12: Sure, forgot to add it in previous commit

web/src/components/jobs/JobDetailPage.tsx
tito12: Sure, added it

web/src/components/Dialog.tsx
tito12: Sorry, can you explain what do you mean by that?

web/src/components/datasets/DatasetDetailPage.tsx
tito12: This component is reusable and located in `"components/Dialog.tsx"`, and use it with different props

web/src/components/datasets/DatasetDetailPage.tsx
tito12: Sure, good point

web/src/components/Dialog.tsx
phixMe: I think we could possibly use smaller buttons here since the dialog is so minute, but maybe it's not necessary.

web/src/components/Dialog.tsx
phixMe: Thanks

phixMe (approved):
Thanks, looks good!

PR #2355, opened 2023-01-11
OL facets - PR2 - read facets from views based on lineage_events table

pawel-big-lebowski:

### Problem

Migration of facets from `lineage_events` to separate facets tables will be asynchronous for some users as it will require manual step after database schema is migrated to newer version. 

We want to have an option to switch Marquez code to read facets from `lineage_events` before the manual migration and switch to using `job_facets`, `dataset_facets` and `run_facets` tables after the migration. We cannot split it into separate releases, as migration is manual and asynchronous and we never know when the user runs it (please note users may run upgrade multiple versions in one run).

### Solution

* `job_facets_vew`, `dataset_facets_view` and `run_facets_view` will be created on the top of existing `lineage_events` table. 
* Marquez codebase will read facets from those tables.
* In the next PR, data will be migrated from `lineage_events` table to facets' tables and views will be replaced to point at `job_facets`, `dataset_facets` and `run_facets` tables. 
* Changelog file will be modified in Ol facets PR3. 

### Checklist


api/src/main/java/marquez/db/migrations/V56_1__FacetViews.java
```diff
@@ -0,0 +1,157 @@
...
+import org.jdbi.v3.core.Jdbi;
+
+/**
+ * Dataset migration written in Java because SQLs used here will be reused in next migration when
+ * processing content of `lineage_events` table to facets' tables.
+ */
+@Slf4j
+public class V56_1__FacetViews implements JavaMigration {
+
+  public static final String DATASET_FACETS_DEFINITION =
```
pawel-big-lebowski, line 20: Migration is written as Java code because facet definitions will be reused in a migration that writes data to facets table.

docker/init-db.sh
```diff
@@ -8,7 +8,7 @@
 set -eu
 
 psql -v ON_ERROR_STOP=1 --username "${POSTGRES_USER}" > /dev/null <<-EOSQL
-  CREATE USER ${MARQUEZ_USER};
+  CREATE USER ${MARQUEZ_USER} SUPERUSER;
```
pawel-big-lebowski, line 11: @wslulciuc @collado-mike @mobuchowski This line is worth looking at and perhaps a discussion. 

As long we're using PSQL 12, super user is required to install `uuid-ossp` extension. We can remove it once upgrading PostgreSQL version to 13. 

Additionally, the migration code should work in a way such that it is allowed to run externally sql:
```
CREATE EXTENSION IF NOT EXISTS "uuid-ossp"
```
and run migration with no extra privileges. 

The CHANGELOG info about this will be included in the PR3.

wslulciuc (changes requested):
We'll have to update the copyright (`2022` -> `2023`):

```

/*
 * Copyright 2018-2023 contributors to the Marquez project
 * SPDX-License-Identifier: Apache-2.0
 */
```

wslulciuc (changes requested):
We can remove [metadata.json](https://github.com/MarquezProject/marquez/blob/ol-facets/PR2-read-data-from-views/metadata.json) as it's not used.

docker/init-db.sh
wslulciuc: It'd be great to [upgrade](https://github.com/MarquezProject/marquez/issues/2364) our postgres version, but we'll want to decouple the upgrade from the migration. Meaning, there's no way arounds having to require users to run the migration as `superuser`. Since requiring `superuser` is only for the migration, providing migration steps in the changelog would be enough.

pawel-big-lebowski:
> We can remove [metadata.json](https://github.com/MarquezProject/marquez/blob/ol-facets/PR2-read-data-from-views/metadata.json) as it's not used.

CI is failing when a file is removed. 
```
744bfdcfcb2e   marquezproject/marquez:0.29.0   "./wait-for-it.sh ap…"   About a minute ago   Exited (1) About a minute ago               seed-marquez-with-metadata

java.io.FileNotFoundException: metadata.json (Is a directory)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:216)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
	at com.fasterxml.jackson.core.JsonFactory.createParser(JsonFactory.java:1029)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3514)
	at marquez.cli.SeedCommand.loadMetadata(SeedCommand.java:150)
	at marquez.cli.SeedCommand.run(SeedCommand.java:135)
	at io.dropwizard.cli.Cli.run(Cli.java:78)
	at io.dropwizard.Application.run(Application.java:94)
	at marquez.MarquezApp.main(MarquezApp.java:59)
```

marquez.dev.yml
```diff
@@ -9,7 +9,7 @@ server:
 
 db:
   driverClass: org.postgresql.Driver
-  url: jdbc:postgresql://${POSTGRES_HOST:-localhost}:5432/marquez
+  url: jdbc:postgresql://postgres:5432/marquez
```
pawel-big-lebowski, line 12: revert line change from PR #2358, which causes Circle CI to fail.

pawel-big-lebowski:
> We can remove [metadata.json](https://github.com/MarquezProject/marquez/blob/ol-facets/PR2-read-data-from-views/metadata.json) as it's not used.

Not sure why anything related to `metadata.json` was in this PR. After merging PR1 and rebasing this one, it's no longer present.

docker/init-db.sh
collado-mike: Hmm, requiring the Postgres upgrade and/or running the migration process as a superuser seems like a serious hurdle. What's the principal gain the user gets from this? I.e., why do the facet tables require a UUID primary key to begin with? AFAIK, we only join with those tables based on the foreign keys (run_uuid, job_uuid, dataset_uuid). If we get rid of the primary keys on those facet records, we don't have to worry about new extensions or postgres upgrades and I don't know if there's any downside to eliminating them.

api/src/main/java/marquez/db/DatasetVersionDao.java
```diff
@@ -168,10 +168,10 @@ WITH selected_dataset_versions AS (
           FROM selected_dataset_versions dv
           LEFT JOIN runs_input_mapping rim
                ON rim.dataset_version_uuid = dv.uuid
-      ), selected_dataset_version_events AS (
-          SELECT dv.uuid, dv.dataset_name, dv.namespace_name, dv.run_uuid, le.event_time, le.event
+      ), selected_dataset_version_facets AS (
+          SELECT dv.uuid, dv.dataset_name, dv.namespace_name, dv.run_uuid, df.lineage_event_time, df.facet
           FROM selected_dataset_version_runs dv
-          LEFT JOIN lineage_events le ON le.run_uuid = dv.run_uuid
+          LEFT JOIN dataset_facets_view df ON df.dataset_uuid = dv.dataset_uuid
```
collado-mike, line 174: The original query returned facets that were attached to the specific dataset version that is being queried. This query returns all facets that are attached to the dataset. This will include things like data quality facets on previous or newer versions of the dataset, which may not reflect the state of the selected dataset version. We need to be able to attach and query facets to specific dataset versions.

docker/init-db.sh
pawel-big-lebowski: Uuid as a primary key has been proposed in proposal #2076 some time ago before I started working on that.
For me it looked like some general pattern in Marquez, that we create primary keys for tables so that database schema can be easily extended in future (easy joining of tables). For these tables, primary key cannot be derived from other columns. But I don't see any blocker in removing those columns.

api/src/main/java/marquez/db/DatasetVersionDao.java
pawel-big-lebowski: Great remark, I've added missing join on `run_uuid`.

docker/init-db.sh
wslulciuc: Unfortunately, yes the migration would be a hurdle for users running Marquez with postgres `12` or below. As @pawel-big-lebowski mentioned, we do have a general pattern for defining a `uuid` (=**PK**) column for all tables (mapping tables excluded ex: `dataset_fields_tag_mapping`). It would be great to keep this convention as we don't know how the facet tables will be used or extended. But, outside breaking convention, I don't see a major downside.

collado-mike: We can still append a new facet for the same run even without the UUID. The primary key is principally useful for fetching a specific record or referencing a specific record from another table. Neither of those is the case here. 

Arguing that we don't know how the facet tables will be used doesn't make sense either, as we can always add the UUID column after the fact. When there's a reason to introduce the primary key, we can argue that the extra work is worth the effort. For now, the current migration requires a very substantial change in the system with no benefit. Following the pattern simply for the sake of the pattern isn't a good reason to require these changes to the system or the deployment process that users are following.

api/src/main/java/marquez/db/DatasetVersionDao.java
```diff
@@ -211,10 +207,10 @@ WITH selected_dataset_versions AS (
           FROM selected_dataset_versions dv
           LEFT JOIN runs_input_mapping rim
                ON rim.dataset_version_uuid = dv.uuid
-      ), selected_dataset_version_events AS (
-          SELECT dv.uuid, dv.dataset_name, dv.namespace_name, dv.run_uuid, le.event_time, le.event
+      ), selected_dataset_version_facets AS (
+          SELECT dv.uuid, dv.dataset_name, dv.namespace_name, dv.run_uuid, df.lineage_event_time, df.facet
           FROM selected_dataset_version_runs dv
-          LEFT JOIN lineage_events le ON le.run_uuid = dv.run_uuid
+          LEFT JOIN dataset_facets_view df ON df.dataset_uuid = dv.dataset_uuid AND df.run_uuid = dv.run_uuid
```
collado-mike, line 213: The `selected_dataset_version_facets` CTE is expensive to create because we need to join on the `runs_input_mapping` table to find all of the runs that read from or wrote to the selected dataset version. Rather than going through this extra hop, can we add the dataset version UUID to the facet table so we can simply query all facets that apply to a specific dataset version?

pawel-big-lebowski: Yeah ;-) I had similar thoughts on this but the counterargument in my head was to avoid doing endlessly more & more in a single PR. 

I think you're right and there won't be better moment to fix this.

pawel-big-lebowski: Yeah ;-) I had similar thoughts on this but the counterargument in my head was to avoid doing endlessly more & more in a single PR.

I think you're right and there won't be better moment to fix this.
I've included it in a second commit within this PR.

collado-mike (approved):
LGTM. There's a comment on the Dataset facet retrieval, but I think it's an optimization and can be done now or later.

wslulciuc (approved):
LGTM 💯 🚀 🥇

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -76,22 +76,22 @@ WITH selected_datasets AS (
...
           FROM selected_datasets d
           INNER JOIN dataset_versions dv ON dv.uuid = d.current_version_uuid
           LEFT JOIN runs_input_mapping rim ON dv.uuid = rim.dataset_version_uuid
           LEFT JOIN LATERAL (
-              SELECT run_uuid, event_time, event FROM lineage_events
-              WHERE run_uuid = rim.run_uuid
-          ) e ON e.run_uuid = rim.run_uuid
+              SELECT dataset_uuid, run_uuid, lineage_event_time, facet FROM dataset_facets_view
+              WHERE dataset_uuid = dv.dataset_uuid AND run_uuid = rim.run_uuid
+         ) df ON df.run_uuid = rim.run_uuid
```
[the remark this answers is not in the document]
pawel-big-lebowski: PR #2407

Issue #2230, opened 2022-11-04
Only first job context is taken into consideration.


JDarDagran:
## Problem
Job context is a structure that serves as code location / SQL container to show them in Marquez UI. Job context upsert takes only checksum on context's body on conflict. This means that when e.g. at the start and the end of job the context is different there would be 2 different entries in `job_context` table for this job. That still might be ok, however this has its result in exposing in API only first captured context which means if you don't send SqlJobFacet in the START event you won't see it even if you send it in the COMPLETE event.

## Solutions
I foresee couple of ways to solve this problem:
1. Update `job_context_uuid` when upserting into `runs` table. This will result in getting only most recent context exposed which might be acceptable but probably not.
2. Add some custom logic to merge arrays when context relates to the same run (or job?).
3. Merge contexts in API. This would change run <--> job_context relation to 1-to-many.
4. Change structure of `job_contexts` table: replace `context` column with 3 following: `code_location_type`, `code_location_url`, `sql` which would be filled on upsert. Some concatenation would still be needed probably.

JDarDagran:
The other solution proposed by @wslulciuc  is to use facets directly and deprecate `job_context`.

mobuchowski:
I agree, those features can be fetched from facets directly. If there's no other uses for `job_contexts` a lot of code could be removed.

wslulciuc:
(Sorry for the late reply, @JDarDagran). The `jobs.context` has been deprecated in favor of using [job facets](https://accounts.google.com/AccountChooser/signinchooser?continue=https%3A%2F%2Faccounts.google.com%2Fo%2Fsaml2%2Finitsso%3Fidpid%3DC03o15ghc%26spid%3D104411315836%26forceauthn%3Dfalse%26from_login%3D1%26as%3DY3mf4An3vKPFOHWe3ELsRA&ltmpl=popup&btmpl=authsub&scc=1&oauth=1&flowName=GlifWebSignIn&flowEntry=AccountChooser) defined by OpenLineage (see https://github.com/MarquezProject/marquez/issues/910). Before OpenLineage, Marquez used key/value string pairs to annotate datasets, jobs, etc with custom metadata (think model extensibility). With the introduction/adoption of the standard, we no longer need to maintain our own extensibility model but rather leverage the extensibility model defined by OpenLineage.

PR #2373, opened 2023-01-20
Remove job context.

JDarDagran:
Signed-off-by: Jakub Dardzinski <kuba0221@gmail.com>

Add call to API in frontend.
Add tests.

Signed-off-by: Jakub Dardzinski <kuba0221@gmail.com>

### Problem

Closes: #2230 

### Solution

This PR removes use of job context. It also adds 2 endpoints for job/run facets per run. Those are called from web components to replace job context with `SQLJobFacet`.

There is also a lot of code removal. It appears that plenty of code was marked as uncovered due to `@NonNull` annotation and caused issues to achieve code coverage goal. According to [lombok docs](https://projectlombok.org/features/NonNull) I've added `lombok.nonNull.exceptionType=JDK` to skip coverage checks on those annotations.

### Checklist


api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -117,6 +119,34 @@ void createLineageEvent(
   LIMIT :limit""")
   List<LineageEvent> getAllLineageEventsAsc(ZonedDateTime before, ZonedDateTime after, int limit);
 
+  @SqlQuery(
+      """
+    SELECT
+      run_uuid,
+      JSON_AGG(event -> 'run' -> 'facets' ORDER BY event_time) AS facets
```
pawel-big-lebowski, line 126: if we were able to merge #2355, it would be great to make use of `run_facets_view`.

api/src/main/java/marquez/service/models/Facets.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.UUID;
+import javax.annotation.Nullable;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+
+@EqualsAndHashCode
+@ToString
+public final class Facets {
```
pawel-big-lebowski, line 18: If it is run related, then consider calling it `RunFacets`.

api/src/main/resources/marquez/db/migration/V56__alter_job_versions_to_drop_job_context_uuid.sql
```diff
@@ -0,0 +1,3 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+ALTER TABLE job_versions DROP COLUMN job_context_uuid;
```
mobuchowski, line 3: Let's drop column in the next version - this would prevent rollback if the feature would be buggy.

wslulciuc (changes requested):
@JDarDagran I think adding an endpoint for obtaining facets for a given run ID is a great addition to the Marquez API, but can we introduce the endpoint in a separate PR (to limit scope)? I've also provided some feedback on the runs facet endpoint.

api/src/main/java/marquez/api/FacetResource.java
```diff
@@ -0,0 +1,60 @@
...
+    super(serviceFactory);
+    this.openLineageDao = openLineageDao;
+  }
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Produces(APPLICATION_JSON)
+  @Path("/run/{id}")
```
wslulciuc, line 42: I would add the endpoint to obtain facets for a given run to [`RunResource`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/RunResource.java#L31). That is, to [receive a run](https://marquezproject.github.io/marquez/openapi.html#operation/getRun), we have the endpoint:

```
GET /api/v1/jobs/runs/{id}
```

Similarly, to obtain facets for a given runs, I would follow the same pattern:

```
GET /api/v1/jobs/runs/{id}/facets
```

JDarDagran:
> @JDarDagran I think adding an endpoint for obtaining facets for a given run ID is a great addition to the Marquez API, but can we introduce the endpoint in a separate PR (to limit scope)?


@wslulciuc Normally I would agree. However, if we don't add job facet endpoint in this PR we have no viable source of data to replace job context with. The closest one would be `/api/v1/events/lineage` but:
1. it does not allow to filter per run id
2. already has performance issues. It would add more if we'd need to filter results from the endpoint to match given run id (as per 1.)
For that reason I think this PR would less readable + less performant when introducing the endpoints in separate PR.

I do agree with your feedback on run facets endpoint, will apply to it.

api/src/main/java/marquez/db/OpenLineageDao.java
wslulciuc: @JDarDagran we were able to ;)

tito12:
Also I am getting error on Graph view - "Something went wrong while fetching job facets". Maybe I don't have some data?

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -437,4 +438,35 @@ WITH RECURSIVE job_names AS (
...
+  /**
+   * @param runUuid
+   */
+  @SqlQuery(
+      """
+      SELECT
+        run_uuid,
+        JSON_AGG(facet ORDER BY lineage_event_time) AS facets
+      FROM
+      run_facets_view
```
pawel-big-lebowski, line 451: Shouldn't this be a part of `RunFacetsDao` ?

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -437,4 +438,35 @@ WITH RECURSIVE job_names AS (
...
+    """)
+  RunFacets findRunFacetsByRunUuid(UUID runUuid);
+
+  @SqlQuery(
+      """
+            SELECT
+                run_uuid,
+                JSON_AGG(facet ORDER BY lineage_event_time) AS facets
+            FROM
+            job_facets_view
```
pawel-big-lebowski, line 465: Same as above?

pawel-big-lebowski (approved):
I cannot speak for webbev part, but the backend part seems good to me.
I've included some minor comments.
I think Java client is missing Run/Job facets added in service package, however this can be added later on in separate PR as this one already got pretty big. 

🚀 👍 💯

api/src/main/java/marquez/db/RunDao.java
JDarDagran: Done.

api/src/main/java/marquez/db/RunDao.java
JDarDagran: Done.

wslulciuc (approved):
Thanks for the clean up @JDarDagran! 🥇 💯 🚀

tito12 (approved):
Looks nice :)

Issue #2320, opened 2022-12-14
Dataset metrics facets aren't being returned by GET Datasets API


conorbev:
I've tried making an `InputDataset` using the Python client like the following:

```
pypi_schema = SchemaDatasetFacet(fields=[SchemaField(name="package_name", type="VARCHAR"),
                                         SchemaField(name="downloads", type="BIGINT"),
                                         SchemaField(name="year", type="INT"),
                                         SchemaField(name="month",type="INT")])
pypi_metrics = DataQualityMetricsInputDatasetFacet(rowCount=20, bytes=1024,
                                                   columnMetrics={
                                                       "package_name": ColumnMetric(
                                                           nullCount=0, distinctCount=20),
                                                       "downloads": ColumnMetric(nullCount=0, distinctCount=10),
                                                   },
                                                   )

pypi_downloads_i = InputDataset("bigquery", "pypi_data.downloads", {"schema": pypi_schema}, {"dataQuality": pypi_metrics})
```

if I then call GET `/namespaeces/bigquery/datasets/pypi_data.downloads` then the data returned includes the `schema` facet but not the data quality data.

GET response shown below:

```
{
  "id": {
    "namespace": "bigquery",
    "name": "pypi_data.downloads"
  },
  "type": "DB_TABLE",
  "name": "pypi_data.downloads",
  "physicalName": "pypi_data.downloads",
  "createdAt": "2022-12-12T17:05:57.203892Z",
  "updatedAt": "2022-12-13T18:27:18.976182Z",
  "namespace": "bigquery",
  "sourceName": "default",
  "fields": [
    {
      "name": "package_name",
      "type": "VARCHAR",
      "tags": [],
      "description": null
    },
    {
      "name": "downloads",
      "type": "BIGINT",
      "tags": [],
      "description": null
    },
    {
      "name": "year",
      "type": "INT",
      "tags": [],
      "description": null
    },
    {
      "name": "month",
      "type": "INT",
      "tags": [],
      "description": null
    }
  ],
  "tags": [],
  "lastModifiedAt": null,
  "lastLifecycleState": "",
  "description": null,
  "currentVersion": "3f462106-f5f7-4260-86d8-9b163ac8e217",
  "columnLineage": null,
  "facets": {
    "schema": {
      "fields": [
        {
          "name": "package_name",
          "type": "VARCHAR"
        },
        {
          "name": "downloads",
          "type": "BIGINT"
        },
        {
          "name": "year",
          "type": "INT"
[...]

PR #2417, opened 2023-02-11
Support inputFacets and outputFacets

pawel-big-lebowski:

### Problem

Marquez does not support `inputFacets` and `outputFacets` sent in Ol event. They're not saved nor exposed in api. 

Closes: #2320

### Solution

We do have `dataset_facets` table designed for this but we don't store `inputFacets` and `outputFacets` there. 
Although `inputFacets` and `outputFacets` contained within Openlineage event as dataset properties, we want to expose them in `api` as part of a run bcz they describe specific runs rather than datasets or dataset's versions. 

### Checklist

### DIFF
--- a/clients/java/src/main/java/marquez/client/models/Run.java
+++ b/clients/java/src/main/java/marquez/client/models/Run.java
@@ -8,6 +8,8 @@ package marquez.client.models;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.google.common.collect.ImmutableMap;
 import java.time.Instant;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import javax.annotation.Nullable;
@@ -27,6 +29,8 @@ public final class Run extends RunMeta {
   @Nullable private final Long durationMs;
   @Nullable private final Instant endedAt;
   @Getter private final Map<String, Object> facets;
+  @Getter private final List<InputDatasetVersion> inputDatasetVersions;
+  @Getter private final List<OutputDatasetVersion> outputDatasetVersions;
 
   public Run(
       @NonNull final String id,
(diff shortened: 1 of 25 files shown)


clients/java/src/main/java/marquez/client/models/Run.java
```diff
@@ -39,7 +43,9 @@ public Run(
       @Nullable final Instant endedAt,
       @Nullable final Long durationMs,
       @Nullable final Map<String, String> args,
-      @Nullable final Map<String, Object> facets) {
+      @Nullable final Map<String, Object> facets,
+      @Nullable final List<RunDatasetFacets> inputFacets,
+      @Nullable final List<RunDatasetFacets> outputFacets) {
```
wslulciuc, line 48: The class `Run` defines a list of `inputVersions` and `inputVersions` as `DatasetVersionId` entries. Given that a class `DatasetVersionId` is needed for `RunDatasetFacets` (along with the in/out facets) can we define a class `DatasetVersionIdAndFacets` as:

```java
class DatasetVersionIdAndFacets {
  final DatasetVersionId datasetVersionId;
  final ImmutableMap<String, Object> facets;
}
```

Then, update [`Run`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/service/models/Run.java) to use `DatasetVersionIdAndFacets` for `inputVersions` and `outputVersions`?

```java
public class Run {
  .
  final List<DatasetVersionIdAndFacets> inputVersions;
  List<DatasetVersionIdAndFacets> outputVersions;
  .
  .
```

It just seems `RunDatasetFacets` is really a `DatasetVersionId` with facets?

wslulciuc: Mind also updating the [`spec/openapi.yml`](https://github.com/MarquezProject/marquez/blob/main/spec/openapi.yml) with the additional properties?

pawel-big-lebowski: @wslulciuc  I agree that merging these two entries is fine, but I would like to avoid `And` in class name. Could we stay within `RunDatasetFacets` which contain both dataset version id's and facets? I've modified the code to return Run in that form. 

Additionally, I've included changes in spec. 
Screenshot 2023-03-03 at 09 29 36

wslulciuc (approved):
@pawel-big-lebowski, approving as to not block the PR based on our offline discussion (i.e. you're cool with my [suggested](https://github.com/MarquezProject/marquez/pull/2417#discussion_r1155714461) naming). Great work, and merge when ready 👍

PR #2430, opened 2023-02-21
This fixes the communication of the Web UI component with Marquez AP…

thomas-delrue:
…I service in the Helm Chart

### Problem

Deploying the current Helm Chart in main results in a broken Marquez Web Component because of a misconfiguration in the Helm Chart. In the container spec in the Web deployment.yaml file two environment variables are set, i.e. MARQUEZ_HOST and MARQUEZ_PORT, for the instance to be able to communicate with the Marquez API service in the K8s cluster.

The MARQUEZ_PORT is erroneously set to 5000, .Values.marquez.port, the port number used internally in the container for the API, but the Marquez API service is exposed via port number 80, .Values.service.port,  within the K8s cluster

### Solution

Small fix: define the value for MARQUEZ_PORT as .Values.service.port, i.e. 80. That way Marquez Web component can communicate via http://marquez:80 with the API backend.

### Checklist


wslulciuc (approved):
Good catch, @thomas-delrue. Thanks!

PR #2448, opened 2023-03-06
Removed jobs_fqn table and moved FQN into jobs directly in order to enforce unique constraints

collado-mike:
### Problem
The introduction of parent jobs and the `jobs_fqn` table intended to allow Marquez to support jobs that had the same name, but were triggered by different parents (e.g., a Spark job fired by different Airflow DAGs). The `jobs` table tracked the simple name of the job, while the `jobs_fqn` table tracked the fully qualified name (FQN). In addition, the `jobs_fqn` table became responsible for tracking the FQN of symlinked jobs, as it was too expensive to determine the new FQN of a job by following symlinks at query time. Instead, the FQN of a symlinked job is updated when the symlink is created so we return only the FQN of the symlink target rather than the FQN of the original job. 

Unfortunately, this means that neither the `jobs` table nor the `jobs_fqn` table can enforce the uniqueness constraint we had on the fully qualified name of a job. Thus, in production, we see errors like the following when trying to load a job by its name:
```
java.lang.IllegalStateException: Multiple values for optional: ['JobRow(uuid=b971d547-ea9d-44c1-908f-9dcc14faba98, type=BATCH, createdAt=2022-12-10T10:01:56.653991Z, updatedAt=2022-12-10T10:01:56.653991Z, namespaceName=...']
```

In particular, this happens on two occasions when receiving Airflow OpenLineage events:
1. We receive a `FAIL` event with no start event - the parent facet of the run is omitted, so Marquez creates a job with no parent, but the same FQN
2. We receive a `FAIL` event _prior_ to the `START` event - usually, this happens when requests are queued by the load balancer or sometimes when the `START` event itself is particularly large and deserializing takes longer than deserializing the `FAIL` event.

### Solution

This change eliminates the `job_fqn` table and reestablishes the uniqueness constraint on the `jobs` table's `name` column. It also adds a `simple_name` column to the table, which is used by the view to return the column of the same name. Tests for the two cases mentioned above are added to ensure we can handle Airflow events that omit the parent facet.

The `jobs_view` is also updated to omit symlinked jobs so that the read queries no longer have to omit them. `aliases` are moved from the `jobs_fqn` table to the `jobs` table so old job names can still be found. 
### Checklist

### DIFF
--- a/api/src/main/java/marquez/api/JobResource.java
+++ b/api/src/main/java/marquez/api/JobResource.java
@@ -187,8 +187,7 @@ public class JobResource extends BaseResource {
             .findJobByName(namespaceName.getValue(), jobName.getValue())
             .orElseThrow(() -> new JobNotFoundException(jobName));
 
-    // Should be simple name from `jobs_fqn`.
-    jobService.delete(namespaceName.getValue(), job.getSimpleName());
+    jobService.delete(namespaceName.getValue(), job.getName().getValue());
     return Response.ok(job).build();
   }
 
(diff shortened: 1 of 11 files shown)


wslulciuc (approved):
Great work, @collado-mike! Make sure to open a follow up issue to remove the `jobs_fqn` table, otherwise 👍

PR #2457, opened 2023-03-27
Fix job update SQL to correctly use simple_name for job updates

collado-mike:
### Problem

In #2448 we updated the job update logic to support adding a parent to an existing job. However, there was a bug in the update SQL that set the `simple_name` field to the fully qualified name instead. This fix addresses that bug and adds an assertion to the test that covers this case.

### Checklist


wslulciuc (approved):
LGTM 👍

PR #2384, opened 2023-01-28
UI: Selection on graph with edges. Status for jobs/datasets

tito12:
### Problem

According to the issue we are not showing status for jobs and datasets in way we need - based on last quality facets on for datasets and last 14 runs for jobs. This logic should be update, to show this info in correct way. 
Also path of selected node and connected to it - is not highligted. 
We need to show status of runs and quality facets also in tables with those data, based on the same logic with last 14 records. 

Closes: https://github.com/MarquezProject/marquez/issues/2169

### Solution

This PR provide those changes

image

### Checklist

- [x] You've [signed-off](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#sign-your-work) your work
- [ ] Your changes are accompanied by tests (_if relevant_)
- [ ] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [ ] You've updated the [`CHANGELOG.md`](https://github.com/MarquezProject/marquez/blob/main/CHANGELOG.md#unreleased) with details about your change under the "Unreleased" section (_if relevant, depending on the change, this may not be necessary_)
- [ ] You've versioned your `.sql` database schema migration according to [Flyway's naming convention](https://flywaydb.org/documentation/concepts/migrations#naming) (_if relevant_)
- [ ] You've included a [header](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#copyright--license) in any source code files (_if relevant_)

PR #2458, opened 2023-03-28
UI: Change color for selected node and edges on graph

tito12:
### Problem

Currently selected node and edges are not stand out from others too much.
Due to proposition from last comment from this PR https://github.com/MarquezProject/marquez/pull/2384

### Solution

Selected node and edges highligthed with primary (green) color which more visible than previous.

image

### Checklist

### DIFF
--- a/web/src/components/lineage/components/edge/Edge.tsx
+++ b/web/src/components/lineage/components/edge/Edge.tsx
@@ -52,7 +52,7 @@ class Edge extends React.Component<EdgeProps> {
             data={edge.points}
             x={(d, index) => (index === 0 ? d.x + 20 : d.x - 25)}
             y={d => d.y}
-            stroke={edge.isSelected ? theme.palette.common.white : theme.palette.secondary.main}
+            stroke={edge.isSelected ? theme.palette.primary.main : theme.palette.secondary.main}
             strokeWidth={1}
             opacity={1}
             shapeRendering='geometricPrecision'
@@ -66,7 +66,7 @@ class Edge extends React.Component<EdgeProps> {
             y={edge.y - ICON_SIZE / 2}
             width={ICON_SIZE}
             height={ICON_SIZE}
-            color={edge.isSelected ? theme.palette.common.white : theme.palette.secondary.main}
+            color={edge.isSelected ? theme.palette.primary.main : theme.palette.secondary.main}
           />
         ))}
       </>
--- a/web/src/components/lineage/components/node/Node.tsx
+++ b/web/src/components/lineage/components/node/Node.tsx
@@ -62,7 +62,7 @@ class Node extends React.Component<NodeProps> {
               style={{ cursor: 'pointer' }}
               r={RADIUS}
               fill={isSelected ? theme.palette.secondary.main : theme.palette.common.white}
-              stroke={isSelected ? theme.palette.common.white : theme.palette.secondary.main}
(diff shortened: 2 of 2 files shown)


tito12:
@howardyoo please take a look and approve if it's what you expect :)

howardyoo:
> @howardyoo please take a look and approve if it's what you expect :)

Tested it and verified! Nice work - the UI now looks more helpful in terms of understanding upstream and downstream dependencies of the job/dataset that was selected. Thank you, @tito12 !

howardyoo (approved):
LGTM, Approved.

wslulciuc:
@tito12, great work! What will the lineage path color be for failed jobs? That is, will the path be red for jobs with a failed state?

howardyoo:
> @tito12, great work! What will the lineage path color be for failed jobs? That is, will the path be red for jobs with a failed state?

According to this PR, there is no color based on the job's status. However, I think that may be a good idea, to distinguish status based on the color. @tito12 , how difficult would it be to include what Willy described within this PR?

tito12:
First idea was about to make green and red color due state, but it was't possible with current api. That's why I used neutral color with bold style previously and due proposition added green color for selected node and edges. Currently it's not possible to highlight all states on graph with state colors to include it here.  @wslulciuc @howardyoo

howardyoo:
> First idea was about to make green and red color due state, but it wasn't possible with current api. That's why I used neutral color with bold style previously and due proposition added green color for selected node and edges. Currently it's not possible to highlight all states on graph with state colors to include it here. @wslulciuc @howardyoo

Hi @tito12 ,
in the Marquez UI, when displaying the lineage graph, here are the list of API calls:
image

In these calls, notice the 'runs' api call, which retrieves the 'run' information of the currently selected job (https://marquezproject.github.io/marquez/openapi.html#operation/getRun)
image

The run json object has an attribute called 'STATE' which is `COMPLETED` at this time.
The list run-states are mentioned here: https://github.com/MarquezProject/marquez/blob/60581e37282c68a75b9f7bf1deb740d90eebedf0/api/src/main/java/marquez/common/models/RunState.java#L8
which are:

  NEW,
  RUNNING,
  COMPLETED,
  ABORTED,
  FAILED

So, I guess we might be able to use this state to indicate whether the job in question can be colored, so I believe the API does exist to extract that.

As for the datasets (not jobs), the data available might be different.
https://marquezproject.github.io/marquez/openapi.html#operation/deleteDataset is the dataset, and
image
and my take is either we use `lifecycleState` in the version of the dataset (could be CREATE, DROP, TRUNCATE, etc.)
or maybe use `createdByRun -> outputVersions.state` that's shown below.
image

I guess the dataset keeps the 'createdByRun' to show information of the run (job)'s result - so the idea might be that
1) if the dataset has 'createdByRun', look at its `outputVersions.state`.
2) if the result is `COMPLETED`, display it normally
3) however, if the result is `ABORTED` or `FAILED`, then show the dataset as RED (or whatever failure color). This is due to the inference that the job run that was responsible to output this dataset has failed, thus the dataset might also have become invalid.

What do you think?

howardyoo:
I've investigated a little bit into this, but for now, it seems like the way the data is coming from the API, and how the lineage graph does NOT have anything related to 'status' suggests that it is not possible for it to render any `status` (like error conditions) to it. That should be considered as another improvement for the future.

PR #2461, opened 2023-03-29
Update SQL in backfill script for facet tables to improve performance

collado-mike:
### Problem

The backfill script to populate the facet tables relies on some SQL that consistently times out on large Marquez installations. E.g., the initial lock query relies on ordering the `lineage_events` table, which takes more than 10 minutes (I don't know how long it actually takes because I stopped increasing the statement timeout at this point) just to determine that the installation is too large to apply the migration. Additionally, populating the facets takes more than 8 mins per batch, even when reducing the batch size from 10,000 to 1,000. In an installation of ~30M records, the migration at that rate will take multiple weeks to complete.

### Solution

The SQL changes here introduce a temp table that tracks the runs that need to be migrated and use that table to handle sorting the runs by time. Using the `run_uuid` allows us to take advantage of the index on `lineage_events` so that fetching the events happens much faster. As there is typically two events per run, it effectively doubles the batch size. Using the default chunk size of 10,000, with an effective batch size of ~20,000, the query now completes in about 45 seconds. 

### Checklist


wslulciuc (approved):
🤩

PR #2464, opened 2023-03-31
Update v61 migration to handle duplicate job names before unique constraint

collado-mike:
### Problem

The v61 migration reapplies a unique constraint on job name and namespace without regard to parent job. Unfortunately, there are some cases when job FQNs are duplicated (mostly due to the issue fixed in https://github.com/MarquezProject/marquez/pull/2097 , but also including cases when a job previously had no parent and now has a parent, but the same FQN). This update to the migration renames jobs that have been symlinked to point to newer versions of themselves so that the job FQN doesn't conflict and the unique constraint can be applied.

Any installations that have already applied this migration will not see any new operations on their data, but installations that have duplicates will need this fix for the migration to successfully complete.

I tested this on an installation with ~200,000 job names and was able to complete the migration successfully.

### Checklist

### DIFF
--- a/api/src/main/resources/marquez/db/migration/V61__unique_job_fqn_index.sql
+++ b/api/src/main/resources/marquez/db/migration/V61__unique_job_fqn_index.sql
@@ -1,10 +1,45 @@
+WITH fqn AS (
+  SELECT f.uuid,
+         f.job_fqn AS name,
+         f.namespace_name,
+         j.name    AS simple_name,
+         j.parent_job_uuid,
+         f.parent_job_name::text,
+         j.symlink_target_uuid
+  FROM jobs_fqn f,
+       jobs j
+  WHERE j.uuid = f.uuid
+)
 UPDATE jobs SET symlink_target_uuid=q.target_uuid
 FROM (
          SELECT j.uuid, j.namespace_name, j.name, j.simple_name, jv.uuid AS target_uuid, jv.simple_name
-         FROM jobs_view j INNER JOIN jobs_view jv ON j.namespace_name=jv.namespace_name AND j.name=jv.name AND j.simple_name != jv.simple_name
+         FROM fqn j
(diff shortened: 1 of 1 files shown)


api/src/main/resources/marquez/db/migration/V61__unique_job_fqn_index.sql
```diff
@@ -1,10 +1,45 @@
...
+         j.name    AS simple_name,
+         j.parent_job_uuid,
+         f.parent_job_name::text,
+         j.symlink_target_uuid,
+         j.created_at
+  FROM jobs_fqn f,
+       jobs j
+  WHERE j.uuid = f.uuid
+)
+UPDATE jobs SET name=(q.simple_name || '_' || q.row)
```
wslulciuc, line 36: This will set a job with a conflict to `{job_name}_{row_num}`?

wslulciuc (approved):
Your writeup in the PR description was helpful in understanding all that's going on in the migration script. Thanks! And, also, 👍

api/src/main/resources/marquez/db/migration/V61__unique_job_fqn_index.sql
collado-mike: Right, but `row_number() over (PARTITION BY j.namespace_name, j.name ORDER BY j.created_at) AS row` means that all jobs that have the same name will have an incrementing counter suffixed so we avoid the conflict. E.g., if we had three jobs with the FQN `a.b`, the names would be
1. `a.b_1`
2. `a.b_2`
3. `a.b_3`
Thus, we can add the uniqueness constraint because they'll all end up with different names.

wslulciuc: Sweet, thanks for clarifying and the examples 👍

Issue #1271, opened 2021-04-28
Proposal: Data Retention Functionality


KevinMellott91:
**Background**
Some organizations require data retention policies to be enforced to meet legal and/or regulatory requirements. This proposal provides a few ways that this functionality could be implemented within Marquez.

**Retention Scope**
Enabling data retention functionality will be an opt-in feature, and will be disabled by default. This ensures that existing deployments are unaffected, but provides the functionality to those who need it.

When targeting data for retention, the scope will be focused on dataset/job versions and job runs. Furthermore, the Marquez configurations will allow retention policies to be set individually for each type of entity (dataset version, job version, job run) to allow for flexible enforcement.

This will ensure that core entities (namespaces, sources, datasets, jobs) are not impacted and will not magically disappear from the system. Lineage events are also out of scope, so that we don't lose the ability to replay those events over time.

**Retention Behavior**
Retention rules can be configured to purge items that are older than a given timeframe AND/OR are X versions away from the current version. Other systems have implemented retention using one of these conditions OR the other; however, I think flexibility is required for this to work with many real world scenarios. 

Here is how this could be configured within a Marquez deployment.
```yaml
# Rules can be combined under a section to ensure both conditions are met prior to data being purged.
retention:
  enabled: true # Enables retention enforcement, for the entities configured below. Omit sections when retention is not desired.
  schedule: "0 3 * * *" # When deployed into Kubernetes, enforce retention policies on this schedule.
  datasetVersions:
    recentItemsToKeep: 10 # Purges any dataset version records older than the 10 most recent.
  jobVersions:
    daysToKeep: 365 # Purges any job version records older than 1 year.
  jobRuns: # Purges job run records that are both older than the 10 most recent AND older than 1 year.
    recentItemsToKeep: 10
    daysToKeep: 365
```

**Retention Enforcement**
Retention enforcement can be applied on a configurable schedule, using a Java container-based batch job. We can provide a script to execute this job via Docker to support local development, and also deploy this into Kubernetes as a [CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) as part of the Helm chart. The CronJob will execute on the schedule defined within the YAML configurations, and will purge the applicable records based on the retention policy rules.

As records are identified and purged, applicable information can be logged and incorporated into the metrics framework.

julienledem:
Hi Kevin, 
Thank you for the thoughtful proposal. This makes a lot of sense. 
Here are some amendments I would make.
[...]

PR #2486, opened 2023-05-11
Add db retention support

wslulciuc:
## Problem

Marquez does not have a configurable data retention policy. Therefore, users are left to periodically increase the storage size of their PostgreSQL instance as disk space fills up. Below, you'll find an example graph for the `FreeStorageSpace` metric in RDS (from a _real_ production deployment of Marquez in the wild). For every increase in storage, we observe a spike in freeable storage (I've lost count on how many times I've personally have done this). Closes: https://github.com/MarquezProject/marquez/issues/1271

Screen Shot 2023-05-15 at 9 37 01 AM

## Solution

To simplify the retention policy query executed by calling [`DbRetention.retentionOnDbOrError()`](https://github.com/MarquezProject/marquez/blob/feature/db-retention/api/src/main/java/marquez/db/DbRetention.java), we introduced migration [`V63__alter_tables_add_on_cascade_delete.sql`](https://github.com/MarquezProject/marquez/blob/feature/db-retention/api/src/main/resources/marquez/db/migration/V63__alter_tables_add_on_cascade_delete.sql) that updates all _foreign key_ constraints to add `ON DELETE CASCADE`; therefore, deleting all derived metadata (=child tables) from tables `datasets`, `jobs`, `runs`, and `lineage_events` when the retention policy is applied. The migration has the added advantage of defining clear relationships between metadata at the database level, as well as, self documenting the order of deletion. Below, we outlined performance considerations on the sql used for the retention policy, how it's applied and options for configuration.

### Retention Policy Performance Considerations

In most cases, for small to medium-size metadata (< `50GB`) stored in Marquez, the retention policy could be scheduled to be applied in a maintenance window, in a single transaction. But, as we evaluated the metadata stored in Marquez at [Astronomer](https://www.astronomer.io/)  (> `500GB`), we quickly realized that we needed to rethink this approach. Below, we outline the general sql structure applied during retention (per metadata object), and the configurable parameters enabling users to best tune the policy specific to their needs.   

```java
// We define a deletion function per metadata object that is called, per metadata object,
// in DbRetention.retentionOnDbOrError(); the deletion function has a return type of INT
// representing the number of rows deleted.
CREATE OR REPLACE FUNCTION delete_<METADATA-OBJECT>_older_than_x_days()
  RETURNS INT AS $$
DECLARE
  // To improve deletion performance (i.e. avoid deleting all rows at once), we delete rows
  // in batches; this divides the deletion process into smaller chunks; we made the
  // number of rows to delete per batch configurable.
  rows_per_batch INT := [NUMBER_OF_ROWS_PER_BATCH]
  rows_deleted INT;
  rows_deleted_total INT := 0;
BEGIN
  // [CREATE-TEMP-TABLES-If-NEEDED]
  //
  // We divide row deletion into smaller chunks by using a loop and limiting batch size via
  // the configurable 'rows_per_batch' parameter per iteration; this approach limits resource
  // impact and maintains overall database performance; the loop continues until all rows are deleted.
  LOOP
    WITH deleted_rows AS (
      DELETE FROM <METADATA-OBJECT>
        WHERE uuid IN (
          SELECT uuid
            FROM <METADATA-OBJECT>
           WHERE updated_at < CURRENT_TIMESTAMP - INTERVAL '<RETENTION-DAYS> days'
             FOR UPDATE SKIP LOCKED // Acquire row-level lock and skip any locked rows.
             // Use [TEMP-TABLES-If-NEEDED]
           LIMIT rows_per_batch
        ) RETURNING uuid
    )
    SELECT COUNT(*) INTO rows_deleted FROM deleted_rows;
    rows_deleted_total := rows_deleted_total + rows_deleted;
    // Exit when no more rows are deleted in a given iteration.
    EXIT WHEN rows_deleted = 0;
    // Pause to allow other transactions to run.
    PERFORM pg_sleep(0.1);
  END LOOP;
  RETURN rows_deleted_total;
END;
$$ LANGUAGE plpgsql;
```

### Applying Retention Policy

When invoking `DbRetention.retentionOnDbOrError()`, retention is applied in the following order (in an order of precedence from left-to-right):

```
jobs > job versions > runs > datasets > dataset versions > lineage events
```

1. Delete **jobs** from `jobs` table if `jobs.updated_at` **older than** `retentionDays`.
2. Delete **job versions** from `job_versions` table if `job_versions.updated_at` **older than** `retentionDays`; a job version will **not** be deleted if the job version is the current version of a given job.
3. Delete **runs** from `runs` table if `runs.updated_at` **older than** `retentionDays`; a run will **not** be deleted if the run is the current run of a given job version.
4. Delete **dataset** from `datasets` table if `datasets.updated_at` **older than** `retentionDays`; a dataset will **not** be deleted if the dataset is an input / output of a given job version.
5. Delete **dataset versions** from `dataset_versions` table if `dataset_versions.created_at` **older than** `retentionDays`; a dataset version will **not** be deleted if the dataset version is the current version of a given dataset version, or the input of a run.
6. Delete **lineage events** from `lineage_events` table if `lineage_events.event_time` **older than** `retentionDays`.

### Configuring Retention Policy

We propose Marquez support the following ways to config retention policies on metadata:

#### `OPTION` `1`: Config-based via `marquez.yml`

In `marquez.yml`, define the `dbRetention` configuration to enable the job `DbRetentionJob` to periodically apply a retention policy to the database on an interval. The default retention policy is `7` days, but this can be adjusted by overriding `retentionDays`. You can also override the run interval for `DbRetentionJob` by using `frequencyMins`.

To **enabled** a retention policy, define the `dbRetention` config in your `marquez.yml`:

```bash
# Adjusts retention policy
dbRetention:
  # Apply retention policy at a frequency of every 'X' minutes (default: 15)
  frequencyMins: ${DB_RETENTION_FREQUENCY_MINS:-15}
  # Maximum number of rows deleted per batch (default: 1000)
  numberOfRowsPerBatch:  ${DB_RETENTION_NUMBER_OF_ROWS_PER_BATCH:-1000}
  # Maximum retention days (default: 7)
  retentionDays: ${DB_RETENTION_DAYS:-7}
```

To **disable** retention policy, omit the `dbRetention` config from your `marquez.yml`:

```yaml
# Disables retention policy
# dbRetention: ...
```

#### `OPTION` `2`: CLI

To run an ad-hoc retention policy on your metadata, use the `db-retention` cmd:

```
       db-retention [--number-of-rows-per-batch NUMBEROFROWSPERBATCH] [--retention-days RETENTIONDAYS] [--dry-run] [-h] [file]

apply one-off ad-hoc retention policy directly to database

positional arguments:
  file                   application configuration file

named arguments:
  --number-of-rows-per-batch NUMBEROFROWSPERBATCH
                         the number of rows deleted per batch (default: 1000)
  --retention-days RETENTIONDAYS
                         the number of days to retain metadata (default: 7)
  --dry-run              only output an estimate of metadata deleted by the retention policy, without applying the policy on database (default: false)
```

```
$ java -jar api/build/libs/marquez-api.jar db-retention --retention-days 14 marquez.yml
INFO marquez.db.DbRetention - Applying retention policy of '30' days to jobs...
INFO marquez.db.DbRetention - Deleted '0' jobs in '22' ms!
INFO marquez.db.DbRetention - Applying retention policy of '30' days to job versions...
INFO marquez.db.DbRetention - Deleted '0' job versions in '30' ms!
INFO marquez.db.DbRetention - Applying retention policy of '30' days to runs...
INFO marquez.db.DbRetention - Deleted '4' runs in '124' ms!
INFO marquez.db.DbRetention - Applying retention policy of '30' days to datasets...
INFO marquez.db.DbRetention - Deleted '0' datasets in '22' ms!
INFO marquez.db.DbRetention - Applying retention policy of '30' days to dataset versions...
INFO marquez.db.DbRetention - Deleted '0' dataset versions in '28' ms!
INFO marquez.db.DbRetention - Applying retention policy of '30' days to lineage events...
INFO marquez.db.DbRetention - Deleted '0' lineage events in '15' ms!
```

### Improvements

* Estimate # number of rows / bytes deleted _before_ running the retention policy as a percentage.

### Checklist


JDarDagran:
I love the change very much. However, is there a reason why only 3 tables are included? My understanding is that those are consuming the majority of space.

wslulciuc:
@JDarDagran, that's a very good point that I didn't (but should have) elaborate on. Due to the migration [`V62__alter_tables_add_on_cascade_delete`](https://github.com/MarquezProject/marquez/blob/feature/db-retention/api/src/main/resources/marquez/db/migration/V62__alter_tables_add_on_cascade_delete.sql) (introduced as part of this PR), we only need the tables `datasets`, `jobs`, and `lineage_events`. Let me write up a section outlining the usage of this migration file and its importance.

api/src/main/java/marquez/MarquezConfig.java
```diff
@@ -11,15 +11,18 @@
...
 import marquez.graphql.GraphqlConfig;
+import marquez.jobs.DbRetentionConfig;
 import marquez.service.models.Tag;
 import marquez.tracing.SentryConfig;
 
 /** Configuration for {@code Marquez}. */
 @NoArgsConstructor
 public class MarquezConfig extends Configuration {
   private static final boolean DEFAULT_MIGRATE_ON_STARTUP = true;
+  private static final int DATA_RETENTION_IN_DAYS = 7;
```
tatiana, line 25: If we make 7 days the default DB retention behaviour, wouldn't this be considered a breaking change?
Could this lead to users who didn't read the changelog losing data while upgrading to a minor release / setting up new projects?

api/src/main/java/marquez/db/DbRetention.java
```diff
@@ -0,0 +1,112 @@
...
+import marquez.db.exceptions.DbRetentionException;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.statement.StatementException;
+
+@Slf4j
+public final class DbRetention {
+  private DbRetention() {}
+
+  /* Bulk delete batch size. */
+  private static final int BULK_DELETE_BATCH_SIZE = 1000;
```
tatiana, line 23: Should we expose this in the CLI as well, so users can control the `LIMIT`?
Was this default value set based on some empirical value?

tatiana (commented):
This is a very useful feature, @wslulciuc , it should help us clean unnecessary old data!

api/src/main/java/marquez/cli/DbRetentionCommand.java
```diff
@@ -0,0 +1,86 @@
...
+  @Override
+  public void configure(@NonNull final net.sourceforge.argparse4j.inf.Subparser subparser) {
+    super.configure(subparser);
+    subparser
+        .addArgument("--retention-days")
+        .dest(CMD_ARG_RETENTION_DAYS)
+        .type(Integer.class)
+        .required(false)
+        .setDefault(DEFAULT_RETENTION_DAYS)
+        .help("the number of days to retain metadata");
```
tatiana, line 56: Would it make sense to add, as part of the help, that this is a one-off run of the command (the user may assume this changes the config for them, not sure - people can be very creative :))

api/src/main/java/marquez/MarquezConfig.java
julienledem: that's a great point.

api/src/main/java/marquez/cli/DbRetentionCommand.java
```diff
@@ -0,0 +1,86 @@
...
+ * <pre>{@code
+ * java -jar marquez-api.jar db-retention --retention-days 14
+ * }</pre>
+ */
+@Slf4j
+public class DbRetentionCommand extends ConfiguredCommand<MarquezConfig> {
+  private static final String DB_SOURCE_NAME = "db-retention-source";
+
+  /* Default retain for metadata. */
+  private static final int DEFAULT_RETENTION_DAYS = 7;
```
julienledem, line 37: Shouldn't we use the same constants for the default retention in config and on the command line param?

api/src/main/java/marquez/db/DbRetention.java
julienledem: Also, this should have an explanation in the comment on why we need to limit this.

api/src/main/java/marquez/db/DbRetention.java
```diff
@@ -0,0 +1,112 @@
...
+                "Deleted '{}' rows in '{}' secs using retention policy: {}",
+                totalNumOfRowsDeleted,
+                stopwatch.elapsed().toSeconds(),
+                retentionQuery);
+            return totalNumOfRowsDeleted;
+          });
+    } catch (StatementException errorOnDbRetentionQuery) {
+      log.error("Failed to apply retention query: {}", retentionQuery);
+      // Propagate throwable up the stack.
+      throw new DbRetentionException(errorOnDbRetentionQuery.getCause());
```
julienledem, line 90: if you log before rethrowing you'll have double logs.
Also I don't understand why you don't chain the exception itself instead of the cause.
I think you should just throw here and add the query to the new message.
Something like:
`throw new DbRetentionException("Failed to apply retention query: " + retentionQuery errorOnDbRetentionQuery);`

api/src/main/java/marquez/db/DbRetention.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /** Returns {@code pk} for the provided {@code table}. */
+  private static String pkInQueryFor(@NonNull final String table) {
+    return LINEAGE_EVENTS.equals(table) ? Columns.RUN_UUID : Columns.ROW_UUID;
+  }
+
+  /** Returns {@code timestamp} for the provided {@code table}. */
+  private static String timestampInQueryFor(@NonNull final String table) {
+    return LINEAGE_EVENTS.equals(table) ? Columns.EVENT_TIME : Columns.CREATED_AT;
+  }
```
julienledem, line 111: I would create a nested enum here with the privateKey and timestamp fields for clarity.
```
public enum Table {
    DATASETS(Columns.ROW_UUID, Columns.CREATED_AT),
    JOBS(Columns.ROW_UUID, Columns.CREATED_AT),
    LINEAGE_EVENTS(Columns.RUN_UUID, Columns.EVENT_TIME);

    public final String privateKey;
    public final String timestamp;

    private Element(String privateKey, String timestamp) {
        this.privateKey = privateKey;
        this.timestamp = timestamp;
    }
}

File: api/src/main/java/marquez/jobs/DbRetentionConfig.java [hunk 7]
```diff
@@ -0,0 +1,18 @@
...
+
+package marquez.jobs;
+
+import lombok.Getter;
+import lombok.Setter;
+
+/** Configuration for {@link DbRetentionJob}. */
+public final class DbRetentionConfig {
+  public static final int DEFAULT_FREQUENCY_MINS = 15;
+  public static final int DEFAULT_RETENTION_DAYS = 7;
```
julienledem, line 14: 3rd constant for the default 😄

api/src/main/java/marquez/jobs/DbRetentionJob.java
```diff
@@ -0,0 +1,82 @@
...
+  protected void runOneIteration() {
+    try {
+      // Attempt to apply a database retention policy. An exception is thrown on failed retention
+      // policy attempts requiring we handle the throwable and log the error.
+      DbRetention.retentionOnDbOrError(jdbi, retentionDays);
+    } catch (DbRetentionException errorOnDbRetention) {
+      log.error(
+          "Failed to apply retention policy of '{}' days to database!",
+          retentionDays,
+          errorOnDbRetention);
```
julienledem, line 73: yes, the logging should happen at the top of the call stack like here.
(referring to previous comment)

julienledem (commented):
It is great to finally have this feature. I agree that we should probably be backwards compatible and not have it on by default.
Or possibly only for new databases? I don't know if we can do that.

api/src/main/java/marquez/MarquezConfig.java
wslulciuc: By default, no retention policy is applied to collected metadata in Marquez. I've introduced an [faq](https://github.com/MarquezProject/marquez/blob/feature/db-retention/docs/faq.md) and added _How do I configure a retention policy for metadata?_ section that will better explain how to configure a retention policy.

api/src/main/java/marquez/db/DbRetention.java
pawel-big-lebowski: Could each of the enum contain `retentionQuery`? Being able to see the whole SQL being executed improves the readability of the code, even if it contains some repeated sql code.

api/src/main/java/marquez/db/DbRetention.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /** Returns {@code pk} for the provided {@code table}. */
+  private static String pkInQueryFor(@NonNull final String table) {
+    return LINEAGE_EVENTS.equals(table) ? Columns.RUN_UUID : Columns.ROW_UUID;
+  }
+
+  /** Returns {@code timestamp} for the provided {@code table}. */
+  private static String timestampInQueryFor(@NonNull final String table) {
+    return LINEAGE_EVENTS.equals(table) ? Columns.EVENT_TIME : Columns.CREATED_AT;
```
pawel-big-lebowski, line 110: If we delete dataset that was created more than `X` days ago, we also delete its' lineage that doesn't meet the retention period defined. Aren't we? 

In other words: with a retention of 7 days, a dataset that was created 8 days ago, and is filled daily with new data, should not be deleted. Otherwise, we lose lineage before the retention time defined. 

I am not sure if this is working like that but I think it definitely makes sense to write a test that assures that.

pawel-big-lebowski (commented):
This is an important and awaited feature and its implementation deserves testing. We need to make sure that we delete rows that we aimed to delete and that we leave intact those should not be affected. Otherwise we only rely on foreign key depenendencies which are difficult to be verified.

I think we need tests for this PR for at least two reasons:
 * make sure retention mechanism does what we want (see my other comment with some doubts)
 * make sure future db schema changes (like adding new foreign key references) won't damage users data.

api/src/main/java/marquez/db/DbRetention.java
julienledem: It would make sense to me to start with deleting old job/dataset versions and delete job/datasets when they have no versions anymore?

api/src/main/java/marquez/cli/DbRetentionCommand.java
wslulciuc: Absolutely! So ok, got a bit too carried away with all these constants 😅

api/src/main/java/marquez/db/DbRetention.java
wslulciuc: I made the chunk size configurable for the CLI and retention job. For example:

```yaml
dbRetention:
  frequencyMins: ${DB_RETENTION_FREQUENCY_MINS:-15}
  # Maximum data chunk size that can be deleted per retention execution (default: 1000)
  chunkSize:  ${DB_RETENTION_CHUNK_SIZE:-1000}
  retentionDays: ${DB_RETENTION_DAYS:-7}
```

api/src/main/java/marquez/cli/DbRetentionCommand.java
wslulciuc: I can see how the cmd description can be misleading and allow for creative interpreation. I've updated the cmd usage to include 

```
       db-retention [--chunk-size CHUNKSIZE] [--retention-days RETENTIONDAYS] [-h] [file]

apply one-off ad-hoc retention policy directly to database

...
```

api/src/main/java/marquez/db/DbRetention.java
wslulciuc: Ok yeah, I agree that this will double log and make things harder to debug by polluting the logs with, essentially, the same error.

api/src/main/java/marquez/jobs/DbRetentionConfig.java
wslulciuc: There's now only a single reference: [`DbRetention.DEFAULT_RETENTION_DAYS`](https://github.com/MarquezProject/marquez/blob/feature/db-retention/api/src/main/java/marquez/db/DbRetention.java#L23)

api/src/main/java/marquez/db/DbRetention.java
wslulciuc: @pawel-big-lebowski, @julienledem: Very valid. The retention query should be smarter about when a job or dataset was last updated before applying a retention policy. In the case for a job: _when was the last recent run_? For a dataset, it may have been created past the allowed retention days, but _are new versions are still being created by job runs?_

I can easily make the adjustments to the retention query, but will keep the [`retention` config](https://github.com/MarquezProject/marquez/blob/feature/db-retention/marquez.example.yml#L31) the same with the plan to expand configuration options similar to what was proposed in https://github.com/MarquezProject/marquez/issues/1271.

api/src/main/java/marquez/db/DbRetention.java
wslulciuc: I avoided the enum all together by defining the SQL in-line for readability as suggested by @pawel-big-lebowski 👍

pawel-big-lebowski (approved):
I really like the retention tests and scenarios they cover and verify 🚀 👍 🏅 

The one thing that could be tested more is: 
 * loading event from `event_full`
 * applying retention 
 * verifying if other tables get cleared on cascade while not throwing an error. 

Leaving @wslulciuc the decision if this is worth the effort and approving PR.

wslulciuc:
Thanks you @julienledem, @pawel-big-lebowski, @tatiana and @JDarDagran for all of your feedback. I've added a detailed descripiton of the retention logic, support for dry runs, and extensive testing (@pawel-big-lebowski I'll follow up with a PR with your suggestions as this PR is getting way to big and the core logic has been covered in the current test suite).

Issue #1868, opened 2022-02-08
Define `OpenLineageResource(OpenLineageService)` constructor


wslulciuc:
When constructing a new `OpenLineageResource` object, a `ServiceFactory` is required therefore not allowing users to inject, possibly their own, `OpenLineageService` implementation. Let's define the constructor `OpenLineageResource(OpenLineageService)` and deprecate the usage of `OpenLineageResource(ServiceFactory)`.

fm100:
Every resources inherits `BaseResource` with a constructor getting `ServiceFactory`, and we may want to do some overall refactoring to create resourses only with the actual dependencies to be injected.

PR #2495, opened 2023-05-30
Ability to decode static metadata events 

pawel-big-lebowski:
### Problem

Static metadata events (known also as runless) where presented within in the proposal:
https://github.com/OpenLineage/OpenLineage/blob/main/proposals/1837/static_lineage.md

and is being introduced to Openlineage specification within the PR: https://github.com/OpenLineage/OpenLineage/pull/1880

Of course, we would like to implement it within Marquez. We would like to achieve this ultimate goal gradually and this PR contains the first step.

The changes introduced within the PR: 
 * Don't affect the way `RunEvent` events are being collected. Events are saved into database and http status code `201 (created)` is returned. 
 * `DatasetEvent` and `JobEvent` are distinguished by `schemaURL` property. If distinction cannot be made, events are treated the default way, as `RunEvent`.
 *   `DatasetEvent` and `JobEvent`events are deserialised by standard Jackso library `DatasetEvent`/`JobEvent` classes. The response returns in such cases `200 OK` http status code but **DOES NOT SAVE** data to database (this will be introduced later). The great benefit of that is that it is a good integration test and a demo of of how to do events' distinction on the server side. 

Relates to: #1868

### Solution

 * EventTypeResolver is introduced which decodes incoming `/create` requests to `LineageEvent`, `DatasetEvent` or `JobEvent` based on the `schemaURL` field.
 * Nothing changes for `RunEvents`.
 * In case of `DatasetEvent` or `JobEvent`, they get deserialised and distinguished, however no operation on database is performed. 

One-line summary:

### Checklist

### DIFF
--- a/api/src/main/java/marquez/api/OpenLineageResource.java
+++ b/api/src/main/java/marquez/api/OpenLineageResource.java
@@ -38,6 +38,7 @@ import lombok.extern.slf4j.Slf4j;
 import marquez.api.models.SortDirection;
 import marquez.db.OpenLineageDao;
 import marquez.service.ServiceFactory;
+import marquez.service.models.BaseEvent;
 import marquez.service.models.LineageEvent;
 import marquez.service.models.NodeId;
 
@@ -61,20 +62,26 @@ public class OpenLineageResource extends BaseResource {
   @Consumes(APPLICATION_JSON)
   @Produces(APPLICATION_JSON)
   @Path("/lineage")
-  public void create(
-      @Valid @NotNull LineageEvent event, @Suspended final AsyncResponse asyncResponse)
+  public void create(@Valid @NotNull BaseEvent event, @Suspended final AsyncResponse asyncResponse)
       throws JsonProcessingException, SQLException {
-    openLineageService
-        .createAsync(event)
-        .whenComplete(
-            (result, err) -> {
-              if (err != null) {
-                log.error("Unexpected error while processing request", err);
-                asyncResponse.resume(Response.status(determineStatusCode(err)).build());
-              } else {
-                asyncResponse.resume(Response.status(201).build());
-              }
-            });
+    if (event instanceof LineageEvent) {
+      openLineageService
+          .createAsync((LineageEvent) event)
+          .whenComplete(
+              (result, err) -> {
+                if (err != null) {
+                  log.error("Unexpected error while processing request", err);
+                  asyncResponse.resume(Response.status(determineStatusCode(err)).build());
+                } else {
+                  asyncResponse.resume(Response.status(201).build());
+                }
+              });
+    } else {
+      log.warn("Unsupported event type {}. Skipping without error", event.getClass().getName());
+
+      // return serialized event
+      asyncResponse.resume(Response.status(200).entity(event).build());
+    }
   }
 
   private int determineStatusCode(Throwable e) {
--- a/api/src/main/java/marquez/service/models/EventTypeResolver.java
+++ b/api/src/main/java/marquez/service/models/EventTypeResolver.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2018-2023 contributors to the Marquez project
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package marquez.service.models;
+
+import static marquez.service.models.EventTypeResolver.EventSchemaURL.LINEAGE_EVENT;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
+import com.fasterxml.jackson.databind.DatabindContext;
+import com.fasterxml.jackson.databind.JavaType;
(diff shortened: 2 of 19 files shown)


api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -61,20 +62,26 @@ public OpenLineageResource(
...
+                  asyncResponse.resume(Response.status(determineStatusCode(err)).build());
+                } else {
+                  asyncResponse.resume(Response.status(201).build());
+                }
+              });
+    } else {
+      log.warn("Unsupported event type {}. Skipping without error", event.getClass().getName());
+
+      // return serialized event
+      asyncResponse.resume(Response.status(200).entity(event).build());
```
wslulciuc, line 83: `200` status code is correct for the new OL events, and feel we should also return a `200` when accepting OL run events (as outlined by the OL [spec](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.yml#L23)). The semantics should be: "Return 200 to signify the OL event has been collected, and eventually will be processed." The `201` status code was never changed during the initial PoC phase of OL.

api/src/main/java/marquez/service/models/JobEvent.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.ToString;
+
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@Setter
+@Getter
+@Valid
+@ToString
+public class JobEvent extends BaseEvent {
```
wslulciuc, line 27: Do we have to define our own OL events, or can we use the [`io.openlineage.server`](https://openlineage.io/apidocs/javadoc/io/openlineage/server/package-summary.html) models?

wslulciuc (approved):
Left a couple questions, otherwise 👍

api/src/test/resources/open_lineage/event_dataset_event.json
```diff
@@ -0,0 +1,22 @@
+{
+  "eventTime": "2020-12-28T19:52:00.001+10:00",
```
wslulciuc, line 2: Should the dataset event have an event type?

api/src/main/java/marquez/service/models/JobEvent.java
pawel-big-lebowski: I would love to see whole Marquez `LineageEvent` being rewritten `io.openlineage.server` models at some point. I thought it does not make much sense to do it partially.

api/src/test/resources/open_lineage/event_dataset_event.json
pawel-big-lebowski: This is tricky, as the current meaning of `eventType` contains rather job statuses like `start`, `success` or `failed`. So it does not fit into this context and we decided not to use it to distinguish between `RunEvent`, `DatasetEvent` and `JobEvent`.

api/src/main/java/marquez/service/models/JobEvent.java
wslulciuc: Agreed. It's been on my radar to switch over for a while and we'll get to it at some point. Follow up PR it is 👍

api/src/main/java/marquez/api/OpenLineageResource.java
pawel-big-lebowski: The idea here was to distinguish between `RunEvent` that get saved into database (201 created) and other event types that do not affect application state. At the end, once Marquez will be capable of storing dataset and job events, it should return 201 for all the cases.

api/src/main/java/marquez/service/models/EventTypeResolver.java
```diff
@@ -0,0 +1,89 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class EventTypeResolver extends TypeIdResolverBase {
+
+  @AllArgsConstructor
+  public enum EventSchemaURL {
+    LINEAGE_EVENT(
```
julienledem, line 25: should this be called RUN_EVENT?

julienledem (approved):
Overall, this looks good to me. I made a few comments

api/src/main/java/marquez/service/models/EventTypeResolver.java
wslulciuc: I agree, let's use `RUN_EVENT`.

api/src/main/java/marquez/api/OpenLineageResource.java
wslulciuc: @julienledem we'll want to write a proposal on how to handle `DatasetEvent`s and `JobEvent`s (see https://github.com/MarquezProject/marquez/issues/2544). For now, let's ensure the event can be accepted (but not stored).

Issue #2410, opened 2023-02-08
UI error: max call stack size exceeded


wslulciuc:
<img width="507" alt="Screen Shot 2023-02-03 at 4 16 40 PM (1)" src="https://user-images.githubusercontent.com/1553185/217618061-2a4c3f55-2b9d-4c41-9799-dfd035da5477.png">

harels:
@wslulciuc do we have some more details here to help repro the issue?

jlukenoff:
I'm getting this error as well. Seems to be coming from this [line of code](https://github.com/MarquezProject/marquez/blob/f9465bf4d2a5c783b1bc88378983667b3b1465df/web/src/components/lineage/Lineage.tsx#L152) getting called recursively and blowing the stack.

![image](https://github.com/MarquezProject/marquez/assets/36255627/8239e60b-7ab7-43dc-9386-abfc7823eda9)

I'm investigating to see if this dataset has a large amount of upstream nodes or something but I don't expect that to be the case.

jlukenoff:
I took a deeper look into this and it seems like the inner helpers in `getSelectedPaths` ([ref](https://github.com/MarquezProject/marquez/blob/f9465bf4d2a5c783b1bc88378983667b3b1465df/web/src/components/lineage/Lineage.tsx#L143-L174)) aren't handling graph cycles well. It feels like this would be better to do iteratively rather than recursively or some better base case should be added to prevent stack overflow. Happy to open a PR for this

PR #2506, opened 2023-06-08
Handle lineage graph cycles on the client

jlukenoff:
### Problem

Currently we blow the stack if there is a cycle in the lineage graph.

Closes: #2410 

### Solution

Memoize the nodes that have been visited so far and only call the recursive function if the node has not yet been visited.

One-line summary:

### Checklist

### DIFF
--- a/docker/metadata.json
+++ b/docker/metadata.json
@@ -610,147 +610,6 @@
         }
       }
     },
-    "inputs": [
-      {
-        "namespace": "food_delivery",
-        "name": "public.top_delivery_times",
-        "facets": {
-          "columnLineage": {
-            "_producer": "https://github.com/MarquezProject/marquez/blob/main/docker/metadata.json",
-            "_schemaURL": "https://openlineage.io/spec/facets/1-0-1/ColumnLineageDatasetFacet.json",
-            "fields": {
-              "order_id": {
-                "inputFields": [
-                  {
-                    "namespace": "food_delivery",
-                    "name": "public.delivery_7_days",
-                    "field": "order_id"
-                  }
-                ]
-              },
-              "order_placed_on": {
-                "inputFields": [
-                  {
-                    "namespace": "food_delivery",
-                    "name": "public.delivery_7_days",
-                    "field": "order_placed_on"
-                  }
-                ]
-              },
-              "order_delivered_on": {
-                "inputFields": [
-                  {
-                    "namespace": "food_delivery",
-                    "name": "public.delivery_7_days",
-                    "field": "order_delivered_on"
-                  }
-                ]
-              },
-              "order_delivery_time": {
-                "inputFields": [
-                  {
-                    "namespace": "food_delivery",
-                    "name": "public.delivery_7_days",
-                    "field": "order_placed_on"
-                  },
-                  {
-                    "namespace": "food_delivery",
-                    "name": "public.delivery_7_days",
-                    "field": "order_delivered_on"
-                  }
-                ]
-              },
-              "customer_email": {
-                "inputFields": [
-                  {
-                    "namespace": "food_delivery",
-                    "name": "public.delivery_7_days",
-                    "field": "customer_email"
-                  }
-                ]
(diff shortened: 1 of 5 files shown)


web/src/components/lineage/Lineage.tsx
```diff
@@ -143,7 +143,12 @@ class Lineage extends React.Component<LineageProps, LineageState> {
   getSelectedPaths = () => {
     const paths = [] as Array<[string, string]>
 
+    // Sets used to detect cycles and break out of the recursive loop
```
jlukenoff, line 146: TBD: How would this work if we had a really large cycle with many nodes/paths involved? Perhaps we still run the risk of blowing the stack with this as well.

jlukenoff:
Confirmed the graph can handle cycles now but will still test a few more scenarios
image

jlukenoff:
Confirmed smaller cycles work as well and don't light up the whole graph erroneously
image

docker/metadata.json
```diff
@@ -963,6 +963,12 @@
         }
       }
     },
+    "inputs": [
+      {
+        "namespace": "food_delivery",
+        "name": "public.restaurants"
+      }
+    ],
```
jlukenoff, line 971: Not sure we actually want to merge this, but it was useful for repro'ing the issue locally and validating the fix.

phixMe: I think we should remove this before a merge.

web/src/components/lineage/Lineage.tsx
```diff
@@ -143,7 +143,16 @@ class Lineage extends React.Component<LineageProps, LineageState> {
...
+    const visitedNodes = {
+      successors: new Set(),
+      predecessors: new Set()
+    }
+
     const getSuccessors = (node: string) => {
+      if (visitedNodes.successors.has(node)) return
+      visitedNodes.successors.add(node)
+
```
phixMe, line 155: The code looks good here. Would you mind looking into a test case to isolate this problem with a little graph instance. I'd also like to improve upon the reliance on the global graph object defined in this component. I think this could take in a graph as a parameter instead of extracting it from a higher place. Let me know what you think and thanks for taking the helm on this!

jlukenoff: Are there any existing tests for `Lineage.tsx` that I could build on? I thought of adding one but wasn't familiar enough with the codebase to know how to stub things in. Agree that it's probably cleaner / more testable if we make `getSelectedPaths` more pure by accepting these things as params. Maybe it makes sense to do that for all methods of `Lineage` in a separate PR.

jlukenoff: Ah looks like I can add one pretty easily in `web/src/__tests__/components/` I'll take a look at that later.

phixMe: Thanks, and yes it should be simple enough to initialize a dummy graph object there and test out this cycle issue.

web/src/__tests__/components/Lineage.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+        {...mockProps}
+        // TODO: fix type errors, probably needs a stub interface?
+        lineage={mockGraphWithCycle}
+        selectedNode={mockGraphWithCycle[0].id}
+      />
+    )
+
+    const instance = wrapper.instance()
+    // TODO: fix type errors, the above stub interface will probably fix this
+    const paths = instance.getSelectedPaths()
```
jlukenoff, line 93: @phixMe would you mind giving some feedback on this testing approach? You can probably see where I'm headed here, but I'm getting some type errors and have a feeling there's a better way to mock this than to stub every prop this component expects.

jlukenoff:
Updated the tests and validated that they fail if I remove my recursive base case. @phixMe lmk if you have any other tests you think we should add to this.

phixMe (approved):
Thanks @jlukenoff Great work!

Issue #2378, opened 2023-01-23
Allow users to set depth of lineage graph in UI


wslulciuc:
In the Marquez UI, the user is not provided with the option to set the depth of the lineage graph even though the [LineageAPI](https://marquezproject.github.io/marquez/openapi.html#operation/getLineage) defines a `depth` param. Therefore, the [lineage call](https://github.com/MarquezProject/marquez/blob/main/web/src/store/requests/lineage.ts#L10) defaults to `20` on the [backend](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/api/OpenLineageResource.java#L47). 

**Benefits**:

1. Reduce lineage graph render complexity i.e. a dynamic `depth` of `N` would return a more digestible lineage UI view for debugging
2. Reduce lineage graph call response time i.e.  a dynamic `depth` of `N` would avoid returning a lineage graph with all connected nodes from `origin` node

davidsharp7:
Related to [2218](https://github.com/MarquezProject/marquez/issues/2218)

PR #2525, opened 2023-06-20
Lineage UI: Add a graph depth config option

jlukenoff:
### Problem
Currently for very large lineage graphs, it can be hard to navigate the graph by dragging and zooming.

Closes: #2378 

### Solution

This PR adds an option to the lineage graph UI where the user can select the depth of the lineage graph.

### Checklist


jlukenoff:
I just added this as floating in the upper right-hand corner for now. Happy to move it wherever it makes the most sense, although this feels pretty usable to me

output

web/src/store/reducers/lineage.ts
```diff
@@ -5,26 +5,31 @@ import {
...
   bottomBarHeight: number
+  depth: number
 }
 
 const initialState: ILineageState = {
   lineage: { graph: [] },
   selectedNode: null,
-  bottomBarHeight: (window.innerHeight - HEADER_HEIGHT) / 3
+  bottomBarHeight: (window.innerHeight - HEADER_HEIGHT) / 3,
+  depth: 1
```
phixMe, line 27: Can we make this default to a larger number that would render the whole demo graph at least?

jlukenoff: Good point, updated it to `5`. Trying to bias towards a lower value as, at least in my experience playing around with the demo, it feels less clunky to slowly increase from a lower number to "zoom out" vs. walking down from a higher number to "zoom in".

phixMe (approved):
Thanks @jlukenoff 🐑 it!

jlukenoff:
Thanks @phixMe ! Just added an entry to the changelog. Would you mind merging? I don't have write access yet

wslulciuc (approved):
Love it! Let's favor usability for now and 🚢 it as is!

Issue #2550, opened 2023-07-24
Add `namespace:` and `updated_at:` query qualifier to `SearchAPI`


wslulciuc:
Our [`searchAPI`](https://marquezproject.github.io/marquez/openapi.html#operation/search) allows callers to query for jobs and datasets **across** namespaces. But, callers might need to narrow their search results by namespace. To enable callers to query for jobs and datasets **within** a namespace, let's add the `namespace:` qualifier:

`namespace:NAMESPACE`: will match jobs or datasets within namespace `NAMESPACE`.

For example, query (with `filter=jobs`):

```
my-best-job namespace:example
```

will match job(s) within namespace `example` with that name.

To filter by time of last updated, we're also adding the `updated_at:` qualifier:

`updated_at:<YYYY-MM-DD`: will match jobs or datasets before `YYYY-MM-DD`.
`updated_at:>YYYY-MM-DD`: will match jobs or datasets after  `YYYY-MM-DD`.

Below, we add the `updated_at:` qualifier to our example from earlier:

```
my-best-job namespace:example updated_at:<2023-07-24
```

now, the query will match job(s) within namespace `example` with that name, that were last updated before `2023-07-24` .

PR #2555, opened 2023-07-28
Add `namespace`, `before` and `after` to SearchDao

tatiana:
### Problem

As of 0.37.0, Marquez Search API does not support filtering search results by namespace or date.

Relates to #2550 

### Solution

This PR is a follow-up to #2553, consisting of the Dao change necessary to support:
* `namespace` - match jobs or datasets within the given namespace.
* `before` - match jobs or datasets before YYYY-MM-DD.
* `after` - match jobs or datasets after YYYY-MM-DD.

The API change itself is part of a separate PR to come.

### Checklist

### DIFF
--- a/api/src/test/java/marquez/db/SearchDaoTest.java
+++ b/api/src/test/java/marquez/db/SearchDaoTest.java
@@ -11,6 +11,7 @@ import com.google.common.collect.ImmutableSet;
 import java.net.URL;
 import java.sql.SQLException;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


api/src/test/java/marquez/db/SearchDaoTest.java
```diff
@@ -106,6 +112,43 @@ public static void setUpOnce(final Jdbi jdbi) throws SQLException {
             });
   }
 
+  @Test
+  public void testSearch_filterByNamespace() {
```
wslulciuc, line 116: Solid test! Can we also add a test for the usage of `after`?

wslulciuc (approved):
Great work, @tatiana! 💯

Issue #2558, opened 2023-07-31
Missing critical lineage path view on Marquez UI


hloombaupgrade:
Currently entire lineage graph is appeared on Marquez UI and critical path is highlighted in green. But on small screens its very unclear to see such complex graph, so proposing to create a view which renders the lineage of selected job or dataset

davidsharp7:
Big + 1 on this one.  It would be a great feature add.

Issue #2014, opened 2022-06-08
display closest nodes not all the nodes in the tree


dvirgiln:
When the datasets in the namespace are growing it is impossible to use marquez ui to follow dependencies. 

I ve attached an image of a subset of the graph
<img width="1578" alt="Screenshot 2022-06-08 at 08 26 48" src="https://user-images.githubusercontent.com/6932080/172557808-c7bcb733-1261-4ce9-a1ea-01e57e9b959a.png">

You can see that is impossible to follow.  This is in my company what everyone is complaining. If they want to see the lineage of one dataset, they see all the tree with things that they are not really interested. Ideally I would display a button or filter where you can decide how many levels you want to see, and by default you just see one level downstream nodes and one level upstream nodes. 

Hope it is clear

wslulciuc:
@dvirgiln: You're absolutely right, for very large graphs it's hard to navigate / follow dependencies (not a great experience). Usability is something we've been meaning to address, and there's plenty of room for improvement here. I can start opening issues that outline some of the changes we'd like to make in the UI (with mockups), but I also encourage you to do the same and open UI issues and labeliing them with [`web`](https://github.com/MarquezProject/marquez/labels/web) and [`feature`](https://github.com/MarquezProject/marquez/labels/feature). I'll make sure to add the issues to our [roadmap](https://github.com/orgs/MarquezProject/projects/1/views/4?visibleFields=%5B%22Title%22%2C%22Assignees%22%2C%22Status%22%2C2807312%2C%22Labels%22%5D)

To give you a sneak peek of what changes we'd like to make the lineage graph navigation, you can checkout [`demo.datakin.com`](https://demo.datakin.com/lineage/job/food_delivery/etl_orders.etl_categories)(Datakin is powered by Marquez ❤️ ):

<img width="1005" alt="Screen Shot 2022-06-10 at 4 51 52 PM" src="https://user-images.githubusercontent.com/1553185/173163631-9a5e5cf1-0ac5-487f-b1b4-f53740e41275.png">

The UI above uses the Marquez LineageAPI, but applies client-side grouping logic. The team is currently discussing the effort to begin moving the dataset and job grouping logic to Marquez. What are your initial thoughts? _Also, I think deciding on how many levels deep in the graph should be rendered is already supported by the LineageAPI, it's just a matter of providing that option to the user 😄

PR #2569, opened 2023-08-05
UI: add full graph toggle

jlukenoff:
### Problem

There are some reasonable scenarios where users would want to view only the direct successors and predecessors of the current selected node in the lineage graph. This PR enables users to toggle a "Full Graph" view, when deselected, the graph will render only the nodes that are along the selected paths. When deselected and changing selected nodes, we expect that the graph should re-render with any new nodes that weren't along the selected path previously.

Closes: #2558

### Solution

- Add a `showFullGraph` property to the lineage state and enables a toggle on the UI. 
- Add a `removeUnselectedNodes` method to `Lineage` that can remove any unselected nodes from the graph.

Screen Recording 2023-08-05 at 9 13 18 AM

One-line summary: add a toggle to the Lineage UI to let users switch between viewing the full graph and a view with only the selected paths.

### Checklist


davidsharp7:
Yeah looks good - only suggestion I would make is to change the background colour of the toggle to Marquez Green when its on so its a little easier to determine state.

jlukenoff:
> Yeah looks good - only suggestion I would make is to change the background colour of the toggle to Marquez Green when its on so its a little easier to determine state.

Good idea, updated it to use the primary green color when enabled.

image

davidsharp7:
> > Yeah looks good - only suggestion I would make is to change the background colour of the toggle to Marquez Green when its on so its a little easier to determine state.
> 
> Good idea, updated it to use the primary green color when enabled.
> 
> image

Yeah perfect!  Might need to get wider feedback but I wonder if its worth inverting the toggle so that by default its shows everything and activating the toggle shows only the critical path.

jlukenoff:
That’s a good point, that may make more sense since the full graph has always been the default and this is something new users can enable. I’m having trouble thinking of a label for the switch that would make sense that isn’t too wordy though. Perhaps “Simple Graph” or something like that? 

Agree it would be great to get some wider feedback.

davidsharp7:
> That’s a good point, that may make more sense since the full graph has always been the default and this is something new users can enable. I’m having trouble thinking of a label for the switch that would make sense that isn’t too wordy though. Perhaps “Simple Graph” or something like that?
> 
> Agree it would be great to get some wider feedback.

What about “Show Critical Path”?

davidsharp7:
Yeah I think we can just reverse the logic and it should work fine

I've added the sections that can be reversed on your branch (could be wrong as this the first time I have ever looked at typescript/react/node)

web/src/store/reducers/lineage.ts
```diff
@@ -18,13 +19,15 @@ export interface ILineageState {
...
+  showFullGraph: boolean
 }
 
 const initialState: ILineageState = {
   lineage: { graph: [] },
   selectedNode: null,
   bottomBarHeight: (window.innerHeight - HEADER_HEIGHT) / 3,
-  depth: 5
+  depth: 5,
+  showFullGraph: true
```
davidsharp7, line 30:
```
showFullGraph: false
```

web/src/i18n/config.ts
```diff
@@ -57,7 +57,8 @@ i18next
           lineage: {
             empty_title: 'No node selected',
             empty_body: 'Try selecting a node through search or the jobs or datasets page.',
-            graph_depth_title: 'Graph Depth'
+            graph_depth_title: 'Graph Depth',
+            full_graph_label: 'Full Graph'
```
davidsharp7, line 61:
```
full_graph_label: 'Show Critical Path'
```

web/src/components/lineage/Lineage.tsx
```diff
@@ -130,6 +140,12 @@ export class Lineage extends React.Component<LineageProps, LineageState> {
       )
       this.getEdges()
     }
+
+    // Always rebuild the graph if the selected node changes and we aren't showing the full graph
+    // Since there may be more or less nodes to change
+    if (this.props.selectedNode !== prevProps.selectedNode && !this.props.showFullGraph) {
```
davidsharp7, line 146:
```
    if (this.props.selectedNode !== prevProps.selectedNode && this.props.showFullGraph) {
      this.buildGraphAll(this.props.lineage.graph)
    }
```

web/src/components/lineage/Lineage.tsx
```diff
@@ -215,6 +240,11 @@ export class Lineage extends React.Component<LineageProps, LineageState> {
         g.setEdge(graph[i].inEdges[j].origin, graph[i].id)
       }
     }
+
+    if (!this.props.showFullGraph) {
```
davidsharp7, line 244:
```
    if (this.props.showFullGraph) {
      this.removeUnselectedNodes()
    }
```

davidsharp7:
Promise this is my last comment - there is a slight bug when you select "etl_delivery_7_days" - notice that there are 2 connected nodes even though they are not on the critical path.  I think its because technically you can't filter out those nodes because they are part of the critical path at some point.  I think a work around is to add something like this

```
    this.setState({
      graph: g,
      edges: this.props.showFullGraph ? this.getEdges().filter(edge => edge.isSelected === true) : this.getEdges(),
      nodes: g.nodes().map(v => g.node(v))
    })
```
Screenshot 2023-08-06 at 9 52 20 pm

jlukenoff:
> Promise this is my last comment - there is a slight bug when you select "etl_delivery_7_days" - notice that there are 2 connected nodes even though they are not on the critical path. I think its because technically you can't filter out those nodes because they are part of the critical path at some point. I think a work around is to add something like this

I'm not sure I understand, doesn't the fact that they are connected by the selected path mean they are on the critical path?

web/src/i18n/config.ts
jlukenoff: Eh, this just starts to feel like a lot of words to express something relatively simple. I tend to err on the side of "less is more" with these things. I really like the [ArgoCD](https://argo-cd.readthedocs.io/) version of this where the text labels show up as a tooltip when you hover over this options. I think that helps preserve the real-estate the configs take up. Maybe we want to think about something like that if we plan to add more configs to this part of the screen.

image

jlukenoff:
Had to do a pretty big refactor to account for #2563 but the functionality it still the same. Plus it's worth it to inherit the react upgrade 😄

davidsharp7:
> > Promise this is my last comment - there is a slight bug when you select "etl_delivery_7_days" - notice that there are 2 connected nodes even though they are not on the critical path. I think its because technically you can't filter out those nodes because they are part of the critical path at some point. I think a work around is to add something like this
> 
> I'm not sure I understand, doesn't the fact that they are connected by the selected path mean they are on the critical path? Could you point out which nodes in your screenshot you think shouldn't be displayed?

So if you select etl_delivery_7_days notice how there is a grey line between public.customers and email_discounts. Given the job selected that is not on the critical path and shouldn’t appear. Simplest solution is to also filter out anything where isSelected is False.

web/src/i18n/config.ts
davidsharp7: Yeah I’m in two minds about that. Like the idea of having a pop up on hover to display basic ( maybe even facets like data quality?) info about the node. Maybe we need an additional menu on the header for global type variables like display and depth?

wslulciuc: As a follow up, the lineage API should support the different graph options (`Full`, `Critical`) to avoid filtering on the UI. But, in the UI is a great first step!

wslulciuc:
> So if you select etl_delivery_7_days notice how there is a grey line between public.customers and email_discounts. Given the job selected that is not on the critical path and shouldn’t appear. Simplest solution is to also filter out anything where isSelected is False.

@davidsharp7 I wanted to add to @jlukenoff response. In the `Full` graph mode, we'll want to show all dataset and job dependencies, but the connections between nodes in the graph don't need to be expanded by default. That is, they can be grouped and collapsed, allowing the user to discover and explore the graph as needed. For some insight into where I see the interaction of the graph going, see https://github.com/MarquezProject/marquez/issues/2014

web/src/__tests__/components/Lineage.test.tsx
```diff
@@ -85,7 +90,7 @@ describe('Lineage Component', () => {
 
   beforeEach(() => {
     g = initGraph()
-    buildGraphAll(g, mockGraphWithCycle, (gResult: graphlib.Graph<MqNode>) => {
+    buildGraphAll(g, mockGraphWithCycle, false, selectedNode, (gResult: graphlib.Graph<MqNode>) => {
```
phixMe, line 93: Can we add a test case here for the node count on the `fullGraph` option?

davidsharp7:
> > So if you select etl_delivery_7_days notice how there is a grey line between public.customers and email_discounts. Given the job selected that is not on the critical path and shouldn’t appear. Simplest solution is to also filter out anything where isSelected is False.
> 
> @davidsharp7 I wanted to add to @jlukenoff response. In the `Full` graph mode, we'll want to show all dataset and job dependencies, but the connections between nodes in the graph don't need to be expanded by default. That is, they can be grouped and collapsed, allowing the user to discover and explore the graph as needed. For some insight into where I see the interaction of the graph going, see #2014

Yeah I don't think its even a binary choice as each of those views is legitimate i.e

1) I am changing dataset X I just need to see the critical path to understand impacts.
2) I am changing dataset X and I want to explore my lineage - expand/collapse
3) I want to see my entire data eco system.

wslulciuc:
@davidsharp7 I think you make good points, and agree a more broad discussion on the graph view can be had, but I would move the discussion to an issue. Mind opening an issue to elaborate on the graph view in the UI and ways to improve the user expereince?

As for this PR, being able to toggle between a "Full" and "Critical Path" view is the main feature added, so I would merge as is with pending feedback from @phixMe. It's a start and we can follow up with improvements!

wslulciuc (approved):
I left some minor feedback on labeling the graph options, but otherwise 👍

phixMe (approved):
Thanks for adding the unit test. Looks great!

web/src/components/lineage/Lineage.tsx
```diff
@@ -126,10 +134,10 @@ export function getSelectedPaths(g: graphlib.Graph<MqNode>, selectedNode: string
 
     const predecessors = g?.predecessors(node)
     if (predecessors?.length) {
-      for (let i = 0; i < node.length - 1; i++) {
+      for (let i = 0; i < predecessors.length; i++) {
         if (predecessors[i]) {
-          paths.push([predecessors[i] as unknown as string, node])
-          getPredecessors(predecessors[i] as unknown as string)
+          paths.push([(predecessors[i] as unknown) as string, node])
+          getPredecessors((predecessors[i] as unknown) as string)
```
jlukenoff, line 140: Not sure how long we've been doing this but it looks like we've been iterating over the characters in the node id 😬 ? Update this to traverse the predecessors array and did the same for getSuccessors

jlukenoff:
Just made a few styling tweaks that I found after adjusting to shorter labels for these configs. This should be good to merge once tests pass (I don't have access to merge yet)

Issue #2580, opened 2023-08-11
Add html tool tip for displaying node information


davidsharp7:
**Description**

Currently the Marquez UI compresses dataset/job name under each node if they are over x characters which leads to a sub optimal user experience when browsing nodes on a graph.  The current tool tip just displays if the object is a job/dataset and so is not massively helpful.

 **Proposed Solution**

Add an HTML tool tip which on hover over displays details about the node.  Initially I would like to just update

[https://github.com/MarquezProject/marquez/blob/main/web/src/components/lineage/components/node/Node.tsx](url)

and reformat the node label to display 

node type (dataset/job)
Namespace
Object Name.

<img width="549" alt="Screenshot 2023-08-12 at 7 50 02 am" src="https://github.com/MarquezProject/marquez/assets/34074888/0074aa20-e8d5-43fa-b77a-66e026bc1fba">

This feature could be expanded to include multiple facets but that is a discussion for the wider community.

PR #2587, opened 2023-08-12
adding HTML node tooltip

davidsharp7:
### Problem

Currently the Marquez UI compresses dataset/job name under each node if they are over x characters which leads to a sub optimal user experience when browsing nodes on a graph. The current tool tip just displays if the object is a job/dataset and so is not massively helpful.

Closes: #2580 

### Solution

One-line summary:

add HTML tooltip which activates on hovering over a node

### Checklist

### DIFF
--- a/web/src/components/lineage/components/node/Node.tsx
+++ b/web/src/components/lineage/components/node/Node.tsx
@@ -53,31 +53,34 @@ const Node: React.FC<NodeProps> = ({ node, selectedNode, setSelectedNode }) => {
     [`& .${tooltipClasses.tooltip}`]: {
       backgroundColor: theme.palette.background.default,
       color: theme.palette.common.white,
-      maxWidth: 400,
+      maxWidth: 600,
       fontSize: theme.typography.pxToRem(14),
       border: '1px solid ' + theme.palette.common.white,
     },
   }))
 
   // return the object namespace/name
-  const addToToolTip = (inputString: string) => {
+  const addToToolTip = (inputData: any) => {
+    const inputString = String(inputData.label)
+    const desc = String(inputData.data.description)
+    const namespace = String(inputData.data.namespace)
+    const objectName = String(inputData.data.name)
     return <React.Fragment>
       <Typography color="inherit">{inputString.split(':')[0]}</Typography>
-      <b>{"Namespace: "}</b>{inputString.split(':')[1]}<br></br>
-      <b>{"Object Name: "}</b>{inputString.split(':')[2]}<br></br>
+      <b>{"Namespace: "}</b>{namespace}<br></br>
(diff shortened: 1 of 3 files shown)


web/src/components/lineage/components/node/Node.tsx
```diff
@@ -43,25 +46,52 @@ const Node: React.FC<NodeProps> = ({ node, selectedNode, setSelectedNode }) => {
...
+  const HtmlTooltip = styled(({ className, ...props }: TooltipProps) => (
+    <Tooltip {...props} classes={{ popper: className }} />
+  ))(({ theme }) => ({
+    [`& .${tooltipClasses.tooltip}`]: {
+      backgroundColor: theme.palette.background.default,
+      color: theme.palette.common.white,
+      maxWidth: 600,
+      fontSize: theme.typography.pxToRem(14),
+      border: '1px solid ' + theme.palette.common.white,
+    },
```
phixMe, line 59: It would be our preference moving forward to avoid this type of css override and instead use the default properties as much as possible since there are now multiple maintainers it's becoming a little challenging to find consensus here. I would try to only use the properties available to you here without stylesheet overrides.

web/src/components/lineage/components/node/Node.tsx
```diff
@@ -43,25 +46,52 @@ const Node: React.FC<NodeProps> = ({ node, selectedNode, setSelectedNode }) => {
     return '/'
   }
 
+  // configure the tool tip
+  const HtmlTooltip = styled(({ className, ...props }: TooltipProps) => (
```
phixMe, line 50: I think we want this to be in a different file by itself called `MQTooltip.tsx` to allow for component reuse.

web/src/components/lineage/components/node/Node.tsx
```diff
@@ -43,25 +46,52 @@ const Node: React.FC<NodeProps> = ({ node, selectedNode, setSelectedNode }) => {
...
+  const addToToolTip = (inputData: any) => {
+    const inputString = String(inputData.label)
+    const desc = String(inputData.data.description)
+    const namespace = String(inputData.data.namespace)
+    const objectName = String(inputData.data.name)
+    return <React.Fragment>
+      <Typography color="inherit">{inputString.split(':')[0]}</Typography>
+      <b>{"Namespace: "}</b>{namespace}<br></br>
+      <b>{"Name: "}</b>{objectName}<br></br>
+      <b>{"Description: "}</b>{desc === 'null' ? "No Description" : desc}
```
phixMe, line 72: I think this should use the `children` prop in order for the component to be generally configurable.

phixMe (commented):
Thanks for the PR, some comments for you!

Issue #2605, opened 2023-08-30
web: Column Level Tags are not displayed in the UI


davidsharp7:
Currently any dataset columns with tags are not displayed in the dataset information pain.  The data is available in the field.tags var.

Solution - add a new column called "tags" to the dataset column view and add the tags associated with the dataset column.

PR #2606, opened 2023-08-30
add available column level tags

davidsharp7:
### Problem

Currently any dataset columns with tags are not displayed in the dataset information pane. The data is available in the field.tags var.

Closes: #2605

### Solution

add a new column called "tags" to the dataset column view and add the tags associated with the dataset column.

Screenshot 2023-09-06 at 05 53 39

One-line summary:

add a new column called "tags" to the dataset column view and add the tags associated with the dataset column.

### Checklist

### DIFF
--- a/web/src/components/datasets/DatasetInfo.tsx
+++ b/web/src/components/datasets/DatasetInfo.tsx
@@ -51,13 +51,7 @@ const formatColumnTags = (tags: string[]) => {
           key={tag}
           label={tag}
           size="small"
-          style={
-            index < tags.length - 1
-              ? {
-                  marginRight: theme.spacing(1),
-                }
-              : {}
-          }
+          style={{display: 'inline', marginRight: index < tags.length - 1 ? theme.spacing(1) : 0}}
         />
       ))}
     </>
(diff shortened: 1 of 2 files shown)


wslulciuc (approved):
LGTM 👍 💯 🥇

web/src/components/datasets/DatasetInfo.tsx
```diff
@@ -39,6 +42,23 @@ type DatasetInfoProps = {
 } & JobFacetsProps &
   DispatchProps
 
+const formatColumnTags = (tags: string[]) => {
+  const theme = createTheme(useTheme())
+  return <>{tags.map((tag, index) => (
+    <span
```
phixMe, line 48: I don't think we need the span here. I think we can just use the chip itself with marginRight applied to that component itself.

For applying `marginRIght` I would say something like:
marginRight: `index < tags.length - 1 ? theme.spacing(1) : 0`

davidsharp7: Cool - away racing this weekend but will fix it up Monday (NZ time)

Issue #2624, opened 2023-09-26
Static Lineage Support in Marquez using OpenLineage


wslulciuc:
### Background	

OpenLineage [`0.29.2`](https://github.com/OpenLineage/OpenLineage/releases/tag/0.29.2) added support for static lineage events: [`DatasetEvent`](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L65C6-L65C18), [`JobEvent`](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L161). Before `0.29.2`, lineage events required a unique `runID` and emitted only at the job run-level. With static lineage events, the spec has evolved to allow for run-less lineage events where events can be emitted outside the context of a job run with the unique `runID` now optional.

### What is Static Lineage?

In Marquez, _static lineage_ would  represent the latest (or current) lineage metadata for a given job. That is, each job node in the lineage graph will reference the _current_ metadata for a given job, but also the _current_ metadata for its input and output dataset(s). Below, we outline the`Job` graph node definition for static lineage:

> **Note:** We use latest (or current) metadata and **latest version**  interchangeably for a job or dataset.

`Job`: `Object`
* `ID` - The globally unique ID of the job (_namespace + job name_)
* `Inputs`:`Set[InputDataset]` - A set of input datasets (_latest versions(s)_) (=`inEdges`)
* `Outputs`: `Set[OutputDataset]` - A set of output datasets  (_latest versions(s)_) (=`outEdges`)

Now, if we look at our [LineageAPI](https://marquezproject.github.io/marquez/openapi.html#operation/getLineage), you'll notice that the contract does not change! That is, the query used for the lineage graph API call always depended on the latest lineage metadata for a given job. But, given that the `runID` was required by OpenLineage, we needed to query the `dataset_versions`, `job_version`, and `runs` tables to return the current lineage graph. 

### How Marquez uses _Static Lineage_ events emitted by OpenLineage?

With `runID` not required for static lineage events, Marquez will handle `DatasetEvent` and `JobEvent` differently. Below, we outline the processing logic for each static lineage event.

#### `DatasetEvent`

Within a `RunEvent`, datasets were either inputs or outputs to a job run. A lookup for the current version of the input datasets (assumed to be present) would ensure Marquez associated the  input versions to the `runID`; output datasets are handlled slightly differently, but more or less the same in terms of processing logic. In the case of output datasets, Marquez applies the following versioning logic:

[...]

PR #2641, opened 2023-10-06
Runless events - consume dataset event 

pawel-big-lebowski:
### Problem

Part of the #2624 to support `DatasetEvent`. 

Closes: #2624 

### Solution

One-line summary:

### Checklist


api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -67,23 +69,28 @@ public void create(@Valid @NotNull BaseEvent event, @Suspended final AsyncRespon
...
-                  asyncResponse.resume(Response.status(201).build());
-                }
-              });
-    } else {
+          .whenComplete((result, err) -> onComplete(result, err, asyncResponse));
+    } else if (event instanceof DatasetEvent) {
+      openLineageService
+          .createAsync((DatasetEvent) event)
+          .whenComplete((result, err) -> onComplete(result, err, asyncResponse));
+    } else if (event instanceof JobEvent) {
```
julienledem, line 77: it looks like this should just be `else` for now.

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -135,6 +145,80 @@ default UpdateLineageRow updateMarquezModel(LineageEvent event, ObjectMapper map
...
+            facets ->
+                datasetFacetsDao.insertDatasetFacetsFor(
+                    record.getDatasetRow().getUuid(),
+                    record.getDatasetVersionRow().getUuid(),
+                    null,
+                    now,
+                    null,
+                    facets));
+
+    // OutputFacets ...
```
julienledem: Do we not have InputFacets as well in this case?

julienledem (commented):
I left some comments/questions but otherwise, this looks good. 
I like the multiple commits. That made it so much easier to review.

api/src/main/java/marquez/db/OpenLineageDao.java
pawel-big-lebowski: Neither `inputFacet` nor `outputFacets` should be  allowed for `DatasetEvent`.
We should have 3 methods: `insertDatasetFacets`, `insertInputDatasetFacets` and `insertOutputDatasetFacets`. 
I fixed this in the 5th commit.

julienledem:
Thanks for the update! 
@wslulciuc : did you have anymore comments?

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -135,223 +149,226 @@ default UpdateLineageRow updateMarquezModel(LineageEvent event, ObjectMapper map
...
-    DatasetVersionDao datasetVersionDao = createDatasetVersionDao();
-    DatasetFieldDao datasetFieldDao = createDatasetFieldDao();
-    DatasetFacetsDao datasetFacetsDao = createDatasetFacetsDao();
-    RunDao runDao = createRunDao();
-    RunArgsDao runArgsDao = createRunArgsDao();
-    RunStateDao runStateDao = createRunStateDao();
-    ColumnLineageDao columnLineageDao = createColumnLineageDao();
-    RunFacetsDao runFacetsDao = createRunFacetsDao();
+  default UpdateLineageRow updateMarquezModel(DatasetEvent event, ObjectMapper mapper) {
+    daos.initBaseDao(this);
```
wslulciuc, line 153: This is pretty cool! I agree a wrapper class like `ModelDaos` helps with readability and minimizes `DAO` boilerplate initialization code. We can probably take the refactoring further an ask ourselves if we really need all of this logic in the DAO layer? Anyways, great to see cleanup efforts in this part of the codebase! 💯

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -93,6 +99,14 @@ void createLineageEvent(
       PGobject event,
       String producer);
 
+  @SqlUpdate(
+      "INSERT INTO lineage_events ("
+          + "event_time, "
```
wslulciuc, line 104: Can we default the event type to `DATASET` for `DatasetEvent`s? And/or add a `schema_url` column? When `JobEvent`s support is added, it will be difficult to  distinguish between dataset and job events.

api/src/main/resources/marquez/db/migration/V65__dataset_facets_lineage_event_type_nullable.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE dataset_facets ALTER COLUMN lineage_event_type DROP NOT NULL;
```
wslulciuc, line 1: I'm not entirely convinced we should drop the `NULL` constraint on the `lineage_event_type` column. Would we also want to drop it on `lineage_event_time`? I think it would be odd to have `lineage_event_type` as `NULL`, but `lineage_event_time` with a value (though I completely understand the reasoning behind this). Just feels inconsistent. Thoughts?

api/src/main/java/marquez/db/OpenLineageDao.java
wslulciuc: Ahh yes thanks for clarifying; was thinking about just this during my review 👀

api/src/main/java/marquez/db/OpenLineageDao.java
pawel-big-lebowski: Yes, you're right and I noticed this when working on the `JobEvent` and fixed this in:
https://github.com/MarquezProject/marquez/pull/2661/commits/84f46702a05f62ae1c3399ad51d876113006f6e4

It adds `ol_spec_type` column in `lineage_events` (didn't want to use `event_type` which we already use in different context)

api/src/main/resources/marquez/db/migration/V65__dataset_facets_lineage_event_type_nullable.sql
pawel-big-lebowski: `event_type` does not describe what it contains and is misleading. But this is how it already is. In fact, the column contains `run_state`. I think it's OK to have nullable `run_state`. 

This is similar problem we had in OL spec 2.0. Although we realised this field does not fill well, we decided it's too much effort to fix it. 

Perhaps we should go with separate PR that renames all `eventType`,`event_type` to  `runState`,`run_state`.

wslulciuc (approved):
💯 💯 💯

PR #2650, opened 2023-10-17
Amend indentation in OpenAPI spec

davidjgoss:
### Problem

I'm referencing the Marquez OpenAPI spec in a Backstage component, via a URL like <https://raw.githubusercontent.com/MarquezProject/marquez/0.41.0/spec/openapi.yml>. Backstage uses a very strict (fussy?) YAML parser which has taken exception to the way a particular JSON block is formed:

```
YAMLParseError: Flow map in block collection must be sufficiently indented and end with a } at line 1910, column 9
```

### Solution

This PR amends the indentation so the aforementioned parser will accept and render the spec.

An alternative would be to convert this chunk of the spec to full YAML, rather than inline JSON - happy to do that instead if it's preferred.

One-line summary:  
Amend indentation in OpenAPI spec

### Checklist

### DIFF
--- a/spec/openapi.yml
+++ b/spec/openapi.yml
@@ -1919,22 +1919,24 @@ components:
         eventTime: 2019-05-09T19:49:24.201361Z
         run: {runId: d46e465b-d358-4d32-83d4-df660ff614dd}
         job: {namespace: my-namespace,name: my-job}
-        outputs: [{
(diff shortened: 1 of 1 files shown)


wslulciuc (approved):
Thanks for the fix, @davidjgoss!

PR #2654, opened 2023-10-17
Runless events - refactor job_versions_io_mapping

pawel-big-lebowski:
### Problem

This is currently a draft PR which is far from being merged. It is missing few tests related to schema changes which are marked with `todo` within the code. I've created a PR to have a better discussion on adding `job_id` to `job_versions_io_mapping`. This PR should be a follow-up of https://github.com/MarquezProject/marquez/pull/2641. 

The assumption was that it should be helpful in optimising get-lineage query. I would like first to clarify how are we going to make benefit of this extra column. 

### Solution

One-line summary:

### Checklist

### DIFF
--- a/api/src/main/java/marquez/db/JobVersionDao.java
+++ b/api/src/main/java/marquez/db/JobVersionDao.java
@@ -202,8 +202,8 @@ public interface JobVersionDao extends BaseDao {
   @SqlUpdate(
       """
     INSERT INTO job_versions_io_mapping (
-      job_version_uuid, dataset_uuid, io_type, job_uuid, job_symlink_target_uuid, is_current_job_version)
-    VALUES (:jobVersionUuid, :datasetUuid, :ioType, :jobUuid, :symlinkTargetJobUuid, TRUE)
+      job_version_uuid, dataset_uuid, io_type, job_uuid, job_symlink_target_uuid, is_current_job_version, made_current_at)
+    VALUES (:jobVersionUuid, :datasetUuid, :ioType, :jobUuid, :symlinkTargetJobUuid, TRUE, NOW())
     ON CONFLICT (job_version_uuid, dataset_uuid, io_type, job_uuid) DO NOTHING
   """)
   void upsertCurrentInputOrOutputDatasetFor(
--- a/api/src/main/java/marquez/db/JobVersionDao.java
+++ b/api/src/main/java/marquez/db/JobVersionDao.java
@@ -192,40 +192,73 @@ public interface JobVersionDao extends BaseDao {
       String namespaceName);
 
   /**
-   * Used to link an input dataset to a given job version.
-   *
-   * @param jobVersionUuid The unique ID of the job version.
-   * @param inputDatasetUuid The unique ID of the input dataset.
-   */
-  default void upsertInputDatasetFor(UUID jobVersionUuid, UUID inputDatasetUuid) {
-    upsertInputOrOutputDatasetFor(jobVersionUuid, inputDatasetUuid, IoType.INPUT);
-  }
-
-  /**
-   * Used to link an output dataset to a given job version.
-   *
-   * @param jobVersionUuid The unique ID of the job version.
-   * @param outputDatasetUuid The unique ID of the output dataset.
-   */
-  default void upsertOutputDatasetFor(UUID jobVersionUuid, UUID outputDatasetUuid) {
-    upsertInputOrOutputDatasetFor(jobVersionUuid, outputDatasetUuid, IoType.OUTPUT);
-  }
-
-  /**
    * Used to upsert an input or output dataset to a given job version.
    *
    * @param jobVersionUuid The unique ID of the job version.
    * @param datasetUuid The unique ID of the output dataset
    * @param ioType The {@link IoType} of the dataset.
+   * @param jobUuid The unique ID of the job.
    */
   @SqlUpdate(
       """
     INSERT INTO job_versions_io_mapping (
-      job_version_uuid, dataset_uuid, io_type)
-    VALUES (:jobVersionUuid, :datasetUuid, :ioType)
-    ON CONFLICT DO NOTHING
+      job_version_uuid, dataset_uuid, io_type, job_uuid, job_symlink_target_uuid, is_current_job_version)
+    VALUES (:jobVersionUuid, :datasetUuid, :ioType, :jobUuid, :symlinkTargetJobUuid, TRUE)
+    ON CONFLICT (job_version_uuid, dataset_uuid, io_type, job_uuid) DO NOTHING
   """)
-  void upsertInputOrOutputDatasetFor(UUID jobVersionUuid, UUID datasetUuid, IoType ioType);
+  void upsertCurrentInputOrOutputDatasetFor(
+      UUID jobVersionUuid,
+      UUID datasetUuid,
+      UUID jobUuid,
+      UUID symlinkTargetJobUuid,
+      IoType ioType);
+
+  @SqlUpdate(
+      """
+    UPDATE job_versions_io_mapping
+    SET is_current_job_version = FALSE
+    WHERE (job_uuid = :jobUuid OR job_symlink_target_uuid = :jobUuid)
+    AND job_version_uuid != :jobVersionUuid
+    AND io_type = :ioType
+    AND is_current_job_version = TRUE;
+  """)
+  void markInputOrOutputDatasetAsPreviousFor(UUID jobVersionUuid, UUID jobUuid, IoType ioType);
+
+  @SqlUpdate(
+      """
+    UPDATE job_versions_io_mapping
+    SET is_current_job_version = FALSE
+    WHERE (job_uuid = :jobUuid OR job_symlink_target_uuid = :jobUuid)
+    AND io_type = :ioType
+    AND is_current_job_version = TRUE;
+  """)
+  void markInputOrOutputDatasetAsPreviousFor(UUID jobUuid, IoType ioType);
+
+  /**
+   * Used to link an input dataset to a given job version.
+   *
+   * @param inputDatasetUuid The unique ID of the input dataset.
+   * @param jobUuid The unique ID of the job.
+   */
+  default void upsertInputDatasetFor(
+      UUID jobVersionUuid, UUID inputDatasetUuid, UUID jobUuid, UUID symlinkTargetJobUuid) {
+    markInputOrOutputDatasetAsPreviousFor(jobVersionUuid, jobUuid, IoType.INPUT);
+    upsertCurrentInputOrOutputDatasetFor(
+        jobVersionUuid, inputDatasetUuid, jobUuid, symlinkTargetJobUuid, IoType.INPUT);
+  }
+
+  /**
+   * Used to link an output dataset to a given job version.
+   *
+   * @param outputDatasetUuid The unique ID of the output dataset.
+   * @param jobUuid The unique ID of the job.
+   */
+  default void upsertOutputDatasetFor(
+      UUID jobVersionUuid, UUID outputDatasetUuid, UUID jobUuid, UUID symlinkTargetJobUuid) {
+    markInputOrOutputDatasetAsPreviousFor(jobVersionUuid, jobUuid, IoType.OUTPUT);
+    upsertCurrentInputOrOutputDatasetFor(
+        jobVersionUuid, outputDatasetUuid, jobUuid, symlinkTargetJobUuid, IoType.OUTPUT);
+  }
 
   /**
    * Returns the input datasets to a given job version.
@@ -366,14 +399,20 @@ public interface JobVersionDao extends BaseDao {
     inputs.forEach(
         i -> {
           jobVersionDao.upsertInputDatasetFor(
-              jobVersionRow.getUuid(), i.getDatasetVersionRow().getDatasetUuid());
+              jobVersionRow.getUuid(),
+              i.getDatasetVersionRow().getDatasetUuid(),
+              jobVersionRow.getJobUuid(),
+              jobRow.getSymlinkTargetId());
         });
 
     // Link the output datasets to the job version.
     outputs.forEach(
         o -> {
           jobVersionDao.upsertOutputDatasetFor(
-              jobVersionRow.getUuid(), o.getDatasetVersionRow().getDatasetUuid());
+              jobVersionRow.getUuid(),
+              o.getDatasetVersionRow().getDatasetUuid(),
+              jobVersionRow.getJobUuid(),
+              jobRow.getSymlinkTargetId());
         });
 
     jobDao.updateVersionFor(jobRow.getUuid(), jobRow.getCreatedAt(), jobVersionRow.getUuid());
@@ -468,14 +507,20 @@ public interface JobVersionDao extends BaseDao {
     jobVersionInputs.forEach(
         jobVersionInput -> {
           jobVersionDao.upsertInputDatasetFor(
-              jobVersionRow.getUuid(), jobVersionInput.getDatasetUuid());
+              jobVersionRow.getUuid(),
+              jobVersionInput.getDatasetUuid(),
+              jobVersionRow.getJobUuid(),
+              jobRow.getSymlinkTargetId());
         });
 
     // Link the output datasets to the job version.
     jobVersionOutputs.forEach(
         jobVersionOutput -> {
           jobVersionDao.upsertOutputDatasetFor(
-              jobVersionRow.getUuid(), jobVersionOutput.getDatasetUuid());
+              jobVersionRow.getUuid(),
+              jobVersionOutput.getDatasetUuid(),
+              jobVersionRow.getJobUuid(),
+              jobRow.getSymlinkTargetId());
         });
 
     // Link the job version to the run.
--- a/api/src/main/java/marquez/db/LineageDao.java
+++ b/api/src/main/java/marquez/db/LineageDao.java
@@ -56,43 +56,45 @@ public interface LineageDao {
   @SqlQuery(
       """
       WITH RECURSIVE
-           -- Find the current version of a job or its symlink target if the target has no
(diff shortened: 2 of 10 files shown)


api/src/main/java/marquez/db/DatasetFacetsDao.java
```diff
@@ -126,9 +126,9 @@ void insertDatasetFacet(
   default void insertDatasetFacetsFor(
       @NonNull UUID datasetUuid,
       @NonNull UUID datasetVersionUuid,
-      @NonNull UUID runUuid,
+      UUID runUuid,
       @NonNull Instant lineageEventTime,
-      @NonNull String lineageEventType,
+      String lineageEventType,
```
wslulciuc, line 131: Can we use `DATASET` or `JOB` as  the `lineageEventType`? I think it would be helpful to know the event type.

api/src/main/resources/marquez/db/migration/V66__job_versions_io_mapping_add_job_reference.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE job_versions_io_mapping ADD COLUMN job_uuid uuid REFERENCES jobs(uuid) ON DELETE CASCADE;
+
+CREATE INDEX job_versions_io_mapping_job_uuid ON job_versions_io_mapping (job_uuid);
+
+UPDATE job_versions_io_mapping
```
wslulciuc, line 5: Do you think it'd be helpful to split the `ALTER` and `UPDATE` ... `SET` into separate migrations? That way, the migration for `ALTER` and new index will be separate from the migration to set the new `job_uuid ` col (which may or may not take a long time to compelete).

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -93,6 +95,14 @@ void createLineageEvent(
       PGobject event,
       String producer);
 
+  @SqlUpdate(
+      "INSERT INTO lineage_events ("
+          + "event_time, "
+          + "event, "
+          + "producer) "
+          + "VALUES (?, ?, ?)")
+  void createLineageEvent(Instant eventTime, PGobject event, String producer);
```
wslulciuc, line 104: Minor: Although the params are generic, this DAO method would only be used for inserting `DatasetEvent`s. I think for clarity, I would use `createDatasetEvent()`. Also, do the remaining columns in the `lineage_events` default to `null`?

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -135,6 +145,80 @@ default UpdateLineageRow updateMarquezModel(LineageEvent event, ObjectMapper map
     return updateLineageRow;
   }
 
+  default UpdateLineageRow updateMarquezModel(DatasetEvent event, ObjectMapper mapper) {
```
wslulciuc, line 148: Minor: There's a far amount of duplication between `updateMarquezModel(DatasetEvent)` and `updateMarquezModel(LineageEvent)`; more of an observation than anything else. We should consider planning future refactoring work. But, for now, this implementation is totally reasonable.

api/src/main/resources/marquez/db/migration/V66__job_versions_io_mapping_add_job_reference.sql
```diff
@@ -0,0 +1,8 @@
...
+SET job_uuid = job_versions.job_uuid
+FROM job_versions
+WHERE job_versions_io_mapping.job_version_uuid = job_versions.version;
```
wslulciuc, line 8: Minor: You want to update your `WHERE` clause to use `job_versions.uuid`:

```
WHERE job_versions_io_mapping.job_version_uuid = job_versions.uuid;
```

api/src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -206,15 +210,17 @@ default void upsertOutputDatasetFor(UUID jobVersionUuid, UUID outputDatasetUuid)
...
+   * @param jobUuid The unique ID of the job.
    */
   @SqlUpdate(
       """
     INSERT INTO job_versions_io_mapping (
-      job_version_uuid, dataset_uuid, io_type)
-    VALUES (:jobVersionUuid, :datasetUuid, :ioType)
+      job_version_uuid, dataset_uuid, io_type, job_uuid)
+    VALUES (:jobVersionUuid, :datasetUuid, :ioType, :jobUuid)
     ON CONFLICT DO NOTHING
```
wslulciuc, line 220: There's a unique index on:

```
job_version_uuid, dataset_uuid, io_type
```

We'll need to update the index to also include `job_uuid`

wslulciuc: I was able to test inserting  static lineage `job` event in `job_versions_io_mapping`, but I had to make the following changes to the table:

```sql
# (1) To allow for job_version_uuid = NULL, we have to remove the foreign key constraint
ALTER TABLE job_versions_io_mapping DROP CONSTRAINT job_versions_io_mapping_job_version_uuid_fkey;
ALTER TABLE job_versions_io_mapping ALTER COLUMN job_version_uuid DROP NOT NULL;

# (2) To add job_uuid to the unique constraint, we first drop the primary key, then recreate it; note given that job_version_uuid can be NULL, we need to check that job_version_uuid != NULL before inserting (duplicate columns otherwise)
ALTER TABLE job_versions_io_mapping DROP CONSTRAINT job_versions_io_mapping_pkey;
ALTER TABLE job_versions_io_mapping ADD CONSTRAINT job_versions_io_mapping_pkey UNIQUE (job_version_uuid,dataset_uuid,io_type,job_uuid);

# (3) Simulate static lineage job event (no job version)
INSERT INTO job_versions_io_mapping (job_version_uuid, dataset_uuid, io_type, job_uuid)
  VALUES (NULL, '4abaa4d6-f874-4d99-a7a4-d0791352b62c', 'INPUT', '86f988b5-3a94-4ded-9a19-fd2a062b0c3a');
 
# (4) Simulate run-level event (with job version)
INSERT INTO job_versions_io_mapping (job_version_uuid, dataset_uuid, io_type, job_uuid)
  VALUES ('c5894e50-c459-4667-a862-3e9aa3051dfa', '4abaa4d6-f874-4d99-a7a4-d0791352b62c', 'INPUT', '86f988b5-3a94-4ded-9a19-fd2a062b0c3a');
```

api/src/main/java/marquez/db/DatasetFacetsDao.java
pawel-big-lebowski: Currently, this column holds run states and its name is kind of misleading. Its name fits static lineage scenario well. However, I don't think we should store in a single column run-state and event-type. Renaming this column would require significant amount of work witch changes all over the project, including the spec.

Assuming this column contains run state update, then keeping it to `null` for `DatasetEvent` makes sense to me.

api/src/main/java/marquez/db/OpenLineageDao.java
pawel-big-lebowski: Yes, so my thought was that trying to make it generic for dataset, job and run event will make it shorter, but less readable. So this is the reason behind duplication. Also, I didn't want to modify existing method without any discussion on that. 

I would go for some improvements, like:
 1. create value class `OpenLineageDao.Daos` with all the daos required and created on demand. For example, `upsertLineageDataset` takes 8 daos arguments. 
 2. Introducing builders for methods like `upsertJob` would reduce already existing code duplication: calling method twice with 9 same params and one different. Same is happening `runDao.upsert`
 3. Introducing methods `insertInputFacets` and `insertOutputFacets`

I don't mind working on those in this PR as along as it is still readable for the reviewer. Let me know what do you think?

api/src/main/resources/marquez/db/migration/V66__job_versions_io_mapping_add_job_reference.sql
pawel-big-lebowski: yeah, added that. I think this deserves an extra test (added todo within the pr for that)

api/src/main/java/marquez/db/JobVersionDao.java
pawel-big-lebowski: The uniqueness of `job_version_uuid, dataset_uuid, io_type`, again, deserves a test which I am happy to implement. 

I think we should not accept null values for `job_version`. It would not make sense for table called `job_versions_io_mapping`.  I think we can leave in that case `job_versions_io_mapping_pkey` as is.

pawel-big-lebowski:
@wslulciuc I've added another commit to the PR keeping in mind upcoming streaming job support and our offline discussion. 

Based on that, I think we should rename `job_versions_io_mapping` to `job_io_mapping` and make `job_version_uuid` nullable to be set at the end of the job.

julienledem:
That's sounds fine to me. I'm curious to hear what Willy thinks.

api/src/main/java/marquez/db/DatasetFacetsDao.java
wslulciuc: I agree it's misleading, but Marquez doesn't necessarily have to adhere exactly to OpenLineage concepts and can be redefine them. We can set lineage_event_type` as `null `, but would prefer we (eventually) remap `eventType` to `runState`.

api/src/main/java/marquez/db/OpenLineageDao.java
wslulciuc: Hmm, we can hold off on the refactoring. Personally, we should look into using a jdbi [`handler`](https://jdbi.org/#_obtaining_a_managed_handle) and not restrict ourselves to using DAOs (which really are great for simpler queries, but less ideal for more complex ones).

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -56,43 +56,45 @@ public record UpstreamRunRow(JobSummary job, RunSummary run, DatasetSummary inpu
...
-      FROM lineage l2
-      INNER JOIN jobs_view j ON j.uuid=l2.job_uuid;
+                 job_io AS (
+                    SELECT
+                           io.job_uuid AS job_uuid,
+                           io.symlink_target_job_uuid AS symlink_target_job_uuid,
+                           ARRAY_AGG(DISTINCT io.dataset_uuid) FILTER (WHERE io.io_type='INPUT') AS inputs,
+                           ARRAY_AGG(DISTINCT io.dataset_uuid) FILTER (WHERE io.io_type='OUTPUT') AS outputs
+                    FROM job_io_mapping io
+                    WHERE io.is_job_version_current = TRUE
```
wslulciuc, line 66: I like the idea of using a current version check, I'd be interesting to see the query plan and how the query may be optimized with the removal of join on that `jobs` table.

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -56,43 +56,45 @@ public record UpstreamRunRow(JobSummary job, RunSummary run, DatasetSummary inpu
...
+                           0 AS depth
+                    FROM job_io
+                    WHERE job_uuid IN (<jobIds>) OR symlink_target_job_uuid IN (<jobIds>)
+                    UNION
+                    SELECT io.job_uuid, io.symlink_target_job_uuid, io.inputs, io.outputs, l.depth + 1
+                    FROM job_io io, lineage l
+                    WHERE (io.job_uuid != l.job_uuid) AND
+                        array_cat(io.inputs, io.outputs) && array_cat(l.inputs, l.outputs)
+                      AND depth < :depth),
+                lineage_outside_job_io(job_uuid) AS (
```
wslulciuc, line 83: How is `lineage_outside_job_io()` used?

api/src/main/resources/marquez/db/migration/V67.1__job_versions_io_mapping_add_job_reference.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE job_versions_io_mapping ADD COLUMN job_uuid uuid REFERENCES jobs(uuid) ON DELETE CASCADE;
+ALTER TABLE job_versions_io_mapping ADD COLUMN symlink_target_job_uuid uuid REFERENCES jobs(uuid) ON DELETE CASCADE;
```
wslulciuc, line 2: Minor: `symlink_target_job_uuid` -> `job_symlink_target_uuid` since it's defined in the `jobs` table as `jobs.symlink_target_uuid`

api/src/main/resources/marquez/db/migration/V67.1__job_versions_io_mapping_add_job_reference.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE job_versions_io_mapping ADD COLUMN is_job_version_current boolean DEFAULT FALSE;
+
+-- To add job_uuid to the unique constraint, we first drop the primary key, then recreate it; note given that job_version_uuid can be NULL, we need to check that job_version_uuid != NULL before inserting (duplicate columns otherwise)
+ALTER TABLE job_versions_io_mapping DROP CONSTRAINT job_versions_io_mapping_pkey;
+
+ALTER TABLE job_versions_io_mapping RENAME TO job_io_mapping;
```
wslulciuc, line 8: Do we need to rename `job_versions_io_mapping` to `job_io_mapping`? It's a major change and would favor adding the additional columns to `job_versions_io_mapping` without the rename if we can avoid it.

api/src/main/resources/marquez/db/migration/V67.1__job_versions_io_mapping_add_job_reference.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE job_versions_io_mapping ADD COLUMN job_uuid uuid REFERENCES jobs(uuid) ON DELETE CASCADE;
+ALTER TABLE job_versions_io_mapping ADD COLUMN symlink_target_job_uuid uuid REFERENCES jobs(uuid) ON DELETE CASCADE;
+ALTER TABLE job_versions_io_mapping ADD COLUMN is_job_version_current boolean DEFAULT FALSE;
```
wslulciuc, line 3: Minor: `is_job_version_current` -> `is_current_job_version`

api/src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -192,40 +192,73 @@ ExtendedJobVersionRow upsertJobVersion(
...
    */
-  default void upsertInputDatasetFor(UUID jobVersionUuid, UUID inputDatasetUuid) {
-    upsertInputOrOutputDatasetFor(jobVersionUuid, inputDatasetUuid, IoType.INPUT);
-  }
+  @SqlUpdate(
+      """
+    INSERT INTO job_io_mapping (
+      job_version_uuid, dataset_uuid, io_type, job_uuid, symlink_target_job_uuid, is_job_version_current)
+    VALUES (:jobVersionUuid, :datasetUuid, :ioType, :jobUuid, :symlinkTargetJobUuid, TRUE)
+    ON CONFLICT (job_version_uuid, dataset_uuid, io_type, job_uuid) DO UPDATE SET is_job_version_current = TRUE
```
wslulciuc, line 207: We set `is_job_version_current = TRUE` as a noop? i.e. just to fulfill the `ON CONFLICT`? Should we just use `DO NOTHING` instead?

api/src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -192,40 +192,73 @@ ExtendedJobVersionRow upsertJobVersion(
...
+
+  @SqlUpdate(
+      """
+    UPDATE job_io_mapping
+    SET is_job_version_current = FALSE
+    WHERE (job_uuid = :jobUuid OR symlink_target_job_uuid = :jobUuid)
+    AND io_type = :ioType
+    AND is_job_version_current = TRUE;
+  """)
+  void markVersionIOMappingObsolete(UUID jobUuid, IoType ioType);
```
wslulciuc, line 235: Is this marking the I/O type for the job as obsolete? I would name the function `markInputOrOutputDatasetAsPreviousFor()`. Also, we don't need the job version?

api/src/main/java/marquez/db/LineageDao.java
pawel-big-lebowski: One can have a job which does have any input nor output datasets. It's a node with the graph with no edges. It should be returned a single node if someone calls getLineage on this job.

api/src/main/java/marquez/db/JobVersionDao.java
pawel-big-lebowski: No. Because before doing that we mark all versions as obsolete in `markVersionIOMappingObsolete`

api/src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -192,40 +192,73 @@ ExtendedJobVersionRow upsertJobVersion(
...
+  @SqlUpdate(
+      """
+    UPDATE job_io_mapping
+    SET is_job_version_current = FALSE
+    WHERE (job_uuid = :jobUuid OR symlink_target_job_uuid = :jobUuid)
+    AND job_version_uuid != :jobVersionUuid
+    AND io_type = :ioType
+    AND is_job_version_current = TRUE;
+  """)
+  void markVersionIOMappingObsolete(UUID jobVersionUuid, UUID jobUuid, IoType ioType);
```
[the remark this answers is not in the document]
pawel-big-lebowski: renamed to `markInputOrOutputDatasetAsPreviousFor` as suggested in the next comment.

api/src/main/java/marquez/db/JobVersionDao.java
pawel-big-lebowski: job version is needed to mark all rows obsolete except for the ones within currently inserted job version.

wslulciuc (approved):
Looking forward to the lineage query perf improvements and follow up analysis!

PR #2656, opened 2023-10-23
Add pagination to JobDetailPage

merobi-hub:
### Problem

We don't yet have pagination on the JobDetailPage's Run History tab for runs like we do for datasets and jobs on those routes.

### Solution

This uses the same approach employed in those cases to add pagination to the JobDetailPage. As in those cases, here, the API is queried to ensure an accurate count. But, unlike those cases, we lacked a query to get the total run count, so one has been added to RunDao. 

One-line summary: adds pagination to JobDetailPage.

### Checklist

Screenshot from 2023-11-06 10-43-13


phixMe:
@merobi-hub We need to update the openapi spec and regenerate the docs for the doc section of the website as well here.

api/src/main/java/marquez/db/RunDao.java
```diff
@@ -192,6 +192,9 @@ LEFT OUTER JOIN (
       """)
   List<Run> findAll(String namespace, String jobName, int limit, int offset);
 
+  @SqlQuery("SELECT count(*) FROM runs_view AS r WHERE r.job_name = :jobName")
+  int countFor(String jobName);
```
phixMe, line 196: I think you need to add namespace here if I understand correctly to make the filter the same between the runs call and the count.

merobi-hub: Thanks, @phixMe . When I try adding the namespace to the filter query, the API returns an error saying that the runs_view lacks a namespace column. The logs also show that the endpoint already receives the correct namespace. So is this OK as-is, or is some other change required?

phixMe: You'd have to compare the runs query to the total count query... They need to be exactly the same except the count query should just have the `select count` at the beginning.

merobi-hub:
@phixMe When I was working on this, I couldn't diagnose the issues with the API. But I'll take another look when I have a moment.

PR #2657, opened 2023-10-23
provide simple implementation of one-level lineage optimized for parent jobs

julienledem:
### Problem

The main lineage graph API focuses on individual jobs and is not easy to use when one wants coverage of a all the children of a parent job.

### Solution

This new endpoint provides a non-recursive one level of lineage for all the children of a given parent job.
This will facilitate for example if someone wants to retrieve all the lineage of a given Airflow DAG.
It will return all its children (tasks) and all the datasets they consume or produce as well as the other tasks and DAGs producing and consuming them.

Example:
GET /api/v1/lineage/simple?nodeId=job:default:order_analysis
```
{
  "parent": {
    "namespace": "default",
    "name": "order_analysis"
  },
  "children": [
    {
      "job": {
        "namespace": "default",
        "name": "order_analysis.find_popular_products"
      },
      "inputs": [
        {
          "dataset": {
            "namespace": "postgres://host.docker.internal:5435",
            "name": "postgres.public.orders"
          },
          "consumers": null,
          "producers": [
            {
               "job": {
                "namespace": "default",
                "name": "order_analysis.import_orders"
               },
               "parent": {
                  "namespace": "default", 
                  "name": "order_analysis"
               }
            }
          ]
        }
      ],
     "outputs": [
         ...   
      ]
    ),
    ...
  ]
}
```

### Checklist

### DIFF
--- a/api/src/main/java/marquez/api/OpenLineageResource.java
+++ b/api/src/main/java/marquez/api/OpenLineageResource.java
@@ -111,6 +111,22 @@ public class OpenLineageResource extends BaseResource {
   @ResponseMetered
   @ExceptionMetered
   @GET
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  @Path("/lineage/direct")
+  public Response getSimpleLineage(
+      @QueryParam("parentJobNodeId") @NotNull NodeId parentJobNodeId) {
+    if (!parentJobNodeId.isJobType()) {
+      throw new IllegalArgumentException("Only job expected, got " + parentJobNodeId.getValue());
+    }
+    throwIfNotExists(parentJobNodeId);
+    return Response.ok(lineageService.parentDirectLineage(parentJobNodeId.asJobId())).build();
+  }
+
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
   @Path("/events/lineage")
   @Produces(APPLICATION_JSON)
   public Response getLineageEvents(
--- a/api/src/main/java/marquez/db/LineageDao.java
+++ b/api/src/main/java/marquez/db/LineageDao.java
@@ -10,23 +10,42 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.BindList;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+import marquez.common.models.DatasetId;
+import marquez.common.models.JobId;
 import marquez.db.mappers.DatasetDataMapper;
 import marquez.db.mappers.JobDataMapper;
 import marquez.db.mappers.JobRowMapper;
 import marquez.db.mappers.RunMapper;
+import marquez.db.mappers.SimpleLineageEdgeMapper;
 import marquez.service.models.DatasetData;
 import marquez.service.models.JobData;
 import marquez.service.models.Run;
-import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
-import org.jdbi.v3.sqlobject.customizer.BindList;
-import org.jdbi.v3.sqlobject.statement.SqlQuery;
 
 @RegisterRowMapper(DatasetDataMapper.class)
 @RegisterRowMapper(JobDataMapper.class)
 @RegisterRowMapper(RunMapper.class)
 @RegisterRowMapper(JobRowMapper.class)
+@RegisterRowMapper(SimpleLineageEdgeMapper.class)
 public interface LineageDao {
 
+  public record SimpleLineage(Collection<SimpleLineageEdge> edges) {
+  }
+
+  public record SimpleLineageEdge(
+      JobId job1,
+      String direction,
+      DatasetId dataset,
+      String direction2,
+      JobId job2,
+      JobId job2parent
+  ) {
+
+  }
   /**
    * Fetch all of the jobs that consume or produce the datasets that are consumed or produced by the
    * input jobIds. This returns a single layer from the BFS using datasets as edges. Jobs that have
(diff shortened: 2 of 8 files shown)


api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -25,8 +28,19 @@
 @RegisterRowMapper(JobDataMapper.class)
 @RegisterRowMapper(RunMapper.class)
 @RegisterRowMapper(JobRowMapper.class)
+@RegisterRowMapper(SimpleLineageEdgeMapper.class)
 public interface LineageDao {
 
+  public record SimpleLineage(Collection<SimpleLineageEdge> edges) {}
```
pawel-big-lebowski, line 34: Instead of `SimpleLineage`, I would prefer `DirectLineage` or `AdjacentLineage` (as adjacent nodes in graph)

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -79,6 +93,39 @@ SELECT DISTINCT ON (j.uuid) j.*, inputs AS input_uuids, outputs AS output_uuids
...
+  @SqlQuery(
+      """
+      SELECT
+          jobs.namespace_name AS job_namespace, jobs.simple_name AS job_name,
+          jvim.io_type AS io1,
+          d.namespace_name AS ds_namespace, d."name" AS ds_name,
+          jvim2.io_type AS io2,
+          jv2.namespace_name AS job2_namespace, jv2.job_name  AS job2_name,
+          pj.namespace_name AS job2_parent_namespace, pj.simple_name AS job2_parent_name
+      FROM jobs
```
pawel-big-lebowski, line 112: I thought we should always make use os `jobs_view` in order to support feature of hidden jobs. 
This would also limit one join I believe.

api/src/main/java/marquez/db/mappers/SimpleLineageEdgeMapper.java
```diff
@@ -0,0 +1,53 @@
...
+import marquez.common.models.DatasetId;
+import marquez.common.models.DatasetName;
+import marquez.common.models.JobId;
+import marquez.common.models.JobName;
+import marquez.common.models.NamespaceName;
+import marquez.db.LineageDao.SimpleLineageEdge;
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.statement.StatementContext;
+
+public final class SimpleLineageEdgeMapper implements RowMapper<SimpleLineageEdge> {
```
pawel-big-lebowski, line 22: some javadoc would be helpful

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -79,6 +93,39 @@ SELECT DISTINCT ON (j.uuid) j.*, inputs AS input_uuids, outputs AS output_uuids
...
+      INNER JOIN job_versions jv ON jv.uuid = jobs.current_version_uuid
+      LEFT JOIN job_versions_io_mapping jvim ON jvim.job_version_uuid = jobs.current_version_uuid
+      LEFT JOIN datasets d ON d.uuid = jvim.dataset_uuid
+      LEFT JOIN job_versions_io_mapping jvim2 ON jvim2.dataset_uuid = d.uuid AND jvim2.job_version_uuid <> jvim.job_version_uuid and jvim2.io_type <> jvim.io_type
+      LEFT JOIN job_versions jv2 ON jv2.uuid = jvim2.job_version_uuid
+      LEFT JOIN jobs j2 ON jv2.job_uuid = j2.uuid
+      LEFT JOIN jobs pj ON j2.parent_job_uuid = pj.uuid
+      WHERE jobs.parent_job_uuid IN (
```
pawel-big-lebowski, line 120: How does it differ to current lineage endpoint? Isn't this just `get jobIds of jobs with parent_job X` and `get lineage with depth 1 for those jobIds`. I understand the difference at the REST api, but would love to understand better why do we need separate DAO methods for that. 

Please also be aware of parallel effort to add `job_id` column to `job_versions_io_mapping` (https://github.com/MarquezProject/marquez/pull/2654/files) which limits amount of joins required for this query.

pawel-big-lebowski (commented):
I like the feature. I put some questions in comments as I would like to understand more why do we need separate DAO methods to support this API call.

api/src/main/java/marquez/db/LineageDao.java
julienledem: Ah yes, I renamed the endpoint and then forgot to do it everywhere.
Do you prefer `direct` or `adjacent`?

api/src/main/java/marquez/db/LineageDao.java
julienledem: just replace jobs by jobs_view here? What do you mean by "limit one join"?

api/src/main/java/marquez/db/mappers/SimpleLineageEdgeMapper.java
julienledem: Ah yes, this is a first draft. If it looks good enough I'll add doc. I also need to update the openapi spec

api/src/main/java/marquez/db/LineageDao.java
julienledem: This is a good point. I could retrieve the list of jobs for the parent and pass that to the current lineage endpoint with a depth of 0 (or 1?). The new endpoint is trying to fetch less data as well. As the existing lineage endpoint fetched a lot. I can separately add a ?parentNode option to the /lineage endpoint that does what you suggest.

pawel-big-lebowski: As discussed offline, writing single purpose queries may be more efficient although some similarity exists.

api/src/main/java/marquez/db/LineageDao.java
pawel-big-lebowski: `jobs_view` already has parent job name and namespace: https://github.com/MarquezProject/marquez/blob/5c42dfd8cd0f2c739f493e301838bb3f4b0c2497/api/src/main/resources/marquez/db/migration/R__1_Jobs_view_and_rewrite_function.sql#L22

api/src/main/java/marquez/db/LineageDao.java
pawel-big-lebowski: Both are fine.

api/src/main/java/marquez/db/LineageDao.java
julienledem: updated

api/src/main/java/marquez/db/LineageDao.java
julienledem: updated with `direct`

api/src/test/java/marquez/OpenLineageIntegrationTest.java
```diff
@@ -390,6 +417,27 @@ public void testOpenLineageJobHierarchyAirflowIntegrationWithParentAndParentRunF
...
+        new marquez.common.models.JobId(NamespaceName.of(NAMESPACE_NAME), JobName.of(dagName));
+    String nodeId = NodeId.of(jobId).getValue();
+    HttpRequest request =
+        HttpRequest.newBuilder()
+            .uri(URI.create(baseUrl + "/api/v1/lineage/direct?parentJobNodeId=" + nodeId))
+            .header("Content-Type", "application/json")
+            .GET()
+            .build();
+
+    HttpResponse<String> resp;
```
pawel-big-lebowski, line 431: The common practice for the tests in this class is to support newly implemented endpoints in Marquez java client and use java client to send api calls and deserialise responses, which allows more accurate tests.

api/src/main/java/marquez/service/LineageService.java
```diff
@@ -41,13 +47,87 @@
...
+  /**
+   * This method is specialized for returning one level of lineage from a parent job. It finds all
+   * the children of the provided parent node It then finds the input and output datasets those
+   * children write to. It finally returns the other jobs consuming or producing those datasets (and
+   * their parent).
+   *
+   * @param parentJobId the parent job
+   * @return 1 level of lineage for all the children jobs of the given parent
+   */
+  public ParentLineage parentDirectLineage(JobId parentJobId) {
```
pawel-big-lebowski, line 79: a unit test for this would be nice

api/src/test/java/marquez/OpenLineageIntegrationTest.java
julienledem: done

api/src/main/java/marquez/service/LineageService.java
julienledem: done

api/src/test/java/marquez/db/LineageDaoTest.java
```diff
@@ -165,6 +165,9 @@ public void testGetLineage() {
           .containsAll(
               expected.getOutput().map(ds -> ds.getDatasetRow().getUuid()).stream()::iterator);
     }
+
+
+    lineageDao.getDirectLineageFromParent("foo", "bar");
```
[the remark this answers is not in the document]
julienledem: fixed

api/src/main/java/marquez/db/mappers/SimpleLineageEdgeMapper.java
julienledem: added

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -107,6 +107,21 @@ public Response getLineage(
...
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  @Path("/lineage/direct")
+  public Response getDirectLineage(@QueryParam("parentJobNodeId") @NotNull NodeId parentJobNodeId) {
```
pawel-big-lebowski, line 117: Please mind updating `openapi.yaml` and `changeling`

pawel-big-lebowski (commented):
Two minor comments added.
I think that SQL query and tests are already fine.

dkt-sophie-ly:
Hi ! Do you know if you will continue to work on this PR ? This feature seems quite interesting.
I tried to update it but it's not that easy given that this PR is a year old :/

PR #2658, opened 2023-10-24
upstream run level lineage implementation

julienledem:
### Problem

When trouble shooting an issue and doing root cause analysis, it is usefull to get the upstream run-level lineage to know exactly what version of each job and dataset a run is depending on.

### Solution

return the upstream run level lineage.

Screen Shot 2023-10-24 at 2 56 07 PM

Example:
http://localhost:3000/api/v1/runlineage/upstream?runId=adc8507c-595e-4d76-9dac-be2bf0ffe1ee
```json
{
    "runs": [
        {
            "job": {
                "namespace": "food_delivery",
                "name": "orders_popular_day_of_week",
                "version": "a8f00e6e-88b0-4e5b-a0d8-e7edd3048292"
            },
            "run": {
                "id": "adc8507c-595e-4d76-9dac-be2bf0ffe1ee",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:46:12Z",
                "status": "COMPLETED"
            },
            "inputs": [
                {
                    "namespace": "food_delivery",
                    "name": "public.top_delivery_times",
                    "version": "18951d81-7201-38ca-a207-3f65c7552373",
                    "producedByRunId": "bd41a42a-bf18-4b74-9bb7-cd62637823d8"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.customers",
                    "version": "68c1e307-f6bb-36f9-8596-14609c7f022b",
                    "producedByRunId": "182a9eaf-881a-4d49-860c-f7e260b8bf60"
                }
            ]
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "delivery_times_7_days",
                "version": "a3ea0de5-f86b-48a5-a5de-5e53a559f59d"
            },
            "run": {
                "id": "bd41a42a-bf18-4b74-9bb7-cd62637823d8",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:58:02Z",
                "status": "COMPLETED"
            },
            "inputs": [
                {
                    "namespace": "food_delivery",
                    "name": "public.delivery_7_days",
                    "version": "6f8f52f5-0230-31ce-a138-08b79e671b33",
                    "producedByRunId": "d5a2a4c4-fc78-428d-ae85-08c942ed8371"
                }
            ]
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_customers",
                "version": "f57c837f-8ef3-49be-98d4-14e4195518da"
            },
            "run": {
                "id": "182a9eaf-881a-4d49-860c-f7e260b8bf60",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:55Z",
                "status": "COMPLETED"
            },
            "inputs": []
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_delivery_7_days",
                "version": "312ac05b-6022-4fce-af4c-16728b27d605"
            },
            "run": {
                "id": "d5a2a4c4-fc78-428d-ae85-08c942ed8371",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:48:12Z",
                "status": "COMPLETED"
            },
            "inputs": [
                {
                    "namespace": "food_delivery",
                    "name": "public.orders_7_days",
                    "version": "d09633c4-4412-36de-bce6-8002c662e18a",
                    "producedByRunId": "ffba2c14-4170-48da-bec3-ab5fd4ec9a3f"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.customers",
                    "version": "68c1e307-f6bb-36f9-8596-14609c7f022b",
                    "producedByRunId": "182a9eaf-881a-4d49-860c-f7e260b8bf60"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.order_status",
                    "version": "676ac323-c8e3-3cea-b172-b468827afb51",
                    "producedByRunId": "b7098939-87f0-4207-878f-dfd8e8804d8a"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.drivers",
                    "version": "93ae26cc-87d8-3eae-9cdd-f9b6fd71f1f7",
                    "producedByRunId": "9f3db1c5-5e9a-4280-8184-18aca4592c77"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.restaurants",
                    "version": "4db26821-7966-390f-9cf9-ac775fe9182b",
                    "producedByRunId": "8ddfb1d9-415f-4850-bcd6-01d02f011abe"
                }
            ]
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_orders_7_days",
                "version": "d6014479-a795-48bf-b620-a4f781436600"
            },
            "run": {
                "id": "ffba2c14-4170-48da-bec3-ab5fd4ec9a3f",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:02Z",
                "status": "COMPLETED"
            },
            "inputs": [
                {
                    "namespace": "food_delivery",
                    "name": "public.menus",
                    "version": "54a67439-3426-3272-b825-c8dabc9da6c1",
                    "producedByRunId": "4d3b8069-69b6-4708-ade0-3275112c9f04"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.menu_items",
                    "version": "88618612-23c0-3fe5-89e7-d939883a2f52",
                    "producedByRunId": "e05901b1-3a06-4b98-8d9c-aaf188c9a28c"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.orders",
                    "version": "1e3f9ed6-eea2-3ff4-83ab-4efa9dc87f1c",
                    "producedByRunId": "a43a8523-349f-4296-807f-3354ac491990"
                },
                {
                    "namespace": "food_delivery",
                    "name": "public.categories",
                    "version": "20a6093a-6edc-3c58-b3ec-43d0efc14035",
                    "producedByRunId": "6f0c13a5-f29b-46a5-90c1-0ffbebbbd1aa"
                }
            ]
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_order_status",
                "version": "ca990d26-7909-4392-a3b8-f9415a6fd304"
            },
            "run": {
                "id": "b7098939-87f0-4207-878f-dfd8e8804d8a",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:52Z",
                "status": "COMPLETED"
            },
            "inputs": []
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_restaurants",
                "version": "4d9a0f8a-98e3-4081-8f4e-e18048881eb9"
            },
            "run": {
                "id": "8ddfb1d9-415f-4850-bcd6-01d02f011abe",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:56Z",
                "status": "COMPLETED"
            },
            "inputs": []
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_drivers",
                "version": "7325dcfd-9250-489f-8b79-993792d0a9be"
            },
            "run": {
                "id": "9f3db1c5-5e9a-4280-8184-18aca4592c77",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:52Z",
                "status": "COMPLETED"
            },
            "inputs": []
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_menus",
                "version": "6e761ef2-1173-4bb9-a80d-1a159982e9ca"
            },
            "run": {
                "id": "4d3b8069-69b6-4708-ade0-3275112c9f04",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:45:52Z",
                "status": "COMPLETED"
            },
            "inputs": []
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_menu_items",
                "version": "07341e00-85fe-47b2-93c9-8c37685c2124"
            },
            "run": {
                "id": "e05901b1-3a06-4b98-8d9c-aaf188c9a28c",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:54Z",
                "status": "COMPLETED"
            },
            "inputs": []
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_categories",
                "version": "133be69f-8648-4e00-903a-2282e895d3b7"
            },
            "run": {
                "id": "6f0c13a5-f29b-46a5-90c1-0ffbebbbd1aa",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:52Z",
                "status": "COMPLETED"
            },
            "inputs": []
        },
        {
            "job": {
                "namespace": "food_delivery",
                "name": "etl_orders",
                "version": "12dca2cf-c0ed-491f-8ad6-d4813f9c7cb6"
            },
            "run": {
                "id": "a43a8523-349f-4296-807f-3354ac491990",
                "start": "2020-02-22T22:42:42Z",
                "end": "2020-02-22T22:44:52Z",
                "status": "COMPLETED"
            },
            "inputs": []
        }
    ]
}
```

### Checklist


julienledem:
TODO:
tests and doc

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -154,4 +171,41 @@ SELECT DISTINCT on(r.job_name, r.namespace_name) r.*, jv.version as job_version
...
+          u_r_uuid, depth) AS (
+    SELECT
+          r.uuid, r.started_at, r.ended_at, r.current_run_state,
+          r.job_uuid, r.job_version_uuid, r.namespace_name, r.job_name,
+          dv.dataset_uuid, dv."version", dv.namespace_name, dv.dataset_name,
+          dv.run_uuid,
+          0 AS depth
+    FROM runs r
+    LEFT JOIN runs_input_mapping rim ON rim.run_uuid = r.uuid
+    LEFT JOIN dataset_versions dv ON dv.uuid = rim.dataset_version_uuid
```
pawel-big-lebowski, line 191: Just thinking loudly. Wouldn't be better to join dataset_versions after the recursion at the bottom of the query once all the runs are identified?

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -154,4 +171,41 @@ SELECT DISTINCT on(r.job_name, r.namespace_name) r.*, jv.version as job_version
...
+    WHERE r.uuid = :runId
+  UNION
+    SELECT
+          ur.u_r_uuid, r2.started_at, r2.ended_at, r2.current_run_state,
+          r2.job_uuid, r2.job_version_uuid, r2.namespace_name, r2.job_name,
+          dv2.dataset_uuid, dv2."version", dv2.namespace_name, dv2.dataset_name,
+          dv2.run_uuid,
+          ur.depth + 1 AS depth
+    FROM upstream_runs ur
+    INNER JOIN runs r2 ON r2.uuid = ur.u_r_uuid
```
pawel-big-lebowski, line 202: `runs` table can get huge. I think it doable to construct here recursion steps on the `runs_input_mapping` table only.

pawel-big-lebowski (commented):
Another PR with a moderate size and huge feature contained 👍 
First-string 💯

api/src/main/java/marquez/db/LineageDao.java
julienledem: I thought of that as well but we actually need dataset_versions in the recursion because this is where we find the run_uuid that produced the DV for the next iteration.

api/src/main/java/marquez/db/LineageDao.java
julienledem: This is where I find the job information (r2.job_uuid, r2.job_version_uuid, r2.namespace_name, r2.job_name). Do you think I could get it a different way?
unless we add it to runs_input_mapping?

julienledem: But I think I could actually move the runs join out of the recursion into the final select statement.

api/src/main/java/marquez/db/LineageDao.java
julienledem: idea: we could add that run uuid to runs_input_mapping at the same time as the dataset_version that would allow to join just on that table in the recursion. That'd be neat.

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -130,6 +131,23 @@ public Response getLineageEvents(
...
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  @Path("/runlineage/upstream")
+  public Response getRunLineageUpstream(
+      @QueryParam("runId") @NotNull RunId runId,
+      @QueryParam("depth") @DefaultValue(DEFAULT_DEPTH) int depth,
+      @QueryParam("facets") String facets) {
```
phixMe, line 144: Why are we pulling this in? I don't see it being used in the service. Did you intend to do something in the TODO block you left in there?

api/src/main/java/marquez/db/LineageDao.java
julienledem: I just updated the query to move the joins on the `runs` table outside of the recursion and in the final select statement.

api/src/main/java/marquez/api/OpenLineageResource.java
julienledem: Yes, my idea is to be able to select what facets to return for each dataset_version job_version and run in the result.
I can remove it for now.

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -154,4 +171,50 @@ SELECT DISTINCT on(r.job_name, r.namespace_name) r.*, jv.version as job_version
...
+    LEFT JOIN dataset_versions dv2 ON dv2.uuid = rim2.dataset_version_uuid
+    -- end condition of the recursion: no input or depth is over the maximum set
+    -- also avoid following cycles (merge statement)
+    WHERE ur.u_r_uuid IS NOT NULL AND ur.u_r_uuid <> ur.r_uuid AND depth < :depth
+  )
+
+  -- present the result: use Distinct as we may have traversed the same edge multiple times if there are diamonds in the graph.
+SELECT DISTINCT ON (upstream_runs.r_uuid, upstream_runs.dataset_version_uuid, upstream_runs.u_r_uuid)
+       upstream_runs.*,
+       -- we add the run information after the recursion so that we join with the large run table only once
```
pawel-big-lebowski, line 213: ❤️

api/src/main/java/marquez/db/LineageDao.java
pawel-big-lebowski: Didn't get the idea: do you want to add dataset_versions to runs_input_mapping or run_uuid to dataset_versions?
Anyway, the current state of the query looks good to me.

pawel-big-lebowski:
The overall idea looks good to me. Once the tests are created, I would be happy to approve it.

julienledem:
> The overall idea looks good to me. Once the tests are created, I would be happy to approve it.

@pawel-big-lebowski: done.

api/src/main/java/marquez/db/LineageDao.java
julienledem: The idea is we we could add the dataset_versions.run_uuid to the runs_input_mapping table. 
The run_uuid is always already set in the dataset_versions table when we write the dataset_versions.uuid to the runs_input_mapping table and it never changes since it is the run that created that dataset version.
Then you just need to recursively join the runs_input_mapping over itself

julienledem: but I would keep that for future improvement.

julienledem:
I also tested this manually on actual production data for performance and it did well.

api/src/main/java/marquez/api/OpenLineageResource.java
julienledem: I removed this parameter in this iteration

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -130,6 +131,28 @@ public Response getLineageEvents(
...
+   * @return the upstream lineage for that run up to `detph` levels
+   */
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  @Path("/runlineage/upstream")
+  public Response getRunLineageUpstream(
```
pawel-big-lebowski, line 149: Please mind documenting this in `openapi.spec`

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -154,4 +171,51 @@ SELECT DISTINCT on(r.job_name, r.namespace_name) r.*, jv.version as job_version
...
+                ur.u_r_uuid,
+                dv2.dataset_uuid, dv2."version", dv2.namespace_name, dv2.dataset_name,
+                dv2.run_uuid,
+                ur.depth + 1 AS depth -- increase depth to check end condition
+          FROM upstream_runs ur
+          LEFT JOIN runs_input_mapping rim2 ON rim2.run_uuid = ur.u_r_uuid
+          LEFT JOIN dataset_versions dv2 ON dv2.uuid = rim2.dataset_version_uuid
+          -- end condition of the recursion: no input or depth is over the maximum set
+          -- also avoid following cycles (ex: merge statement)
+          WHERE ur.u_r_uuid IS NOT NULL AND ur.u_r_uuid <> ur.r_uuid AND depth < :depth
```
pawel-big-lebowski, line 207: Have you considered any anti-join step within recursion step to avoid following cycles? Something like:
```
LEFT JOIN upstream_runs eur ON ur.u_r_uuid = eur.r_uuid WHERE eur.r_uuid is NULL
```

Otherwise we may expect recursive query result to explode. Although we do `distinct` at the bottom, this may still affect performance.

api/src/test/java/marquez/db/LineageDaoTest.java
```diff
@@ -888,4 +889,72 @@ public void testGetCurrentRunsGetsLatestRun() {
...
+  public void testGetRunLineage() {
+
+    UpdateLineageRow writeJob =
+        LineageTestUtils.createLineageRow(
+            openLineageDao,
+            "writeJob",
+            "COMPLETE",
+            jobFacet,
+            Arrays.asList(),
+            Arrays.asList(dataset));
```
pawel-big-lebowski, line 903: Would it make sense to add input dataset here read from some other job? 
Currently recursion is not tested.

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -154,4 +171,51 @@ SELECT DISTINCT on(r.job_name, r.namespace_name) r.*, jv.version as job_version
...
+
+      -- present the result: use Distinct as we may have traversed the same edge multiple times if there are diamonds in the graph.
+      SELECT * FROM ( -- we need the extra statement to sort after the DISTINCT
+          SELECT DISTINCT ON (upstream_runs.r_uuid, upstream_runs.dataset_version_uuid, upstream_runs.u_r_uuid)
+            upstream_runs.*,
+            r.started_at, r.ended_at, r.current_run_state as state,
+            r.job_uuid, r.job_version_uuid, r.namespace_name as job_namespace, r.job_name
+          FROM upstream_runs, runs r WHERE upstream_runs.r_uuid = r.uuid
+        ) sub
+      ORDER BY depth ASC;
```
pawel-big-lebowski, line 218: Would it make sense to `ORDER BY depth ASC, r.job_name ASC;`
Tests rely on the ordering of the query output.

pawel-big-lebowski (commented):
This looks good to me. 
Please look at the comments and let me know do you think.
I am happy to approve PR afterwards.

api/src/main/java/marquez/db/LineageDao.java
julienledem: I see your point. That would avoid following the same edge twice even when there's no cycle. Today it happens when there is a diamond in the graph for example. This is why I added a distinct statement at the end.
On the other end it does add an extra join on the temporary recursive table, so it's probably a trade-off in term of performance. It might slow down in the common case to speed up some corner cases. Although the temporary table upstream_runs is small so not by much.

I think this might be unnecessary since I think the only cycle that can happen on runs are pointing to themselves (this happens with the SQL merge statement when a table is an input and an output of the job). 
Here we avoid following them with "ur.u_r_uuid <> ur.r_uuid" 

I think I can add the statement you are suggesting. Then I would remove the DISTINCT constraint in the end as we'd prevent completely following the same edge twice.

What do you think?

api/src/main/java/marquez/db/LineageDao.java
julienledem: good idea

api/src/main/java/marquez/db/LineageDao.java
pawel-big-lebowski: Yeah, you're right it's just a corner case when upstream nodes are visited. 
This becomes a problem when one traverses both upstream and downstream edges at the same time and can reach the same node with different depth oftentimes.

pawel-big-lebowski (approved):

api/src/main/java/marquez/db/LineageDao.java
julienledem: done

api/src/test/java/marquez/db/LineageDaoTest.java
julienledem: done

api/src/main/java/marquez/api/OpenLineageResource.java
julienledem: done

PR #2678, opened 2023-11-05
Upgrade docker/down.sh to compose v2

merobi-hub:
### Problem

The down script in `docker` does not execute properly on at least one widely-used flavor of Linux. It turns out that the issue was actually being caused by the version of Docker Compose, not the OS.

### Solution

This migrates the script to Compose v2. This should be safe, as v1 has been deprecated for a while now.

One-line summary: migrates docker/down.sh to compose v2

### Checklist


wslulciuc (changes requested):
Since [`docker/up.sh`](https://github.com/MarquezProject/marquez/blob/main/docker/up.sh) requires docker compose, I'm not sure we should have an exception for `docker/down.sh`. My suggestion would be that the we add a check in `docker/up.sh` ensuring that docker compose is installed, if not, log a warning that only the volumes will be deleted (flags `--volumes`, `--help` would no longer be needed).

merobi-hub:
@wslulciuc I did a little more experimenting and found that a differently formatted Compose command will work on Linux. So I removed the flag (but kept the help one because, why not? -- we have a usage string) and command-removal condition. Instead, this now checks for Linux and executes a differently formatted version of the same command. It's been tested on Mint and works as long as the env vars in docker-compose.yml have been set.

docker/down.sh
```diff
@@ -22,14 +22,35 @@ project_root=$(git rev-parse --show-toplevel)
...
-  docker volume rm marquez_data && \
+API_PORT=${RANDOM}
+API_ADMIN_PORT=${RANDOM}
+WEB_PORT=${RANDOM}
+TAG=${RANDOM}
+
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+  docker compose ${compose_files} down ${compose_args}
+else
+  docker-compose $compose_files down $compose_args
```
wslulciuc, line 50: We can fully [migrate](https://docs.docker.com/compose/migrate/) off docker compose v1 to v2. That is, we can replace `docker-compose` with `docker compose` as docker compose v1 is no longer recieving updates after July 2023 (of this year).. This would also simplify the changes in the PR.

merobi-hub:
Tested again today and can confirm it works

wslulciuc (approved):

PR #2679, opened 2023-11-07
Fixes Empty Search Experience

phixMe:
### Problem

We previously had an issue where we were using the previous search value for our search function which was resulting in a bad request for the first character. This was due to the fact we were using a value from an async `setState` operation immediately after it's set.

image
Error is gone from screenshot.


wslulciuc (approved):
Thanks, @phixMe! 💯 🥇

PR #2682, opened 2023-11-15
Support streaming jobs in Marquez

pawel-big-lebowski:
### Problem

Currently, `job_version` is created once the job finishes. This assumption no longer holds for streaming jobs which can run for days or weeks while continuously writing output datasets. 

It makes sense in this case, to create job version and identify input/output datasets at the beginning of the job and modify it on the fly if a job changes. 

In particular, a `lineage` endpoint shall include streaming jobs in lineage graph as they're running and datasets being written by streaming jobs should be available (and have dataset version modified) once the job is started. 

### Solution

 * add extra logic to create job_version entry and its references before the job completes,
 * make sure new version is written only once,
 * support `JobTypeJobFacet` to determine if a job is `batch` or `streaming`
 * store job characteristics in `jobs` table. 

One-line summary:

### Checklist

### DIFF
--- a/api/src/main/java/marquez/common/Utils.java
+++ b/api/src/main/java/marquez/common/Utils.java
@@ -57,6 +57,9 @@ import marquez.common.models.Version;
 import marquez.service.models.DatasetMeta;
 import marquez.service.models.DbTableMeta;
 import marquez.service.models.LineageEvent;
+import marquez.service.models.LineageEvent.Job;
+import marquez.service.models.LineageEvent.JobFacet;
+import marquez.service.models.LineageEvent.JobTypeJobFacet;
 import marquez.service.models.LineageEvent.ParentRunFacet;
 import marquez.service.models.StreamMeta;
 import org.apache.commons.lang3.tuple.Triple;
@@ -311,6 +314,21 @@ public final class Utils {
             .runId(runId)
             .build();
     return newDatasetVersionFor(data);
+  }
+
+  /**
+   * Verifies if a job is a streaming job.
+   *
+   * @param job
+   * @return
+   */
+  public static boolean isStreamingJob(Job job) {
+    return Optional.ofNullable(job)
+        .map(Job::getFacets)
+        .map(JobFacet::getJobType)
+        .map(JobTypeJobFacet::getProcessingType)
+        .filter(type -> type.equalsIgnoreCase("STREAMING"))
+        .isPresent();
   }
 
   /**
--- a/api/src/main/java/marquez/common/Utils.java
+++ b/api/src/main/java/marquez/common/Utils.java
@@ -57,9 +57,6 @@ import marquez.common.models.Version;
 import marquez.service.models.DatasetMeta;
 import marquez.service.models.DbTableMeta;
 import marquez.service.models.LineageEvent;
-import marquez.service.models.LineageEvent.Job;
-import marquez.service.models.LineageEvent.JobFacet;
-import marquez.service.models.LineageEvent.JobTypeJobFacet;
 import marquez.service.models.LineageEvent.ParentRunFacet;
 import marquez.service.models.StreamMeta;
 import org.apache.commons.lang3.tuple.Triple;
@@ -314,21 +311,6 @@ public final class Utils {
             .runId(runId)
             .build();
     return newDatasetVersionFor(data);
-  }
-
-  /**
-   * Verifies if a job is a streaming job.
-   *
-   * @param job
-   * @return
-   */
-  public static boolean isStreamingJob(Job job) {
-    return Optional.ofNullable(job)
-        .map(Job::getFacets)
-        .map(JobFacet::getJobType)
-        .map(JobTypeJobFacet::getProcessingType)
-        .filter(type -> type.equalsIgnoreCase("STREAMING"))
-        .isPresent();
   }
 
   /**
--- a/api/src/main/java/marquez/db/JobVersionDao.java
+++ b/api/src/main/java/marquez/db/JobVersionDao.java
@@ -281,6 +281,14 @@ public interface JobVersionDao extends BaseDao {
   }
 
   /**
+   * Verifies if a job with a specified job version is present in table.
+   *
+   * @param version Version identifier
+   */
+  @SqlQuery("SELECT EXISTS (SELECT 1 FROM job_versions WHERE version = :version)")
+  boolean versionExists(UUID version);
+
+  /**
    * Returns the input or output datasets for a given job version.
    *
    * @param jobVersionUuid The unique ID of the job version.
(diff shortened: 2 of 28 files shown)


api/src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -280,6 +280,14 @@ default List<UUID> findOutputDatasetsFor(UUID jobVersionUuid) {
     return findInputOrOutputDatasetsFor(jobVersionUuid, IoType.OUTPUT);
   }
 
+  /**
+   * Returns the input or output datasets for a given job version.
```
wslulciuc, line 284: Minor: You'll want to update the java docs to describe `versionExists()`

api/src/main/java/marquez/db/JobVersionDao.java
```diff
@@ -453,11 +461,12 @@ private static ExtendedDatasetVersionRow toExtendedDatasetVersionRow(DatasetReco
    * @param transitionedAt The timestamp of the run state transition.
    * @return A {@link BagOfJobVersionInfo} object.
    */
-  default BagOfJobVersionInfo upsertJobVersionOnRunTransition(
+  default BagOfJobVersionInfo upsertJobVersionReferences(
```
wslulciuc, line 464: Do we need to update the method name? `upsertJobVersionReferences()` is far less clear and doesn't provide much in terms of the methods logic.

As an alternative I would introduce `upsertJobVersionForStreaminJob()` that calls then calls `upsertJobVersionOnRunTransition()` if a version should be created.

```java
default upsertJobVersionForStreaminJob(...) {
    final JobVersionDao jobVersionDao = createJobVersionDao();
    if (!jobVersionDao.versionExists(jobVersion.getValue())) {
      upsertJobVersionOnRunTransition(...)
    }
}
```

Then, you can avoid the added logic in `upsertJobVersionOnRunTransition()` to check if a job is a streaming job or not. We can then be more explicit further up the call stack that we are attempting to create a streaming job version by calling `upsertJobVersionForStreaminJob()`.

Note, I'd also provide extensive java docs for `upsertJobVersionForStreaminJob()`.

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -167,9 +168,21 @@ SELECT count(*)
   default UpdateLineageRow updateMarquezModel(LineageEvent event, ObjectMapper mapper) {
     UpdateLineageRow updateLineageRow = updateBaseMarquezModel(event, mapper);
     RunState runState = getRunState(event.getEventType());
-    if (event.getEventType() != null && runState.isDone()) {
+
+    boolean isStreaming =
+        Optional.ofNullable(event.getJob())
```
wslulciuc, line 173: Since this check exists in multiple location in code, can we add:

```java
Job.isStreaming()
```
or
```java
Utils.isStreamingJob(Job)
```

api/src/main/java/marquez/service/models/Job.java
```diff
@@ -41,6 +41,9 @@ public final class Job {
   @Nullable @Setter private Run latestRun;
   @Getter private final ImmutableMap<String, Object> facets;
   @Nullable private UUID currentVersion;
+  @Getter @Nullable private String jobType;
```
wslulciuc, line 44: Is `jobType` and `processingType` type the same?

api/src/main/java/marquez/service/models/Job.java
```diff
@@ -41,6 +41,9 @@ public final class Job {
...
+  @Getter @Nullable private String processingType;
+  @Getter @Nullable private String integration;
```
wslulciuc, line 46: I would prefer we set `integration` as a label than part of the core job model. Can we hold off an exposing the `integration` type until we decide on the best way to label datasets (_redshift_, `postgres`, etc) and jobs (_spark_, _flink_, etc)?

api/src/main/java/marquez/service/models/Job.java
```diff
@@ -56,7 +59,10 @@ public Job(
       @Nullable final String description,
       @Nullable final Run latestRun,
       @Nullable final ImmutableMap<String, Object> facets,
-      @Nullable UUID currentVersion) {
+      @Nullable UUID currentVersion,
+      @Nullable final String jobType,
+      @Nullable final String processingType,
+      @Nullable final String integration) {
     this.id = id;
     this.type = type;
```
wslulciuc, line 67: Note, `type` here can ether be `BATCH` or `STREAMING` and can drop `jobType` (see [`common/models/JobType.java`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/common/models/JobType.java)).

api/src/test/resources/open_lineage/event_full.json
```diff
@@ -61,6 +61,13 @@
         },
         "query": "SELECT * FROM foo"
       },
+      "jobType": {
+        "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
+        "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet",
```
wslulciuc, line 66: If `MyCustomJobFacet` the correct facet type?

api/src/main/java/marquez/db/JobVersionDao.java
wslulciuc: 🙌

pawel-big-lebowski: Sure, I refactored this already.

api/src/main/java/marquez/service/models/Job.java
wslulciuc: Since `Job.Type` is already defined, having both `Job.Type` and `Job.JobType` would be very confusing (specifically on usage). I don't think Marquez needs to be one-to-one in naming with OpenLineage (though I think we should certainly try).

First, I would map `processingType` to the [`JobType`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/common/models/JobType.java) defined by Marquez. And, I _think_ we can view `Job.JobType` as a sub type of `Job.Type`. That is, a `Job.Type` of `BATCH` can be a `DAG` or `TASK`. I also view `integration` as a qualifier of the `Job`. So, I'd favor having both the `JobType` and `integration` defined by OpenLineage as `labels`  as mentioned in my earlier [comment](https://github.com/MarquezProject/marquez/pull/2682#discussion_r1408311828)

api/src/main/java/marquez/service/models/Job.java
wslulciuc: Yeah, let's define labels in this PR, but maybe just add a `labels` column in `jobs` (would also make sense for `datasets` in the future) and define it as string array?

api/src/main/java/marquez/service/models/Job.java
wslulciuc: As [mentioned](https://github.com/MarquezProject/marquez/pull/2682#discussion_r1415230257) here, it's totally OK to remap OpenLineage properties to the Marquez specific model. I'd favor mapping `processingType` to the `JobType` defined in Marquez and document it in the API / docs.

api/src/main/java/marquez/common/Utils.java
```diff
@@ -313,6 +316,21 @@ public static Version newDatasetVersionFor(
     return newDatasetVersionFor(data);
   }
 
+  /**
+   * Verifies if a job is a streaming job.
+   *
+   * @param job
+   * @return
+   */
+  public static boolean isStreamingJob(Job job) {
```
wslulciuc, line 325: _Minor_: Actually we can go one step further and define this method within the [`LineageEvent.Job`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/service/models/LineageEvent.java#L204) class:

```
job.isStreaming()
```

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -745,16 +750,58 @@ default Set<DatasetId> toDatasetId(List<Dataset> datasets) {
...
+  /**
+   * A separate method is used as the logic to update Marquez model differs for streaming and batch.
+   * The assumption for batch is that the job version is created when task is done and cumulative
+   * list of input and output datasets from all the events is used to compute the job version UUID.
+   * However, this wouldn't make sense for streaming jobs, which are mostly long living and produce
+   * output before completing.
+   *
+   * <p>In this case, a job version is created based on the list of input and output datasets
+   * referenced by this job. If a job starts with inputs:{A,B} and outputs:{C}, new job version is
+   * created immediately at job start. If a following event produces inputs:{A}, outputs:{C}, then
```
wslulciuc, line 777: If a job has `{A,B}` as initial inputs, the **only** `{A}` shouldn't that be a new job version?

api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -773,7 +820,7 @@ default String formatNamespaceName(String namespace) {
   }
 
   default JobType getJobType(Job job) {
-    return JobType.BATCH;
+    return Utils.isStreamingJob(job) ? JobType.STREAM : JobType.BATCH;
```
wslulciuc, line 823: _Minor_: We can go one step further and define this method within the context of [`LineageEvent.Job`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/service/models/LineageEvent.java#L204) class:

```bash
# Returns the Marquez job type.
job.type() 
```

api/src/main/java/marquez/service/OpenLineageService.java
```diff
@@ -142,7 +142,16 @@ public CompletableFuture<Void> createAsync(LineageEvent event) {
             .thenAccept(
                 (update) -> {
                   if (event.getEventType() != null) {
-                    if (event.getEventType().equalsIgnoreCase("COMPLETE")) {
+                    boolean isStreaming =
+                        Optional.ofNullable(event.getJob())
```
wslulciuc, line 146: We can use the following method as described [here](https://github.com/MarquezProject/marquez/pull/2682/files#r1424584879):

```
event.getJob().isStreaming()
```

api/src/test/java/marquez/service/LineageServiceTest.java
```diff
@@ -425,6 +427,59 @@ public void testLineageWithWithCycle() {
...
+    Lineage lineageFromOutput =
+        lineageService.lineage(
+            NodeId.of(
+                new DatasetId(new NamespaceName(NAMESPACE), new DatasetName("output-dataset"))),
+            5,
+            true);
+
+    assertThat(lineageFromInput.getGraph()).hasSize(5); // 2 datasets + 3 jobs
+    assertThat(lineageFromInput.getGraph()).isEqualTo(lineageFromOutput.getGraph());
+  }
```
wslulciuc, line 481: Amazing test ❤️

api/src/test/java/marquez/service/OpenLineageServiceIntegrationTest.java
```diff
@@ -548,6 +554,172 @@ void testJobEvent() throws ExecutionException, InterruptedException {
...
+                .inputs(Collections.singletonList(input))
+                .outputs(Collections.singletonList(output))
+                .build())
+        .get();
+    Optional<Dataset> datasetRow = datasetDao.findDatasetByName(NAMESPACE, DATASET_NAME);
+
+    // (3) Assert that output is present and has dataset_version written
+    assertThat(datasetRow).isPresent().flatMap(Dataset::getCurrentVersion).isPresent();
+
+    // (4) Assert that job is present although job_version entry is not
```
wslulciuc, line 608: I _think_ a job version for the streaming job should exist based on what's described in the PR?

> It makes sense in this case, to create job version and identify input/output datasets at the beginning of the job and modify it on the fly if a job changes.

wslulciuc (approved):
We finally have support for streaming jobs! 💯 💯 💯

api/src/main/java/marquez/db/OpenLineageDao.java
pawel-big-lebowski: Existing algorithm to compute version of the job relies on all the inputs & datasets for the particular run. We should not modify it, as this would cause new job version for all the existing jobs, but we could create a separate version to evaluate version of the streaming job if we wanted. 

However, when looking into the approach, I found it useful. It's a cumulative approach, where new job version is created if a new input/output dataset is involved in processing. If some dataset was included in the past events, but is no longer present, the version does not get change. 

The benefit of this assumption is that we don't require producer to emit all the datasets all the time. If you emit amount of bytes written into output dataset, without containing input dataset in the event, it doesn't mean there is new job version without the inputs. 

Is it OK?

wslulciuc: I do feel if a stream is removed, you'll want to remove that edge from the _current_ lineage graph. But, I also understand the limitations here, as you mentioned, with bytes written to an _output_ dataset present with no _input_ datasets specified; that said, and after thinking it through, this may be a noop and the logic you have is reasonable.

For example, let's say we have a streaming job `X` with inputs `{A,B}` and outputs `C`. The job runs and is marked as `RUNNING` with the run ID `74f117af-eb90-4ffd-98e1-c1bc215934df`. To change the inputs from `{A,B}` to `{B}` (or similarly for the outputs), the user will have to redeploy the job (with new code possibly) and therefore be associated with a new run ID. So, what I think you have is logically correct given how streaming jobs are deployed. For batch jobs, versioning is more straight forward as we version from run-to-run.

Issue #2687, opened 2023-11-24
possibility to filter Namespaces


yanlibert:
We are implementing Marquez at scale, and we often end up with the creation of a lot of namespaces containing neither datasets nor jobs. 
This can be caused by the handling of symlinks (see [this issue](https://github.com/MarquezProject/marquez/issues/2645)) or an issue with the open lineage integrations. 
We would like to add the possibilty to filter the list of namespaces returned by the API by providing a regex in the `marquez.yml` config file
Exemple 
```
namespaceFilter: "foo*"
```
A PR is on the way to propose this feature

Issue #2645, opened 2023-10-10
[BUG] Symlink facet create an empty namespace with 0 dataset in it


dkt-sophie-ly:
Hi all ! 

According to our conversation on slack i'm opening this issue.

<img width="583" alt="image" src="https://github.com/MarquezProject/marquez/assets/130998593/9503e6f5-07f0-4024-a68d-272223b81188">

When a symlink is specified an empty namespace is created with no dataset in it.

Way to reproduce the issue:

I'm sending this event to marquez:
```
{
  "job": {
    "name": "symlink_job_0",
    "facets": {},
    "namespace": "test_namespace"
  },
  "run": {
    "runId": "3c50b3d8-6776-11ee-9331-00163e015999",
    "facets": {}
  },
  "inputs": [
    {
      "name": "dataset_0",
      "facets": {
        **"symlinks": {
          "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.1.0/client/python",
          "_schemaURL": "https://openlineage.io/spec/facets/1-0-0/SymlinksDatasetFacet.json",
          "identifiers": [
            {
              "name": "symlink_prefix",
              "type": "DB_TABLE",
              "namespace": "s3://symlynk_test"
            }
          ]
        }**,
        "SchemaDatasetFacet": {...}
      },
      "namespace": "test_namespace"
    }
  ],
  "outputs": [...],
  "producer": "https://github.com/OpenLineage/OpenLineage/tree/0.17.0/client/python",
  "eventTime": "2023-10-10T14:06:41.079514Z",
  "eventType": "START",
  "schemaURL": "https://openlineage.io/spec/1-0-5/OpenLineage.json#/definitions/RunEvent"
}
```

Current behavior:

It creates an empty namespace named 's3://symlynk_test' with no dataset in it.

![image](https://github.com/MarquezProject/marquez/assets/130998593/9b95a6a0-404e-4f75-a835-f501d701d121)

Expected behavior:

Not sure how it should be done but in our example dataset `s3://symlynk_test.symlink_prefix` and `test_namespace.dataset_0` must be the same dataset so they should have the same lineage.

wslulciuc:
Thanks @dkt-sophie-ly for opening this issue! I've added the bug to our [roadmap](https://github.com/orgs/MarquezProject/projects/1/views/4) (but haven't yet tagged a fix to a possible release).

PR #2688, opened 2023-11-24
Add namespace filter option in config file

yanlibert:
### Problem
We are implementing Marquez at scale, and we often end up with the creation of a lot of namespaces containing neither datasets nor jobs.
This can be caused by the handling of symlinks (see https://github.com/MarquezProject/marquez/issues/2645) or an issue with the open lineage integrations.
We would like to add the possibilty to filter the list of namespaces returned by the API by providing a regex in the marquez.yml config file

Closes: #2687

### Solution
This solution provides an optional `namespaceFilter` parameter to be used in the `marquez.yml` config file
Example: 
```
server:
  applicationConnectors:
  - type: http
    port: ${MARQUEZ_PORT:-5000}
    httpCompliance: RFC7230_LEGACY
  adminConnectors:
  - type: http
    port: ${MARQUEZ_ADMIN_PORT:-5001}

db:
  driverClass: org.postgresql.Driver
  url: jdbc:postgresql://localhost:5432/marquez
  user: marquez
  password: marquez

migrateOnStartup: true

graphql:
  enabled: true

logging:
  level: INFO
  appenders:
    - type: console

tags:
  - name: PII
    description: Personally identifiable information
  - name: SENSITIVE
    description: Contains sensitive information

namespaceFilter: "foo*"
```

One-line summary:

### Checklist


wslulciuc:
@yanlibert, thanks for extending our `marquez.yml` to support filtering metadata based off a pattern! We'll also want to have a fix for https://github.com/MarquezProject/marquez/issues/2645 (the root cause), but think supporting a config-based why to exclude metadata make sense.

I'd suggest we define a `exclude` option (in place of `namespaceFilter`):
```
exclude:
  namespaces: [<pattern0>, <pattern1>,...]
```
which can then define the environment variable `EXCLUDE_NAMESPACES` to override patterns for namespaces:
```
exclude:
  namespaces: ${EXCLUDE_NAMESPACES}
```

Then, as a follow up PR, we can expand filtering to exclude datasets and jobs:
```
exclude:
  namespaces: [<pattern0>, <pattern1>,...]
  datasets: [<pattern0>, <pattern1>,...]
  jobs: [<pattern0>, <pattern1>,...]
```

wslulciuc:
@yanlibert as a follow up to my [comment](https://github.com/MarquezProject/marquez/pull/2688#issuecomment-1831024264) above, we may also want to consider how we might avoid storing metadata as well (therefore, avoid polluting metadata all together). That is, if we provide an `exclude` option on read, it would also make sense to provide an `exclude` on write. So something like:

```yaml
exclude:
  namespaces:
    onRead: boolean
    onWrite: boolean
    patterns: [<pattern0>, <pattern1>,...]
```

You would be able to control filtering of read/write using `exclude.namespaces.onRead` and `exclude.namespaces.onWrite`.  I know this is now going beyond your initial work / proposal, but I think I'd be great to establish a simple configuration around exclusion that can be expanded. Thoughts?

yanlibert:
I think this is a great addition to the feature. In fact, we are already filtering some events before sending them to Marquez and it would be great if we could do that directly with Marquez. 
I'm going to have a try at your suggestion and propose a first iteration of a solution

yanlibert:
@wslulciuc I follow your [recommendations ](https://github.com/MarquezProject/marquez/pull/2688#issuecomment-1832572046) and refactored the way of getting the filter from the config file. I created a separate filter Class, so that it opens the way of a finer control for filtering. I also covered this new behavior with a test case. 
Let me know if it's ok, I'd be happy to improve on this if needed.

YLibert:
@wslulciuc I increased the test coverage, please let me know if this suits your requirements.

api/src/main/java/marquez/api/filter/exclusions/ExclusionsConfig.java
```diff
@@ -0,0 +1,15 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Getter;
+
+public class ExclusionsConfig {
+  @Getter @JsonProperty public NamespaceExclusion namespaces;
+
+  public static class NamespaceExclusion {
+    @Getter @JsonProperty public boolean onRead;
+    @Getter @JsonProperty public boolean onWrite;
+    @Getter @JsonProperty public String patterns;
```
pawel-big-lebowski, line 12: Is this a single pattern or a list of patterns separated in some way?

pawel-big-lebowski (commented):
Would love to have more clarification on the usage. For example: how to pass multiple filters

api/src/main/java/marquez/api/filter/exclusions/ExclusionsConfig.java
yanlibert: @pawel-big-lebowski yes, sorry about that, I forgot to include the proper documentation. I added a section to the FAQ to explain the usage, I'd be happy to move it to another place that is more suitable if needed

api/src/main/java/marquez/api/NamespaceResource.java
```diff
@@ -74,8 +75,16 @@ public Response get(@PathParam("namespace") NamespaceName name) {
   public Response list(
       @QueryParam("limit") @DefaultValue("100") @Min(value = 0) int limit,
       @QueryParam("offset") @DefaultValue("0") @Min(value = 0) int offset) {
-    final List<Namespace> namespaces = namespaceService.findAll(limit, offset);
-    return Response.ok(new Namespaces(namespaces)).build();
+    final String namespaceFilter = ExclusionsFilter.getNamespacesReadFilter();
```
wslulciuc, line 78: To clean up the code, I suggest:

1. Define **class** `Exclusions` with a cleaner interface defined in pkg `marquez.common` (see below) in place of `ExclusionsFilter`
2. Use lambdas to call `NamespaceService.findAll()` or `NamespaceService.findAllWith()`

**`Exclusions.java`**
```java
package marquez.common;

import com.google.common.collect.ClassToInstanceMap;
import com.google.common.collect.MutableClassToInstanceMap;
import javax.annotation.Nullable;
import lombok.NonNull;

public final class Exclusions {
  private Exclusions() {}

  private static final ClassToInstanceMap<Object> EXCLUSIONS = MutableClassToInstanceMap.create();
 

  public static void use(@NonNull ExclusionsConfig config) {
    // Build exclusions map using config.
    EXCLUSIONS.put(NamespaceExclusion.class, new NamespaceExclusion(...));
  }

  public static NamespaceExclusion namespaces() {
    return EXCLUSIONS.getInstance(NamespaceExclusion.class);
  }

  public record NamespaceExclusion(@Nullable String onRead, @Nullable String onWrite) {}
}
```

Then, you can chain your calls:

**`NamespaceResource.list()`**

```java
final List<Namespace> namespaces =
  Optional.ofNullable(Exclusions.namespaces().onRead())
    .map(pattern -> namespaceService.findAllWith(pattern, limit, offset))
    .orElseGet(() -> namespaceService.findAll(limit, offset));

return Response.ok(new Namespaces(namespaces)).build();
```

wslulciuc: @yanlibert I'm wondering, as we work through the `Exclusions` code, would it make sense to just have:

```
exclude:
  namespaces:
     enabled: boolean
     onRead: "<pattern0>|<pattern1>|..."
     onWrite: "<pattern0>|<pattern1>|..."
```

This would give greater control and flexibility over what to exclude 😅

yanlibert: I think it's a great suggestion, I refactored the code accordingly

api/src/main/java/marquez/db/NamespaceDao.java
```diff
@@ -78,6 +78,10 @@ default Namespace upsertNamespaceMeta(
   @SqlQuery("SELECT * FROM namespaces ORDER BY name LIMIT :limit OFFSET :offset")
   List<Namespace> findAll(int limit, int offset);
 
+  @SqlQuery(
+      "SELECT * FROM namespaces WHERE name !~ :excluded ORDER BY name LIMIT :limit OFFSET :offset")
+  List<Namespace> findAllFilter(String excluded, int limit, int offset);
```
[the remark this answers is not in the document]
yanlibert: Done

api/src/main/java/marquez/api/NamespaceResource.java
yanlibert: @wslulciuc I also think option 2 where `pattern` is a string of regex is simpler, and more straightforward to implement as I'm not a java dev 😄 
I increased the coverage and refactored the code as suggested by implementing a cleaner Exclusions interface.

wslulciuc (approved):
Nice! This looks great to me and really like how simple it is to define exclusions 💯 💯 💯

PR #2693, opened 2023-11-29
change url path encoding to match jersey decoding

davidjgoss:
### Problem

When an entity that is referred to in a path, like a Dataset or Job, has spaces in the name, the Java client will fail to retrieve it, with a 404 from Marquez.

This is because the Marquez API (via Jersey) seems to decode paths based on proper URI encoding, where spaces should be represented with `%20`, but the existing implementation used `URLEncoder` which only does HTML form encoding, where spaces are represented with `+`.

In other words...

- With a dataset named `My Dataset`
- The URL that will be matched by the API is `namespaces/{namespace}/datasets/My%20Dataset`
- But the URL that the Java client currently forms is `namespaces/{namespace}/datasets/My+Dataset`

...hence the 404.

It's worth noting that query parameters don't suffer from this, so e.g. the lineage and column-lineage endpoints are unaffected.

### Solution

Swap out the implementation of `MarquezPathV1::encode` to use `UrlEscapers` path segment escaper which does proper URI encoding.

This does change behaviour regarding colons - the new implementation does not replace `:` with `%3A` as the old one did. The API is tolerant of either so this shouldn't be an issue. And again query parameters are forgiving of all variations.

One-line summary: change url path encoding to match jersey decoding

### Checklist

### DIFF
--- a/clients/java/src/main/java/marquez/client/MarquezPathV1.java
+++ b/clients/java/src/main/java/marquez/client/MarquezPathV1.java
@@ -6,9 +6,7 @@
 package marquez.client;
 
 import com.google.common.annotations.VisibleForTesting;
-import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
+import com.google.common.net.UrlEscapers;
 import java.util.Arrays;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


wslulciuc (approved):
Nice! Looks good to me 💯

Issue #2676, opened 2023-11-01
Proposal: Version dataset schemas separately to datasets


davidjgoss:
## Problem

Many real-world jobs have these characteristics:

- Runs very frequently
- Output dataset schema(s) change _infrequently_

Because of Marquez's philosophy of versioning every output dataset on every run, this means a lot of data is generated that is for all intents and purposes duplicate, in the `dataset_versions_field_mapping` table.

Taking an oversimplified example, a dataset with a stable 20-column schema that is written to by a job that runs every 10 minutes for 30 days will result in 864,000 rows in `dataset_versions_field_mapping` when arguably there are only 20 meaningful rows of data. This is regardless of how many of those columns are actually touched by the job.

A dataset must still be versioned each time it occurs as an output, since it may contain facets that relate to the state of its structure and/or data at that moment in time beyond just the schema - including column-level lineage. But the schema itself doesn't need to be re-versioned unless it has actually changed.

## Solution

Introduce the concept of a `dataset_schema_version`, so that:

- When a dataset is first created (or, the next time it's versioned after we introduce this change), a `dataset_schema_version` is created which is referenced by the `dataset_version` in a new column.
- A new mapping table `dataset_schema_versions_field_mapping` links a `dataset_schema_version` to each `dataset_field`.
- When the dataset is next versioned, we compare the current version's schema to the incoming schema - if they are equal, then we re-use the same `dataset_schema_version` for this new `dataset_version` and don't create any new mappings.

Considerations:

- How do we determine if the schemas are "equal"? Certainly the same fields must be there (by name and type), and in the same order - what about description?
- As it stands, this won't change anything for `column_lineage` which suffers from a similar problem to `dataset_versions_field_mapping` in getting new rows every run even when the lineage doesn't change. We could look at this separately.
- Which read queries will we need to update for this? How can we keep it backwards compatible and (ideally) avoid copying data between tables? Would we eventually have a plan to retire the old table?

cc @wslulciuc @phixMe

PR #2696, opened 2023-12-05
add proposal for dataset schema versions

davidjgoss:
Official proposal for #2676.

### Checklist


davidjgoss:
@wslulciuc @pawel-big-lebowski I just pushed some updated to this based on our discussion yesterday:

- `version` column omitted from new table, with reasoning (see #2071)
- Detail added on how equality check and matching on existing schema versions will work - the idea of using a hash is borrowed from Pact which uses it to good effect
- Detail added on how the migration script would work, roughly

pawel-big-lebowski:
Superb proposal with A1 diagram 👍 

The proposal contains an example of job run every 10 mins resulting in 864,000 rows in `dataset_versions_field_mapping` . Would this result in 21600 dataset schema versions after the migration? Are you planning just to migrate the data or also clean redundant entries?

davidjgoss:
Added a section regarding the behaviour with input datasets, where the current dataset version is updated, and how we will handle this differently for dataset schema versions.

proposals/2676-version-dataset-schemas-separately.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Implementation
+
+### Schema changes
+
+- New table `dataset_schema_versions` added with columns:
+  - `created_at` (timestamp)
+  - `schema_hash` (content hash for equality checking)
+  - `dataset_uuid` (fk to `datasets`)
+  - `uuid` (pk)
```
wslulciuc, line 53: To avoid the column `schema_hash`, can we generate the "hash" as a `UUID` as we do, for example, job [versions](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/common/Utils.java) using `UUID.nameUUIDFromBytes(bytes)`?

proposals/2676-version-dataset-schemas-separately.md
```diff
@@ -0,0 +1,170 @@
...
+
+Once done, this will mean we are dealing only with the new tables for both writes and reads.
+
+### Phase 3 (cleanup)
+
+- Drop `dataset_versions_field_mapping` table
+
+## Next Steps
+
+Pull request for phase 1?
```
wslulciuc, line 165: Yes 😄

proposals/2676-version-dataset-schemas-separately.md
wslulciuc: Oops, I didn't finish my thought. If we hash to a `UUID`, we can use that value as the `uuid` for the schema version in place of `schema_hash` (that we'd no longer need).

davidjgoss: Ah okay, I'll take a look at this an update it!

davidjgoss: Updated now.

wslulciuc: Thanks!

wslulciuc (approved):
🚀 🚀 💯

Issue #2692, opened 2023-11-29
Adding a DELETE end point for dataset level tags


davidsharp7:
Current tags are added to a dataset via the end point

```
POST http://localhost:5000/api/v1/namespaces/{namespace}/datasets/{dataset}/tags/{tag}
```

In the background this updates 

datasets_tag_mapping

with the dataset_uuid and the tag_uuid.  The UI then returns any tags attached to a given dataset

It would be useful to have a DELETE end point to remove that linkage.  Simplest approach would be to just delete the linkage with the following query.

so 
```
DELETE http://localhost:5000/api/v1/namespaces/{namespace}/datasets/{dataset}/tags/{tag}
```

would run 

```        
DELETE FROM datasets_tag_mapping dtm
        WHERE EXISTS (
            SELECT 1
            FROM datasets d
            JOIN tags t ON d.uuid = dtm.dataset_uuid AND t.uuid = dtm.tag_uuid
            WHERE d.name = :datasetName AND t.name = :tagName
        );
```

wslulciuc:
@davidsharp7, I added the API for tag removal to our [roadmap](https://github.com/orgs/MarquezProject/projects/1/views/4). Let us know if you need any help our guidance. Thanks!

PR #2698, opened 2023-12-09
add DELETE end point for dataset tags

davidsharp7:
### Problem

Currently we have no way via the API to a delete a tag from a dataset meaning datasets could potentially be "mis tagged"

Closes: #2692

### Solution

Solution is to create a DELETE end point which removes the linkage between the dataset and the tag in

```
datasets_tag_mapping
```

One-line summary:

Add DELETE end point for dataset tags.

### DIFF
--- a/api/src/main/java/marquez/api/DatasetResource.java
+++ b/api/src/main/java/marquez/api/DatasetResource.java
@@ -216,7 +216,8 @@ public class DatasetResource extends BaseResource {
         tagName.getValue(),
         datasetName.getValue(),
         namespaceName.getValue());
-    datasetService.deleteDatasetTag(datasetName.getValue(), tagName.getValue());
+    datasetService.deleteDatasetTag(
+        namespaceName.getValue(), datasetName.getValue(), tagName.getValue());
     Dataset dataset =
         datasetService
             .findDatasetByName(namespaceName.getValue(), datasetName.getValue())
--- a/api/src/main/java/marquez/api/DatasetResource.java
+++ b/api/src/main/java/marquez/api/DatasetResource.java
@@ -201,6 +201,28 @@ public class DatasetResource extends BaseResource {
   @Timed
   @ResponseMetered
   @ExceptionMetered
+  @DELETE
+  @Path("/{dataset}/tags/{tag}")
+  @Produces(APPLICATION_JSON)
+  public Response deleteDatasetTag(
+      @PathParam("namespace") NamespaceName namespaceName,
+      @PathParam("dataset") DatasetName datasetName,
+      @PathParam("tag") TagName tagName) {
+    throwIfNotExists(namespaceName);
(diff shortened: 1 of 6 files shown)


api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -272,6 +272,18 @@ DatasetRow upsert(
...
   Optional<DatasetRow> delete(String namespaceName, String name);
 
+  @SqlUpdate(
+      """
+        DELETE FROM datasets_tag_mapping dtm
+        WHERE EXISTS (
+            SELECT 1
+            FROM datasets d
+            JOIN tags t ON d.uuid = dtm.dataset_uuid AND t.uuid = dtm.tag_uuid
+            WHERE d.name = :datasetName AND t.name = :tagName
```
pawel-big-lebowski, line 282: Shouldn't where condition contain namespace filter as well?

pawel-big-lebowski (commented):
I really like @wslulciuc approach to inregration tests introduced here -> https://github.com/MarquezProject/marquez/pull/2700

I know the testing approach is newer than this PR, but the code here seems to be the perfect candidate to write a neat integration test. This will also require implenting tag deleting feature in Marquez client.

api/src/main/java/marquez/api/DatasetResource.java
```diff
@@ -198,6 +198,32 @@ public Response tag(
     return Response.ok(dataset).build();
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @DELETE
+  @Path("/{dataset}/tags/{tag}")
+  @Produces(APPLICATION_JSON)
+  public Response deleteDatasetTag(
```
wslulciuc, line 207: We'l also want to support deleting a tag from a dataset field via `DatasetResource.deleteDatasetFieldTag()`

davidsharp7: Yeah I was going to do this as a separate PR as there were a couple of differences due to facets which I wanted to understand more.

api/src/main/java/marquez/db/DatasetDao.java
davidsharp7: Not sure it would add much given the data model of 

datasets -> datasets_tag_mapping <- tags

Suspect when the tag linkage is created in the first place that the dataset_uuid is added based on the namespace provided so not sure it adds much in this case but happy to put it in.

Totally agree on the integration tests though @wslulciuc has been really helpful with that.

api/src/main/java/marquez/api/DatasetResource.java
wslulciuc: Ahh, right deleting tags for dataset fields is a bit more involved and makes sense to do as a follow up.

PR #2713, opened 2023-12-20
Fix `py` client version

wslulciuc:


merobi-hub (approved):
Thank you, @wslulciuc

PR #2736, opened 2024-01-23
fix symlink display on marquez

sophiely:
### Problem

Sending an event with a dataset symlink create an empty namespace with 0 dataset in it.

For example, this event:

````
{
  ...
  },
  "inputs": [
    {
      "namespace": "test_namespace"
      "name": "dataset_0",
      "facets": {
        "symlinks": {
          "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.1.0/client/python",
          "_schemaURL": "https://openlineage.io/spec/facets/1-0-0/SymlinksDatasetFacet.json",
          "identifiers": [
            {
              "name": "symlink_prefix",
              "type": "DB_TABLE",
              "namespace": "symlink_test"
            }
          ]
        },
      },
    }
  ],
  ...
}
````

create an empty namespace called symlink_test

Closes: [#2645](https://github.com/MarquezProject/marquez/issues/2645)

### Solution

- Display the dataset create by a symlink facet in their own namespace (here in the example the namespace symlink_test will contain a dataset called symlink_prefix)
- The lineage of this dataset will be redirect to the lineage of the "main" dataset (the dataset test_namespace.dataset_0 in our example)

Example:

If we run these 2 runs:

new_dataset_a {facet: new_dataset_sym_a} ------- new_symlink_job_a ----------> new_dataset_b

then:
new_dataset_sym_a ------- new_symlink_job_b --------------> new_dataset_sym_b

On marquez we'll have

image

image

Please find more detailed explanation on the comments below.

This fix include another issue on the front though, the version/facets of the selected dataset are not directly displayed.

Since the selected dataset is the symlink and not the primary dataset, the front doesn't recognize the selected dataset as part of the lineage as a result the version endpoint is not run.

image

But if we click on the dataset new_dataset_a additional, the dataset version query is run and information are displayed

image

One-line summary:

### Checklist

- [ ] Your changes are accompanied by tests (_if relevant_) **TODO**
- [x] You've included a one-line summary of your change for the [`CHANGELOG.md`] (https://github.com/MarquezProject/marquez/blob/main/CHANGELOG.md#unreleased) (_Depending on the change, this may not be necessary_).

### CODE REVIEW

File: api/src/main/resources/marquez/db/migration/V68__alter_datasets_view_to_keep_symlinks.sql [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+    CASE
+    	 WHEN (d.namespace_name = namespaces.name AND d.name = symlinks.name) THEN d.namespace_name
+    	 ELSE namespaces.name
+    END
+    AS namespace_name,
+    d.source_name,
+    d.is_deleted
+   FROM datasets d
+     JOIN dataset_symlinks symlinks ON d.uuid = symlinks.dataset_uuid
+     JOIN namespaces ON symlinks.namespace_uuid = namespaces.uuid
```
dkt-sophie-ly, line 32: The query is not much different from before, but now the dataset uuid is not the primary key anymore since a dataset and his symlink has the same dataset uuid (that's why the group by is not here anymore).
We join this view with dataset_symlinks to identify if it's a primary dataset or not. if it is a primary dataset, the values in the row remain the same as before. If not, the namespace, name and namespace uuid are replaced by the one from the symlinks (value from the join table).

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -102,16 +102,27 @@ INNER JOIN jobs_view j ON (j.uuid=l2.job_uuid OR j.uuid=l2.job_symlink_target_uu
       """
       SELECT ds.*, dv.fields, dv.lifecycle_state
       FROM datasets_view ds
-      LEFT JOIN dataset_versions dv on dv.uuid = ds.current_version_uuid
-      WHERE ds.uuid IN (<dsUuids>)""")
+      LEFT JOIN dataset_versions dv ON dv.uuid = ds.current_version_uuid
+      LEFT JOIN dataset_symlinks dsym ON dsym.namespace_uuid = ds.namespace_uuid and dsym.name = ds.name
+      WHERE dsym.is_primary = true
+      AND ds.uuid IN (<dsUuids>)""")
```
dkt-sophie-ly, line 108: So here since the view datasets_views can have several rows with the same uuid we choose the one flagged as primary.

api/src/main/java/marquez/db/LineageDao.java
```diff
@@ -102,16 +102,27 @@ INNER JOIN jobs_view j ON (j.uuid=l2.job_uuid OR j.uuid=l2.job_symlink_target_uu
...
+      LEFT JOIN dataset_symlinks dsym ON dsym.namespace_uuid = ds.namespace_uuid and dsym.name = ds.name
+      INNER JOIN (
+        SELECT uuid
+        FROM datasets_view as u
+        WHERE
+            u.name = :datasetName
+            AND u.namespace_name = :namespaceName
+        ) as u
+      	on u.uuid = ds.uuid
+      WHERE dsym.is_primary is true""")
```
dkt-sophie-ly, line 125: So here since the view datasets_views can have several rows with the same uuid we choose the one flagged as primary.

api/src/main/java/marquez/service/LineageService.java
```diff
@@ -114,15 +114,20 @@ public Lineage lineage(NodeId nodeId, int depth, boolean withRunFacets) {
...
+          this.getDatasetData(datasetId.getNamespace().getValue(), datasetId.getName().getValue());
+
+      if (!datasetIds.contains(datasetData.getUuid())) {
+        log.warn(
+            "Found jobs {} which no longer share lineage with dataset '{}' - discarding",
+            jobData.stream().map(JobData::getId).toList(),
+            nodeId.getValue());
+        return toLineageWithOrphanDataset(nodeId.asDatasetId());
+      }
+    }
```
dkt-sophie-ly, line 130: Now we check if the uuid of the node and not the namespace+name

pawel-big-lebowski (commented):
Would it be OK to add some test to `LineageService` which is failing prior to the code change introduced and is passing afterwards?

api/src/main/java/marquez/service/LineageService.java
wslulciuc: Nice! Thanks for adding the `warn` log 💯

api/src/main/resources/marquez/db/migration/V68__alter_datasets_view_to_keep_symlinks.sql
wslulciuc: Valid, since we recreate the `dataset_view` on every marquez deploy, we can make your changes in [`R__3_Datasets_view.sql`](https://github.com/MarquezProject/marquez/blob/main/api/src/main/resources/marquez/db/migration/R__3_Datasets_view.sql) as @pawel-big-lebowski suggested.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## [Unreleased](https://github.com/MarquezProject/marquez/compare/0.44.0...HEAD)
 
+### Fixed
```
wslulciuc, line 5: Thanks for updating the changelog 💯

dkt-sophie-ly:
Hi @pawel-big-lebowski @wslulciuc 

Just updated the code accordingly yo your comments
- Modify `R__3_Datasets_view.sql`
- Use ` INNER JOIN datasets_view AS d ON d.uuid = df.dataset_uuid
 WHERE CAST((:namespaceName, :datasetName) AS DATASET_NAME) = ANY(d.dataset_symlinks)` as a filter
- Add a test in `LineageService` 
Concerning this test I feel like the way I add a symlinks facets is not working (the dataset is not considered as a symlink). I may need a little help on that if that's ok with you :)

Thanks for your review !

sophiely:
Hi @wslulciuc @pawel-big-lebowski  :)
Did you have the time to check this PR ? 

Thanks for your review

wslulciuc (approved):
Amazing work and test, @sophiely 💯

PR #2725, opened 2024-01-02
Web: Updates for Table and Column Lineage 

phixMe:
### Problem

We currently have a feature gap for column lineage where we don't have a display for it. In order to correctly display this we needed to have something to handle nested graphs so that columns can be nested under tables. 

This PR both creates a new page for column lineage and an updated view for lineage with a common set of shared principles.

#### Feature Set
- Redesigned lineage page 
- New column lineage page
- Refresh button (lineage + column lineage)
- Compact mode for lineage display
- Zooming and framing capabilities and zoom controls
- Open view for graphs with a side panel instead of a draggable bar that limited graph navigation and visualization.
- Shared action menu for graph context and actions
- Tuning down the graph depth and preserving more in the url.

#### Technical 
- New client side rendering engine with out of the box support for arbitrarily nested graphs (subgraphs) and hot swappable rendering strategies. There is also an easy to use and type safe framework to construct node types
- New custom React based d3 manipulation framework for zooming, panning, transitions, framing and focal points that uses React paradigms instead of direct DOM access like a conventional d3 approach
- A Minimap that allows for context awareness in a larger graph along with auto positioning extents

https://github.com/MarquezProject/marquez/assets/7514204/0f3083fb-eed6-472b-9c57-5bd29011920d

image
image
image

One-line summary:
New views for both table and column lineage with shared new engine. Marks column lineage as `beta` for preview.

### Checklist

- [ ] You've [signed-off](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#sign-your-work) your work
- [ ] Your changes are accompanied by tests (_if relevant_)
- [ ] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [ ] You've included a one-line summary of your change for the [`CHANGELOG.md`](https://github.com/MarquezProject/marquez/blob/main/CHANGELOG.md#unreleased) (_Depending on the change, this may not be necessary_).
- [ ] You've versioned your `.sql` database schema migration according to [Flyway's naming convention](https://flywaydb.org/documentation/concepts/migrations#naming) (_if relevant_)
- [ ] You've included a [header](https://github.com/MarquezProject/marquez/blob/main/CONTRIBUTING.md#copyright--license) in any source code files (_if relevant_)

PR #2742, opened 2024-02-02
api: fill data in column lineage input nodes.

JDarDagran:
### Problem

As stated in #2725 data is null for output nodes in column lineage endpoint.

### Solution

Fill data field.

### Checklist

### DIFF
--- a/api/src/main/java/marquez/db/models/ColumnLineageNodeData.java
+++ b/api/src/main/java/marquez/db/models/ColumnLineageNodeData.java
@@ -42,6 +42,17 @@ public class ColumnLineageNodeData implements NodeData {
     this.inputFields = inputFields;
   }
 
+  public ColumnLineageNodeData(InputFieldNodeData data) {
+    this.namespace = data.namespace;
+    this.dataset = data.dataset;
+    this.datasetVersion = data.datasetVersion;
+    this.field = data.field;
+    this.fieldType = "";
+    this.transformationDescription = data.transformationDescription;
+    this.transformationType = data.transformationType;
+    this.inputFields = ImmutableList.of();
+  }
+
   /**
    * @deprecated Moved into {@link ColumnLineageInputField} to support multiple jobs writing to a
    *     single dataset. This method is scheduled to be removed in release {@code 0.30.0}.
--- a/api/src/main/java/marquez/service/ColumnLineageService.java
+++ b/api/src/main/java/marquez/service/ColumnLineageService.java
@@ -68,10 +68,14 @@ public class ColumnLineageService extends DelegatingDaos.DelegatingColumnLineage
               NodeId nodeId = toNodeId(columnLineageNodeData, includeVersion);
               graphNodes.put(nodeId, Node.datasetField().data(columnLineageNodeData).id(nodeId));
               columnLineageNodeData.getInputFields().stream()
-                  .map(i -> toNodeId(i, includeVersion))
                   .forEach(
-                      inputNodeId -> {
-                        graphNodes.putIfAbsent(inputNodeId, Node.datasetField().id(inputNodeId));
+                      inputNode -> {
+                        NodeId inputNodeId = toNodeId(inputNode, includeVersion);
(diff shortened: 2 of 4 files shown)


pawel-big-lebowski (approved):

pawel-big-lebowski (approved):

api/src/main/java/marquez/db/models/ColumnLineageNodeData.java
```diff
@@ -42,6 +42,17 @@ public ColumnLineageNodeData(
...
+  public ColumnLineageNodeData(InputFieldNodeData data) {
+    this.namespace = data.namespace;
+    this.dataset = data.dataset;
+    this.datasetVersion = data.datasetVersion;
+    this.field = data.field;
+    this.fieldType = "";
+    this.transformationDescription = data.transformationDescription;
+    this.transformationType = data.transformationType;
+    this.inputFields = ImmutableList.of();
+  }
```
phixMe, line 54: I'm a little confused by some of these fields like `fieldType` and `inputFields`. Would these ever be set here?

JDarDagran: The truth is that if I set `InputFieldNodeData` to implement `NodeData` interface jackson couldn't deduce what was type of `NodeData` was in the output node (the one that previously had `data` field null). For that reason I basically made a workaround to simply create `ColumnLineageNodeData` from `InputFieldNodeData`.

I'd need more guidance how to fix jackson deduction :|

tatiana:
@wslulciuc @phixMe - while talking to @JDarDagran, he mentioned that there were concerns this PR may not solve the original issue and that you'd be logging an issue with further details. Could you confirm if that is the case?

wslulciuc:
@tatiana: yeah, although this is a "fix", I think the issue is more fundemtail to our colum lineage query. That is, @phix and I looked into this a few days ago and walked through every OL call (@JDarDagran provided us) and it was enlightening. I’d like to share what we found, just haven’t had a window of time to write up our findings but the tl;dr is: the **output_dataset_field_uuid** in our [column lineage query](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/ColumnLineageDao.java#L169) doesn’t return values for the **ColumnLineageNode**s on the edges of the graph (ie. graph nodes with no input columns, see below). I’d like to further confirm this, but given that our seed data works just fine, there’s something we are not accounting for.

IMG_6440

pawel-big-lebowski:
> @tatiana: yeah, although this is a "fix", I think the issue is more fundemtail to our colum lineage query. That is, @phix and I looked into this a few days ago and walked through every OL call (@JDarDagran provided us) and it was enlightening. I’d like to share what we found, just haven’t had a window of time to write up our findings but the tl;dr is: the **output_dataset_field_uuid** in our [column lineage query](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/db/ColumnLineageDao.java#L169) doesn’t return values for the **ColumnLineageNode**s on the edges of the graph (ie. graph nodes with no input columns, see below). I’d like to further confirm this, but given that our seed data works just fine, there’s something we are not accounting for.
> 

Hi @wslulciuc, I spent some time investigating nulls and I think @JDarDagran's PR solves the problem we were able to reproduce. The PR modifies a test and then applies changes need to fix it. That's why I think it is worth merging from that perspective. 

What do you say? Once you provide more details on other issue, we can resume discussion while not blocking this.

wslulciuc:
@pawel-big-lebowski yeah, I'll merge 😉. I'm fine with that as we don't want to further delay the merge of https://github.com/MarquezProject/marquez/pull/2725. Mind opening an issue to dive deep into the column-level lineage query and move our discussion there?

wslulciuc (approved):
Thanks for the fix, @JDarDagran! We'll want to open an issue to dive deeper into the column-level lineage query and move our discussion there. But, giving this a +1 to not block https://github.com/MarquezProject/marquez/pull/2742

PR #2761, opened 2024-03-07
Update dataset/dataset field tagging experience

davidsharp7:
### Problem

The Adding and deleting of tags was not providing an intuitive and seemless experience.

### Solution

Based on the feedback i've changed how the datasets details/tags/info components work by introducing a switch which allows field level tags to be exposed.  Refresh only now occurs when the slide toggle is true.  You can add multiple tags in one go now as well.

Screenshot 2024-04-03 at 16 22 46

Screenshot 2024-04-03 at 16 23 03

Screenshot 2024-04-05 at 16 41 18

One-line summary:

Update dataset/dataset field tagging experience.

### Checklist


api/src/main/java/marquez/db/DatasetFieldDao.java
```diff
@@ -129,7 +129,7 @@ void deleteDatasetVersionFieldTag(
 
   @SqlUpdate(
       "INSERT INTO dataset_fields_tag_mapping (dataset_field_uuid, tag_uuid, tagged_at) "
-          + "VALUES (:rowUuid, :tagUuid, :taggedAt)")
```
phixMe, line 132: Can you add a Java language test for this?

web/src/store/reducers/datasets.ts
```diff
@@ -32,7 +32,7 @@ export type IDatasetsState = {
   totalCount: number
   init: boolean
   deletedDatasetName: string
-  refreshTags: boolean
+  refreshTags: string
```
phixMe, line 35: Why would this be a string and not a boolean?

davidsharp7: So the thinking is twofold - one to bring it into line with

https://github.com/MarquezProject/marquez/blob/78a191b3cb73a50dc7466103e0588e80c3374042/web/src/components/datasets/DatasetDetailPage.tsx#L94

where we are refreshing based on the change in dataset name.

Second reason was for trouble shooting purposes.  Based on the concatanation of namespace/datasetname/field/tag it was just easier to understand/diagnose what was happening.

api/src/main/java/marquez/db/DatasetFieldDao.java
davidsharp7: Cool will add an additional insert which should result in no change.

web/src/store/reducers/datasets.ts
davidsharp7: Can always change the dependency to 
```
datasets.refreshTags === true
```
and change the reducer to set refreshTags to False when its updating then True on success.

api/src/main/java/marquez/db/DatasetFieldDao.java
phixMe: Thanks

web/src/store/reducers/datasets.ts
phixMe: Ok, let's just go with what you have here.

web/src/components/datasets/DatasetTags.tsx
```diff
@@ -119,19 +93,37 @@ const DatasetTags: React.FC<IProps> = (props) => {
...
-      ? addDatasetFieldTag(namespace, datasetName, listTag, datasetField)
-      : addDatasetTag(namespace, datasetName, listTag)
+      datasetField
+        ? addDatasetFieldTag(namespace, datasetName, newTag, datasetField)
+        : addDatasetTag(namespace, datasetName, newTag)
+    }
   }
 
   const handleDelete = (deletedTag: string) => {
+    const index = selectedTags.indexOf(deletedTag)
```
phixMe, line 119: We can do this a little more functionally like:
```tsx
const newSelectedTags = selectedTags.filter((tag) => deletedTag !== tag);
setSelectedTags(newSelectedTags);
```

phixMe: Maybe let's make this change and I'll merge if that sounds good?

davidsharp7: Surely can.

davidsharp7: Could I also do this or does it make it less readable?

```
setSelectedTags(selectedTags.filter((tag) => deletedTag !== tag))
```

wslulciuc (approved):
🤩

api/src/test/java/marquez/api/TagResourceIntegrationTest.java
```diff
@@ -100,4 +100,38 @@ public void testApp_testDatasetTagFieldDelete() {
     // assert that only SENSITIVE remains
     assertThat(taggedDatasetFieldDelete.getFields().get(0).getTags()).containsExactly("SENSITIVE");
   }
+
+  @Test
+  public void testApp_testDatasetTagFieldConflict() {
```
wslulciuc, line 105: ❤️

PR #2763, opened 2024-03-09
feat: add dataset schema versions

davidjgoss:
### Problem

This PR is the first step towards implementing the proposal from https://github.com/MarquezProject/marquez/blob/main/proposals/2676-version-dataset-schemas-separately.md.

### Solution

The idea is that, without changing what already gets written to the database, we start writing to the new `dataset_schema_versions` and `dataset_schema_versions_field_mapping` tables.

#### Todo

- [x] Create new database tables
- [x] Upsert dataset schema version when creating a new dataset version
  - [x] OpenLineage code path
  - [x] Legacy `DatasetDao` code path
  - [x] ~Legacy `RunDao` code path~ not viable to tackle in this PR
- [x] Handle in retention cleanup
  - [x] Ensure schema versions and fields get swept up in the cascade when a dataset is deleted
- [x] ~Handle input dataset where schema doesn't match current version~ for now will not change schema version even when input dataset schema has drifted - to be discussed in #2764

One-line summary: Add dataset schema versions to model and start writing to it

### Checklist


api/src/main/java/marquez/db/OpenLineageDao.java
```diff
@@ -913,7 +913,20 @@ default DatasetRecord upsertLineageDataset(
         Optional.ofNullable(ds.getFacets())
             .map(DatasetFacets::getSchema)
             .map(SchemaDatasetFacet::getFields)
-            .orElse(null);
```
wslulciuc, line 916: `null` vs an _empty_ field set may have different meanings. That is, `null` may mean no schema was provided, while an empty set can be ambiguous:
* `null`, but default empty set was used
* or the schema was an empty set

In. Marquez, we do want to make that distinction here and align with OpenLineage: the schema facet is present if a schema was extracted, or `null` if not present.

api/src/main/java/marquez/db/models/ExtendedDatasetVersionRow.java
```diff
@@ -25,12 +25,21 @@ public ExtendedDatasetVersionRow(
       @NonNull Instant createdAt,
       @NonNull UUID datasetUuid,
       @NonNull UUID version,
+      UUID schemaVersionUuid,
```
wslulciuc, line 28: Minor: The `schemaVersionUuid ` can be `null`. I would use `@Nullable ` to communicate that.

wslulciuc (approved):
💯 💯 💯

PR #2834, opened 2024-06-07
OpenSearch indexing and queries

phixMe:
### Problem
Our search right now does not enable nested queries on OpenLineage facets, code, linked entities, and ids. We want to 
enable our search to be the very best place to absorb and catalog OpenLineage based data.

https://github.com/user-attachments/assets/7a56fd5d-7f5d-47fa-a7b8-38a43e167761

### Includes
- Indexing of documents on OL events
- New APIs to retrieve search for jobs and datasets
- Search UI updates that can be enabled or fallback to legacy search system.
- Local and Helm updates for new OpenSearch containers needed for this. (Including dashboard)

### To Follow Up
- Update jobs and datasets for all tag related tag actions to include in indexes.
- Adding support for more facets and integrations (Trino, Flink...)

### Checklist


api/src/main/java/marquez/MarquezApp.java
```diff
@@ -5,8 +5,13 @@
 
 package marquez;
 
+import co.elastic.clients.elasticsearch.ElasticsearchClient;
```
pawel-big-lebowski, line 8: ElasticSearch derived from open-source project which currently forked into OpenSearch to remain community driven. Is it possible to bring some abstraction layer here to allow other implementations and vendors? For AWS users, https://aws.amazon.com/opensearch-service/ may seem a more natural case to use.

build.gradle
```diff
@@ -64,6 +66,8 @@ subprojects {
 
     dependencies {
         implementation "org.projectlombok:lombok:${lombokVersion}"
+        implementation "co.elastic.clients:elasticsearch-java:${elasticsearchVersion}"
```
pawel-big-lebowski, line 69: Do we want to include this within Marquez jar or let users add the jar manually in case they turn on the feature? The benefit of the latter approach is that it's easier to switch to another implementation if needed.

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -87,6 +89,96 @@ public void create(@Valid @NotNull BaseEvent event, @Suspended final AsyncRespon
...
+        indexDatasets(event.getInputs(), runUuid, event);
+      }
+      if (event.getOutputs() != null) {
+        indexDatasets(event.getOutputs(), runUuid, event);
+      }
+      indexJob(runUuid, event);
+    }
+  }
+
+  private Map<String, Object> buildJobIndexRequest(UUID runUuid, LineageEvent event) {
```
pawel-big-lebowski, line 117: How about creating separate class for this?

pawel-big-lebowski (commented):
I saw the demo of the feature and this is cracking. Amazing work, great feature, fab capabilities. 

My concerns relate to the vendor chosen. I think we should either add the ability to switch the search backend vender or stick to OpenSearch which Apache license based -> https://github.com/opensearch-project/OpenSearch/blob/main/LICENSE.txt This should be a simple switch at this stage.

api/src/main/java/marquez/MarquezApp.java
phixMe: I will look into making this swap out.

api/src/main/java/marquez/api/OpenLineageResource.java
phixMe: Yeah, there is some work to be done reorganizing this code and also creating an independent service to manage the queries as well.

JDarDagran:
Just a suggestion as I did not explore this too deeply. Wouldn't PSQL's [text search](https://www.postgresql.org/docs/14/textsearch.html) be enough? I've seen some comparisons, e.g. [here](https://supabase.com/blog/postgres-full-text-search-vs-the-rest) and it seems to be pretty viable option.

api/src/main/java/marquez/MarquezApp.java
```diff
@@ -154,6 +171,35 @@ public void run(@NonNull MarquezConfig config, @NonNull Environment env) {
     Exclusions.use(exclusions);
   }
 
+  private OpenSearchClient newOpenSearchClient(SearchConfig searchConfig) {
```
wslulciuc, line 174: We should encapsulate the `OpenSearchClient` instantiation in the constructor of `SearchService` by passing `SearchConfig`:

```
new SearchService(SearchConfig)
```

api/src/main/java/marquez/MarquezApp.java
```diff
@@ -154,6 +171,35 @@ public void run(@NonNull MarquezConfig config, @NonNull Environment env) {
...
+        new AuthScope(host),
+        new UsernamePasswordCredentials(searchConfig.getUsername(), searchConfig.getPassword()));
+    final RestClient restClient =
+        RestClient.builder(host)
+            .setHttpClientConfigCallback(
+                httpClientBuilder ->
+                    httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider))
+            .build();
+
+    JacksonJsonpMapper jsonpMapper = new JacksonJsonpMapper();
```
wslulciuc, line 188: We may also want to consider `PoolingAsyncClientConnectionManager` support as outlined in the OpenSearch [docs](https://opensearch.org/docs/latest/clients/java/) for the `jvm` client. Not a requirement for this PR as this feature will be in `beta` and load testing will help us decide on whether pooling is needed.

api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -68,6 +68,7 @@ public OpenLineageResource(
   public void create(@Valid @NotNull BaseEvent event, @Suspended final AsyncResponse asyncResponse)
       throws JsonProcessingException, SQLException {
     if (event instanceof LineageEvent) {
+      serviceFactory.getSearchService().indexEvent((LineageEvent) event);
```
wslulciuc, line 71: We'' want to wrap the index call in a `try/catch`.

api/src/main/java/marquez/MarquezApp.java
phixMe: 💡 Great idea. This would move the init code out of this non relevant final.

api/src/main/java/marquez/api/OpenLineageResource.java
wslulciuc: A log for now is good enough 👍

api/src/main/java/marquez/api/SearchResource.java
```diff
@@ -41,9 +49,12 @@ public class SearchResource {
   private static final String DEFAULT_LIMIT = "10";
   private static final int MIN_LIMIT = 0;
 
+  private final SearchService searchService;
```
wslulciuc, line 52: Given that the new search API is experimantl, do we want to prefix the `job` and `dataset` endpoints with `v2beta/search`?

api/src/main/java/marquez/api/SearchResource.java
```diff
@@ -72,6 +83,48 @@ public Response search(
     return Response.ok(new SearchResults(searchResults)).build();
   }
 
+  @Timed
+  @ResponseMetered
+  @ExceptionMetered
+  @GET
+  @Produces(APPLICATION_JSON)
+  @Path("/jobs")
+  public Response searchJobs(@QueryParam("q") @NotBlank String query) throws IOException {
```
wslulciuc, line 92: We'll want to expand the query (`q`) to also allow for qualifiers. For example, to search for metadata within a given namespace, the user can provide the qualifier `namespace:<NAMESPACE>`:

```
q=etl namespace:my-namespace
```

api/src/main/java/marquez/api/SearchResource.java
```diff
@@ -72,6 +83,48 @@ public Response search(
...
+        response.hits().hits().stream().map(Hit::source).collect(Collectors.toList());
+    List<Map<String, List<String>>> highlights =
+        response.hits().hits().stream().map(Hit::highlight).collect(Collectors.toList());
+
+    return Response.ok(new EsResult(hits, highlights)).build();
+  }
+
+  @ToString
+  public static final class EsResult {
+    @Getter private final List<ObjectNode> hits;
```
wslulciuc, line 117: For now, this payload would work, but we would need to generalize search results:

```
{
  "results": [Dataset,Job]
}
```

api/src/main/java/marquez/service/SearchService.java
```diff
@@ -0,0 +1,237 @@
...
+      log.warn("Search not configured");
+    }
+  }
+
+  public OpenSearchClient getClient() {
+    return this.openSearchClient;
+  }
+
+  public SearchResponse<ObjectNode> searchDatasets(String query) throws IOException {
+    String[] fields = {
```
wslulciuc, line 77: _Minor_: Can we make these fields a class-level constant?

```
private static final String[] FIELDS = [...]
```

api/src/main/java/marquez/service/SearchService.java
```diff
@@ -0,0 +1,237 @@
...
+                                    HighlighterType.of(
+                                        fn -> fn.builtin(BuiltinHighlighterType.Plain))));
+                      }
+                      return hl;
+                    }),
+        ObjectNode.class);
+  }
+
+  public SearchResponse<ObjectNode> searchJobs(String query) throws IOException {
+    String[] fields = {
```
wslulciuc, line 116: _Minor_: Similar comment as [above](https://github.com/MarquezProject/marquez/pull/2834/files#r1704687986)

api/src/main/java/marquez/service/SearchService.java
```diff
@@ -0,0 +1,237 @@
...
+        s ->
+            s.index("datasets")
+                .query(
+                    q ->
+                        q.multiMatch(
+                            m ->
+                                m.query(query)
+                                    .type(TextQueryType.PhrasePrefix)
+                                    .fields(Arrays.stream(fields).toList())
+                                    .operator(Operator.Or)))
```
wslulciuc, line 99: Minor: We'll want to support operators as part of the search query as mentioned [here](https://github.com/MarquezProject/marquez/pull/2834/files#r1704682610). But, not a blocker on the first pass.

api/src/main/java/marquez/service/SearchService.java
```diff
@@ -0,0 +1,237 @@
...
+                return hl;
+              });
+          return s;
+        },
+        ObjectNode.class);
+  }
+
+  public void indexEvent(@Valid @NotNull LineageEvent event) {
+    UUID runUuid = runUuidFromEvent(event.getRun());
+    log.info("Indexing event {}", event);
```
wslulciuc, line 155: _Minor_: `info` - > `debug`

api/src/main/java/marquez/service/SearchService.java
```diff
@@ -0,0 +1,237 @@
...
+                                        "DATASET:%s:%s",
+                                        jsonMap.get("namespace"), jsonMap.get("name")))
+                                .document(jsonMap))));
+  }
+
+  private void index(IndexRequest<Map<String, Object>> request) {
+    try {
+      this.openSearchClient.index(request);
+    } catch (IOException e) {
+      log.info("Failed to index event OpenSearch not available.", e);
```
wslulciuc, line 234: _Minor_: `info` -> `error`

api/src/main/java/marquez/service/SearchService.java
```diff
@@ -0,0 +1,237 @@
...
+        new AuthScope(host),
+        new UsernamePasswordCredentials(searchConfig.getUsername(), searchConfig.getPassword()));
+    final RestClient restClient =
+        RestClient.builder(host)
+            .setHttpClientConfigCallback(
+                httpClientBuilder ->
+                    httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider))
+            .build();
+
+    JacksonJsonpMapper jsonpMapper = new JacksonJsonpMapper();
```
wslulciuc, line 58: Can we use the `ObjectMapper` defined in our [commons](https://github.com/MarquezProject/marquez/blob/main/api/src/main/java/marquez/common/Utils.java#L89) lib?

examples/airflow/docker-compose.yml
```diff
@@ -87,6 +87,7 @@ services:
     environment:
       - MARQUEZ_HOST=marquez
       - MARQUEZ_PORT=5000
+      - REACT_APP_ADVANCED_SEARCH=false
```
wslulciuc, line 90: Do we need a flag for `psql` vs `engine` based search?

api/src/main/java/marquez/service/SearchService.java
```diff
@@ -0,0 +1,237 @@
...
+import org.opensearch.client.opensearch.core.IndexRequest;
+import org.opensearch.client.opensearch.core.SearchResponse;
+import org.opensearch.client.opensearch.core.search.BuiltinHighlighterType;
+import org.opensearch.client.opensearch.core.search.HighlighterType;
+import org.opensearch.client.transport.OpenSearchTransport;
+import org.opensearch.client.transport.endpoints.BooleanResponse;
+import org.opensearch.client.transport.rest_client.RestClientTransport;
+
+@Slf4j
+public class SearchService {
```
wslulciuc, line 40: As a follow up, we'll want to make search pluggable via a `SearchService` **interface** as described in @pawel-big-lebowski proposal: [_Proposal: Search Service @ Marquez_](https://github.com/MarquezProject/marquez/blob/main/proposals/2180-search-service.md)

api/src/main/java/marquez/service/SearchService.java
phixMe: The docs say to use this `JacksonJsonpMapper` as opposed to the standard object mapper. I can give it a try though.

api/src/main/java/marquez/api/SearchResource.java
phixMe: So yes, we wish to do this, it's actually more work to support this kind of thing on the UI side than anything because you need to provide typeahead assistance in expression completion to have this experience be fluid. Once we get some feedback here about what is interesting and what subset of the search DSL is interesting to users, we can add it.

examples/airflow/docker-compose.yml
phixMe: This is the flag, but I can rename it if it is confusing.

api/src/main/java/marquez/MarquezApp.java
```diff
@@ -137,7 +137,11 @@ public void run(@NonNull MarquezConfig config, @NonNull Environment env) {
 
     final Jdbi jdbi = newJdbi(config, env, source);
     final MarquezContext marquezContext =
-        MarquezContext.builder().jdbi(jdbi).tags(config.getTags()).build();
+        MarquezContext.builder()
+            .jdbi(jdbi)
+            .searchConfig(config.getSearchConfig())
```
wslulciuc, line 142: I would make ensure `search.enabled=true`

api/src/main/java/marquez/api/OpenLineageResource.java
wslulciuc: If `search.enabled=false`, will the index call fail/error?

wslulciuc, line 71: If `search.enabled=false`, will the index call fail/error?

api/src/main/java/marquez/api/OpenSearchResource.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2018-2024 contributors to the Marquez project
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package marquez.api;
```
wslulciuc, line 6: I would move the `pkg` to `marquez.api.v2beta.*` until we promote the API to `v2`.

build.gradle
```diff
@@ -64,6 +64,8 @@ subprojects {
 
     dependencies {
         implementation "org.projectlombok:lombok:${lombokVersion}"
+        implementation 'org.opensearch.client:opensearch-rest-client:2.15.0'
+        implementation 'org.opensearch.client:opensearch-java:2.6.0'
```
wslulciuc, line 68: Move to `api/build.gradle`

api/src/main/java/marquez/api/OpenLineageResource.java
phixMe: Handling in the `indexEvent` method:
```java
  public void indexEvent(@Valid @NotNull LineageEvent event) {
    if (!searchConfig.isEnabled()) {
      log.debug("Search is disabled, skipping indexing");
      return;
    }
    UUID runUuid = runUuidFromEvent(event.getRun());
    log.debug("Indexing event {}", event);

    if (event.getInputs() != null) {
      indexDatasets(event.getInputs(), runUuid, event);
    }
    if (event.getOutputs() != null) {
      indexDatasets(event.getOutputs(), runUuid, event);
    }
    indexJob(runUuid, event);
  }
```

wslulciuc: Ok, when we follow up with an SearchService **interface**, we'll want to bind search to an engine (`psql`, `opensearch`) and can do away with the flag.

web/src/types/api.ts
```diff
@@ -282,3 +282,153 @@ export interface ColumnLineageOutEdge {
...
+type OpenSearchDatasetFacets = {
+  schema?: SchemaFacet
+  columnLineage?: ColumnLineageFacet
+}
+
+type DatasetHit = {
+  run_id: string
+  name: string
+  namespace: string
+  eventType: string
```
wslulciuc, line 427: How's `eventType` used?

wslulciuc (approved):
Amazing work, @phixMe! We definitly have some [follow up work](https://github.com/MarquezProject/marquez/pull/2834/files#r1704697626), but great to see the progress on search 💯 🚀 🥇

wslulciuc (approved):
🔍 ❤️

PR #2850, opened 2024-07-08
Add endpoint method and path to metrics name.

JDarDagran:
### Problem

Currently, in metrics endpoint there is information gathered that contains SQL Object name + method name. It might be more informative if additional information about endpoints would be added as well.

### Solution

Introduce labels (DAO name, DAO method, endpoint method, endpoint path).

### Checklist


JDarDagran:
It is breaking backwards compatibility - however, do we need it in metrics?
If so, I may suggest 2 options:
* doubling old and new names
* making new names optional via config

api/src/main/java/marquez/logging/SmarterNameStrategy.java
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright 2018-2023 contributors to the Marquez project
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package marquez.logging;
+
+import com.codahale.metrics.jdbi3.strategies.DefaultNameStrategy;
+import com.codahale.metrics.jdbi3.strategies.DelegatingStatementNameStrategy;
+
+public class SmarterNameStrategy extends DelegatingStatementNameStrategy {
```
pawel-big-lebowski, line 11: `MarquezMetricNameStrategy` ?

api/src/main/java/marquez/logging/LoggingMdcFilter.java
```diff
@@ -28,19 +32,30 @@ public class LoggingMdcFilter
...
     MDC.put(REQUEST_ID, UUID.randomUUID().toString());
     MDC.put(METHOD, requestContext.getMethod());
     MDC.put(PATH, requestContext.getUriInfo().getPath());
+    StringBuilder pathWithPlaceholders = new StringBuilder();
+
+    List<UriTemplate> matchedTemplates = uriInfo.getMatchedTemplates();
+    for (int i = matchedTemplates.size() - 1; i >= 0; i--) {
+      pathWithPlaceholders.append(matchedTemplates.get(i).getTemplate());
+    }
+    MDC.put(PATH_WITH_PARAMETERS, pathWithPlaceholders.toString());
```
pawel-big-lebowski, line 58: Can we have a test which that calls some Marquez API endpoint and then calls metrics endpoint to see if the metrics with desired name are present?

pawel-big-lebowski (commented):
Great there is a solution to this. See the comments.

api/src/main/java/marquez/logging/SmarterNameStrategy.java
JDarDagran: :+1:

api/src/main/java/marquez/logging/LoggingMdcFilter.java
JDarDagran: Added test for that.

wslulciuc:
@JDarDagran I agree that more investment is needed in our metrics, and correlating an API call `->` query is nice but adding the HTTP method feels unrelated to the metric. This metric is for a DB call, not an HTTP call, though the method invocation chain is implicit. We need consistent naming strategies for not only DAO metrics, but all metrics emitted (ex: should we add the DAO method name to the API metrics? If we go down this route, metrics would contain, so it would seem, unrelated parts). I personally, I'm not a fan of the auto-generated metric names by dropwizard.

I would favor first defining a proposal for metrics and their importance, naming strategy (ie: no longer relying on dropwizard auto-generated names), labels, etc. If we were to introduce breaking changes (and I think we have to), we need a precise metric naming strategy and list of metrics that are critical for debugging DB perf issue before moving forward.

What about adding the HTTP call as a label? This will at least unblock the PR and not introduce a breaking change and new naming strategy.

JDarDagran:
Reworked it. Now v2 endpoint `/metrics/v2` contains following metrics:
```
marquez_sql_duration_seconds_bucket{object_name="marquez.db.TagDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/tags",le="0.005",} 0.0
marquez_sql_duration_seconds_bucket{object_name="marquez.db.TagDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/tags",le="0.01",} 0.0
marquez_sql_duration_seconds_bucket{object_name="marquez.db.TagDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/tags",le="0.025",} 1.0
...
marquez_sql_duration_seconds_bucket{object_name="marquez.db.TagDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/tags",le="+Inf",} 1.0
marquez_sql_duration_seconds_count{object_name="marquez.db.TagDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/tags",} 1.0
marquez_sql_duration_seconds_sum{object_name="marquez.db.TagDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/tags",} 0.014148125
marquez_sql_duration_seconds_bucket{object_name="marquez.db.NamespaceDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/namespaces",le="0.005",} 0.0
marquez_sql_duration_seconds_bucket{object_name="marquez.db.NamespaceDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/namespaces",le="0.01",} 1.0
...
marquez_sql_duration_seconds_bucket{object_name="marquez.db.NamespaceDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/namespaces",le="+Inf",} 1.0
marquez_sql_duration_seconds_count{object_name="marquez.db.NamespaceDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/namespaces",} 1.0
marquez_sql_duration_seconds_sum{object_name="marquez.db.NamespaceDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/namespaces",} 0.005786375
```

pawel-big-lebowski (approved):

api/src/main/java/marquez/MarquezApp.java
```diff
@@ -57,7 +60,9 @@ public final class MarquezApp extends Application<MarquezConfig> {
 
   // Monitoring
   private static final String PROMETHEUS = "prometheus";
+  private static final String PROMETHEUS_V2 = "prometheus_v2";
   private static final String PROMETHEUS_ENDPOINT = "/metrics";
+  private static final String PROMETHEUS_ENDPOINT_V2 = "/metrics/v2";
```
wslulciuc, line 65: The standard is usually to expose `/metrics` for metric scraping. For example, in datadog, you set the.`openmetrics_endpoint` to point to `/metrics` to [begin scraping metrics via HTTP](https://docs.datadoghq.com/containers/kubernetes/prometheus/?tab=kubernetesadv2#simple-metric-collection).

I think exposing a separate endpoint provides the benefit of non-breaking changes, but does introduce infra updates which might be a heave lift for some orgs.

wslulciuc:
I think the metric (below) for our DB calls is a great start! But, what the metric is really trying to do is trace the HTTP call to the DB query (i.e. [span](https://opentelemetry.io/docs/concepts/signals/traces/)). I do find it confusing (as I mentioned before) as these labels are unrelated to the metric itself `endpoint_method`, etc. Though I do think this is better represented as a span, I have some suggestions that will best represent the metric:

**Given:**

```
marquez_sql_duration_seconds_sum{object_name="marquez.db.NamespaceDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/namespaces",} 0.005786375
```
**Define Metric:**

```
# The time to make the DB call for a given HTTP endpoint.
marquez_db_duration_seconds_by_http_call
```

**Labels:**

* `sql_class`
* `sql_method`
* `http_method`
* `http_path`

api/src/main/java/marquez/MarquezApp.java
JDarDagran: `v2beta/metrics` used now.

JDarDagran:
> I think the metric (below) for our DB calls is a great start! But, what the metric is really trying to do is trace the HTTP call to the DB query (i.e. [span](https://opentelemetry.io/docs/concepts/signals/traces/)). I do find it confusing (as I mentioned before) as these labels are unrelated to the metric itself `endpoint_method`, etc. Though I do think this is better represented as a span, I have some suggestions that will best represent the metric:
> 
> **Given Example Metric:**
> 
> ```
> marquez_sql_duration_seconds_sum{object_name="marquez.db.NamespaceDao",method_name="findAll",endpoint_method="GET",endpoint_path="/api/v1/namespaces",} 0.005786375
> ```
> 
> **Define Metric:**
> 
> ```
> # The time to make the DB call for a given HTTP endpoint.
> marquez_db_duration_seconds_by_http_call
> ```
> 
> **Labels:**
> 
> * `sql_class`: the name of the DAO class (replaces `object_name`)
> * `sql_method`: the name of the method called by the DAO class (replaces `method_name`)
> * `http_method`:  the name of the HTTP method for the given call
> * `http_path`: the name of the HTTP path for the given call

@wslulciuc  I did follow your advice and renamed metric and labels as per your suggestion. If you / any commiter could run CI tests (probably with `git-push-fork-to-upstream-branch` - I don't know why it doesn't trigger circleCI by default) and review, I'd really appreciate that.

wslulciuc (approved):
Thanks @JDarDagran for improving our metrics! It's long over due. We'll be investment in this area more and more as we move towards a stable release of Marquez.

PR #2852, opened 2024-07-12
Add paging on jobs panel

davidsharp7:
### Problem

Currently we pull back the last 100 runs of a job.  This PR is to allow for paging so that users can easily scroll through the
runs 10 at a time.

 image

I've also added the total run count to the returned JSON payload on this end point

```bash
GET /api/v1/namespaces/my-namespace/jobs/test-job/runs
```

Screenshot 2024-07-13 at 07 26 46

I will do the PR for dataset versions in due course.

Closes: #2614 

### Solution

Implement paging similar to that on the sidebar datasets/jobs pages.

One-line summary:

Job level paging of runs.

### Checklist


wslulciuc (approved):
LGTM 👍

davidsharp7:
> I think the duration should maybe be `N/A` in the event that the job is currently running.

Yeah agree - also added the 'New' state as well.

web/src/components/jobs/JobDetailPage.tsx
```diff
@@ -88,8 +86,9 @@ const JobDetailPage: FunctionComponent<IProps> = (props) => {
   }
   const i18next = require('i18next')
 
+  // return only latest row
   useEffect(() => {
-    fetchRuns(job.name, job.namespace)
+    fetchRuns(job.name, job.namespace, 10, 0)
     fetchJobTags(job.namespace, job.name)
   }, [job.name])
```
phixMe, line 93: Shouldn't we be setting this to a variable unless I'm missing something to page on this?

davidsharp7: It's because the "runs" are used by the job detail page so the thinking being that you pull the latest 10 runs so that job details works and then control all the paging from "Runs". Think there is a similar thing with "Dataset Versions" as that's pulled in the top level component and pushed downwards.

phixMe (commented):
I pulled down your code and the one thing I'm seeing is that the loading state is wiping out the entire page for loading the runs instead of just the run table. Would it be possible to move the loading rendering detection up the chain and only display a loading indicator for this subsection?

davidsharp7:
Yeah funnily enough that's part of the reason I split the workload between Runs and JobDetailsPage but agree it's not the best user experience to refresh everything - I will test it on a larger workload as the toy dataset won't cover these kind of use cases.

phixMe (approved):
Thanks for the updates @davidsharp7

PR #2853, opened 2024-07-12
Use `io.openlineage.server.*` pkg and add class `Metadata`

wslulciuc:
This PR removes the OpenLineage models that were first introduced in the PoC in favor of `io.openlineage.server.*` defined specifically for consumers. This PR also introduces classes `Metadata` and `VersionId`. Below, we outlined more specifically on their importance and usage.

#### class `Metadata`

Wrapper class for parsing:

* class `OpenLineage.RunEvent`
* class `OpenLineage.JobEvent`
* class `OpenLineage.DatasetEvent`

**Methods**: 

* `Metadata.Run.forEvent(OpenLineage.RunEvent):` **`Metadata.Run`**
* `Metadata.Job.forEvent(OpenLineage.JobEvent):` **`Metadata.Job`**
* `Metadata.Dataset.forEvent(OpenLineage.DatasetEvent):` **`Metadata.Dataset`**

#### class `VersionId`

Factory class for `VersionId`s.

**Methods**: 

* `VersionId.forJob():` **`VersionId`**
* `VersionId.forDataset():` **`VersionId`**

Closes: https://github.com/MarquezProject/marquez/issues/1650


api/src/main/java/marquez/api/models/Metadata.java
```diff
@@ -0,0 +1,406 @@
...
+
+    public Optional<IO> getIo() {
+      return Optional.ofNullable(io);
+    }
+  }
+
+  @Builder
+  @ToString
+  @EqualsAndHashCode
+  public static final class Dataset {
```
phixMe, line 200: I think it's worth including column lineage as well as the integration type.

wslulciuc:
[PR #2853 - Use io.openlineage.server.* pkg and add class Metadata](https://trello.com/c/EIGZaK4Y/65-pr-2853-use-ioopenlineageserver-pkg-and-add-class-metadata)

api/src/main/java/marquez/api/models/Metadata.java
wslulciuc: Given this PR has been out for a while, and the `Metadata` class isn't yet used within the codebase, I'll add column lineage facet extraction as follow up.

phixMe (approved):
Great work! This is going to improve the write path and improve interacting with and enhancing the core events with updates to the OL Spec.

PR #2854, opened 2024-07-17
version history: deduplicate the versions displayed

namyyys:
### Problem

In the version history, a dataset will sometimes have the same version displayed multiple times. 
Capture d’écran 2024-07-16 à 10 55 07

After investigating, i found out that this comes from the join clause between the dataset_versions and datasets_view in [this](https://github.com/MarquezProject/marquez/blob/fcfebc55c80b412e975d3b70e487bd9898f2097f/api/src/main/java/marquez/db/DatasetVersionDao.java#L254-L293) function. The symlinks are included in the datasets_view using the same uuid as the dataset and so are displayed in the same logic as the primary dataset. 

### Solution

- Exclude the symlinks from the result of the query to display the version history
- this will be done by creating a CTE 'dataset_symlinks_names' with all the symlinks name. Then we check that all the datasets name included in the list of version does not have a symlink name (and so is not a symlink)

For the same dataset in the above example we have now this display
Capture d’écran 2024-07-16 à 11 14 47

One-line summary:

### Checklist

### DIFF
--- a/api/src/test/java/marquez/DatasetIntegrationTest.java
+++ b/api/src/test/java/marquez/DatasetIntegrationTest.java
@@ -620,7 +620,6 @@ public class DatasetIntegrationTest extends BaseIntegrationTest {
                         .build()))
             .outputs(Collections.emptyList())
             .build();
-    System.out.println("event is " + lineageEvent);
     final CompletableFuture<Integer> resp = sendEvent(lineageEvent);
     assertThat(resp.join()).isEqualTo(201);
     List<DatasetVersion> versions = client.listDatasetVersions(NAMESPACE_NAME, DB_TABLE_NAME);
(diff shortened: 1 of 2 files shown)


api/src/test/java/marquez/DatasetIntegrationTest.java
```diff
@@ -574,4 +576,58 @@ public void testApp_doesNotShowDeletedDatasetAfterUndeleteNamespace() throws IOE
...
+            .inputs(
+                Collections.singletonList(
+                    LineageEvent.Dataset.builder()
+                        .namespace(NAMESPACE_NAME)
+                        .name(DB_TABLE_NAME)
+                        .facets(datasetFacets)
+                        .build()))
+            .outputs(Collections.emptyList())
+            .build();
+    System.out.println("event is " + lineageEvent);
```
phixMe, line 623: Remove log statement here.

phixMe (commented):
Thanks for the change, happy to merge after the log is removed!

api/src/test/java/marquez/DatasetIntegrationTest.java
namyyys: i pushed a new version removing the log statement

phixMe (approved):
Thanks for the fix! 💯 👍

PR #2859, opened 2024-07-25
[FIX] Dataset query to get only the latest facet for each version

sophiely:
### Problem

Closes: https://github.com/MarquezProject/marquez/issues/2860

### Solution

Since the same facet type is replicated a lot of times, we can rank the facet partition by dataset version and facet name ands so as we can take only the most recent facet for each dataset uuid and type.
The UI seems to display only one facet per type (facet name) and dataset version anyway so we don't need to query as much facet (which are just duplicates anyway).

### Checklist

- [] Your changes are accompanied by tests (_if relevant_)

### DIFF
--- a/api/src/main/java/marquez/db/DatasetDao.java
+++ b/api/src/main/java/marquez/db/DatasetDao.java
@@ -126,7 +126,7 @@ public interface DatasetDao extends BaseDao {
                   df.facet,
                   df."name",
                   df.created_at,
-                  rank() OVER (PARTITION BY "name"
+                  rank() OVER (PARTITION BY df.dataset_version_uuid, "name"
                                ORDER BY created_at DESC) AS r
           FROM dataset_facets AS df
           WHERE df.facet IS NOT NULL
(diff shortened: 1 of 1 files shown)


phixMe (approved):
Amazing, thanks, do we have a similar problem elsewhere? Or is this the only instance of this problem that you have seen.

api/src/main/java/marquez/db/DatasetDao.java
```diff
@@ -121,30 +121,51 @@ default void setFields(Dataset ds) {
...
-          WHERE d.namespace_name = :namespaceName
-          ORDER BY d.name
-          LIMIT :limit OFFSET :offset
-          """)
+      WITH facets_t AS
+          (SELECT df.dataset_version_uuid,
+                  df.facet,
+                  df."name",
+                  df.created_at,
+                  rank() OVER (PARTITION BY df.dataset_version_uuid, "name"
```
wslulciuc, line 129: ❤️

wslulciuc (approved):
LGTM 💯 🚀 🔥

dkt-sophie-ly:
> Amazing, thanks, do we have a similar problem elsewhere? Or is this the only instance of this problem that you have seen.

For now i don't think so but ofc I will let you know if I notice similar issues :)

PR #2866, opened 2024-08-05
Web: encode job name in API requests

dolfinus:
### Problem

After upgrading to 0.48 opening some job details lead to 404 errors.
This is because I have job names with slashes, like `my/session/name`, resulting API query to look like this:
`http://localhost:8080/api/v1/namespaces/yarn%3A%2F%2Fsomecluster/jobs/my/session/name.execute_insert_into_hadoop_fs_relation_command`

But it should be:
`http://localhost:8080/api/v1/namespaces/yarn%3A%2F%2Fsomecluster/jobs/my%2Fsession%2Fname.execute_insert_into_hadoop_fs_relation_command`

Same could be applied to dataset, tag or field names.

### Solution

One-line summary:

Urlencode job, dataset, tag and field names while sending API request.

### Checklist

- [X] You've included a one-line summary of your change for the [`CHANGELOG.md`](https://github.com/MarquezProject/marquez/blob/main/CHANGELOG.md#unreleased) (_Depending on the change, this may not be necessary_).

### DIFF
--- a/web/src/store/requests/datasets.ts
+++ b/web/src/store/requests/datasets.ts
@@ -6,9 +6,8 @@ import { Dataset, DatasetVersions, Datasets } from '../../types/api'
 import { genericFetchWrapper } from './index'
 
 export const getDatasets = async (namespace: string, limit = 20, offset = 0) => {
-  const url = `${API_URL}/namespaces/${encodeURIComponent(
-    namespace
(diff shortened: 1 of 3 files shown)


phixMe (approved):
Thanks, this is the kind of issue that we keep getting back to with every path we have. We really need some unit test coverage on this. @wslulciuc We should come up with something here.

Issue #2875, opened 2024-08-12
Column lineage query returning null for namespace and name


mattwparas:
It looks like with this change #2821, the column lineage query returns null values, which causes the get dataset endpoint to 500 since namespace and name are marked as non null. I've tested with the old query and it returns the column lineage properly. I have not yet investigated as to why

davidsharp7:
Have you got some examples we could look at?

mattwparas:
Yeah, I can paste some examples - its a big hard since we've got quite a bit of data so I'll try to distill it down to something manageable

sophiely:
Hi all ! 

I face the same issue here.
I feel like it's because the namespace and name are provided thanks to the `datasets_view` but this view only contains the latest version of a dataset. 
Sometimes the column lineage is created by an ulterior version so this version won't be find if the `datasets_view`, I fix the query so as we read in the `datasets_version` table and not the `dataset_views`.
Let me know what you think :)

PR #2881, opened 2024-08-26
[FIX] Column lineage query with null namespace and name

sophiely:
### Problem

Closes: [#2875](https://github.com/MarquezProject/marquez/issues/2875)

In some case the input dataset used seems not to be in dataset_views table so when joining the two tables it results in null for namespace and name and make the list dataset endpoint fail. 
dataset_views only contains the latest dataset version so if the column lineage is created with an old version, this old version will ne be in the dataset_view and then results in an empty namespace and name.

### Solution

We can get the namespace and name in the dataset_versions table to make sure to have all the version of a dataset.

### Checklist


wslulciuc (approved):
LGTM 💯 💯

PR #2891, opened 2024-09-05
Materialized views for lineage events graph

phixMe:
### Problem

We wish to enable counts for lineage events via a new graph on our main application page!

image

### Solution
Creates two materialized views that get updated via Crons managed through Dropwizard at respective intervals.

We are supporting `daily` and `hourly` rollups for now.

> Note: This PR does not included current timeframe rollups.

### Checklist


api/src/main/java/marquez/api/OpsResource.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Response;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import marquez.service.OpsService;
+import marquez.service.ServiceFactory;
+
+@Slf4j
+@Path("/api/v1/ops")
+public class OpsResource {
```
wslulciuc, line 24: Minor suggestions on naming:

Use endpoint `GET /api/v1/stats/<stats-type>?period=<stats-period>`; where:
* `<stats-type>` - specifies the type of stats to return.
* `<stats-period>` - specifies whether to return statistics for a particular day or week.


this would allow us to provide stats for other in the future by replacing `<stats-type>` with `lineage_events`, `run_history`, `datasets`, etc.

For lineage events stats the would power our events dashboard in our overview page, I would use the API call:

```
GET /api/v1/stats/lineage_events?period=day
```

api/src/main/resources/marquez/db/migration/V71__add_mat_views_lineage_metrics.sql
```diff
@@ -0,0 +1,23 @@
+CREATE MATERIALIZED VIEW IF NOT EXISTS lineage_metrics_hourly AS
```
wslulciuc, line 1: Using `lineage_metrics_hourly` does seem to imply that the table would have different metrics per row that can be aggregated.

I would suggest we define a more specific name for the view:

```
lineage_events_by_type_hourly_view
```

wslulciuc (approved):
We'll need to follow up with a more extensive test suite but great work! 💯 💯 💯

dolfinus:
What is represented by this plot on the image? What do these colors mean, why there are 3 on them? That are x and y axis are showing?

phixMe:
Hey, this isn't quite available in our interfaces yet, but we're working on the totality of the page. It's designed to keep track of the types of lineage events coming into the system... There are various tooltips built into the UI to describe what is going on as well as button hovers.

https://github.com/user-attachments/assets/0061ae05-e621-4412-8f43-ca5a9629e9b3

PR #2900, opened 2024-09-24
Current Lineage Metrics

phixMe:
### Problem
We currently in our materialized view lineage tables don't provide the current hour and day respectively, which is the timeframe that our materialized view did not get 

### Solution

- Separate tables for `current_hour_lineage_metrics` and `current_day_lineage_metrics`
- Deletes all table data at hourly and daily intervals as part of the cleanup. (Only retains current hour and current day)
- Inserts to simple table on OL events.

One-line summary: Includes current events in the lineage metrics aggregations.

### Checklist


api/src/main/java/marquez/api/OpenLineageResource.java
```diff
@@ -68,6 +69,11 @@ public OpenLineageResource(
   public void create(@Valid @NotNull BaseEvent event, @Suspended final AsyncResponse asyncResponse)
       throws JsonProcessingException, SQLException {
     if (event instanceof LineageEvent) {
+
+      Instant eventTime = ((LineageEvent) event).getEventTime().toInstant();
```
wslulciuc, line 73: Minor: Given [`LineageEvent.eventTime`](https://github.com/search?q=repo%3AMarquezProject%2Fmarquez%20getEventTime&type=code) usage, do we want to use `UTC`:

```
((LineageEvent) event).getEventTime().withZoneSameInstant(ZoneId.of("UTC")).toInstant()
```

api/src/main/java/marquez/db/StatsDao.java
```diff
@@ -5,68 +5,104 @@
...
+                           ON ds.start_interval = (mv.start_interval AT TIME ZONE 'UTC') AT TIME ZONE :timezone
+                 LEFT JOIN (
+            SELECT
+                DATE_TRUNC('day', ln.local_now) AS start_interval,
+                DATE_TRUNC('day', ln.local_now) + INTERVAL '1 day' AS end_interval,
+                COALESCE(SUM(CASE WHEN state = 'FAIL' THEN 1 ELSE 0 END), 0) AS fail,
+                COALESCE(SUM(CASE WHEN state = 'START' THEN 1 ELSE 0 END), 0) AS start,
+                COALESCE(SUM(CASE WHEN state = 'COMPLETE' THEN 1 ELSE 0 END), 0) AS complete,
+                COALESCE(SUM(CASE WHEN state = 'ABORT' THEN 1 ELSE 0 END), 0) AS abort
+            FROM current_day_lineage_metrics
```
wslulciuc, line 98: Minor: Do you need the cross join? You can reference the CTE directly:

```
FROM current_day_lineage_metrics, local_now
```

api/src/main/resources/marquez/db/migration/V72__add_current_lineage_metrics.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS current_hour_lineage_metrics
```
wslulciuc, line 1: Do we need a table, or can we create a view `lineage_events_by_current_hour_view`, `lineage_events_by_current_day_view `?

api/src/main/java/marquez/api/StatsResource.java
```diff
@@ -37,12 +37,12 @@ public StatsResource(@NonNull final ServiceFactory serviceFactory) {
   @GET
   @Produces(APPLICATION_JSON)
   @Path("/lineage-events")
-  public Response getStats(@QueryParam("period") Period period) {
-
+  public Response getStats(
+      @QueryParam("period") Period period, @QueryParam("timezone") String timezone) {
```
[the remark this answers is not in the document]
phixMe: I think it really needs to be mandatory because the day slices would be UTC without any guidance to the client at all. I'll make a change for this.

api/src/main/java/marquez/db/StatsDao.java
phixMe: Yeah, that is simpler for sure... Good point.

api/src/main/resources/marquez/db/migration/V72__add_current_lineage_metrics.sql
phixMe: So the idea of this current day and hour table is to not be based off another table, and have it get wiped by dw on the corresponding hour and day intervals... If we wanted to do a view I guess we would base off of `lineage_events` but that is ostensibly the largest table in the db.

api/src/main/java/marquez/api/StatsResource.java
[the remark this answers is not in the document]
wslulciuc: Makes sense! 👍

wslulciuc (approved):
💯 💯 🥇

PR #2936, opened 2024-10-21
Repurpose db-migrate to run all pending migrations

davidjgoss:
### Problem

There was no Marquez command to run all pending Flyway migrations. This is needed in the context of wanting to run migrations in a short-lived container ahead of the application deployment, so migrations are not coupled to startup. The `db-migrate` command sort of looks like it should do this, but actually only supports running a single specific Java migration which is from ~20 versions ago.
 
### Solution

This PR repurposes `db-migrate` to run all pending migrations.

The original intent was to leave the existing functionality around the V57 migration in place and support omitting the `version` argument to run all pending migrations. However, this was dependent on the command remaining an `EnvironmentCommand` which caused its own problem, because such a command by nature requires the application's `run` method to have been called, which contains the migration-on-startup logic. So the scenario would go:

1. Set `migrateOnStartup: false` in the config
2. Execute `marquez.jar db-migrate my-config.yml`
3. `MarquezApp::run` is called by Dropwizard
3. `migrateDbOrError` detects pending migrations and throws because they ostensibly aren't about to run
4. We never make it to the `db-migrate` command

So instead we have reworked the command to be a `ConfiguredCommand` which only requires the resolved configuration and bootstrap object, and can be run without the application's `run` method ever being touched.

One-line summary: Repurpose db-migrate to run all pending migrations

### Checklist


api/src/main/java/marquez/cli/DbMigrateCommand.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2018-2022 contributors to the Marquez project
```
wslulciuc, line 2: Minor: _Copyright 2018-2024_...

wslulciuc (approved):
LGTM 💯

api/src/main/java/marquez/cli/DbMigrateCommand.java
wslulciuc: Actually, we should probably check other files that have an older copyright. Let's do this as follow up instead.

PR #2984, opened 2024-11-21
encodeURIComponent namespace and name in GET (column)lineage requests

MaartenHubrechts:
### Problem
When a jobname contains special characters (such as `+`) the (column)lineage GET requests breaks because the URL used for the request isn't URL encoded. This leads to the lineage page not loading properly when clicking a job in the job pages that contains any of such special characters that break the request URL.

### Solution
Use the `encodeURIComponent` function on both the `namespace` and `name` arguments so that special characters like `+` get encoded to `%2B`. This approach is also used in the datasets and jobs requests.

### Checklist

### DIFF
--- a/web/src/store/requests/columnlineage.ts
+++ b/web/src/store/requests/columnlineage.ts
@@ -15,8 +15,9 @@ export const getColumnLineage = async (
   name: string,
   depth: number
 ) => {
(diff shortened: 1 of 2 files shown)


phixMe (approved):
Thanks for the contribution!

PR #3078, opened 2025-11-01
Use bitnamilegacy for postgres image

mkay1375:
### Problem

As bitnami changed it distribution policy, now we should use bitnamilegacy for old images


merobi-hub:
@mkay1375 would you please sign your commits to clear the DCO check?

0guban0v:
@mkay1375 this is important fix that is abandoned, although nobody rejected the change.
I encountered this while validating unrelated Helm chart fix. Would you prefer new signed PR superseding this PR, or  more durable update to bundled PostgreSQL dependency instead of relying on bitnamilegacy?

PR #3111, opened 2026-07-22
Allow commas in namespace names

### LINKED ISSUE #3110: Commas not allowed in namespace names
This is a valid JDBC URL according to Postgres documentation: `jdbc:postgresql://node1,node2,node3/accounting?targetServerType=preferSecondary&loadBalanceHosts=true`.

OpenLineage removes query parameters when sanitizing URLs, but it explicitly supports multiple hosts: https://github.com/OpenLineage/OpenLineage/blob/main/client/java/src/main/java/io/openlineage/client/utils/jdbc/GenericJdbcExtractor.java#L15

Marquez therefore should support commas in namespace names.

kalra-mohit:
Namespace values that contain a comma get rejected with `IllegalArgumentException`. Showed up in practice with SQL Server/JDBC connection-string style namespaces, which use commas to separate options — e.g. `sqlserver://host;databaseName=DB1,DB2`.

Root cause is boring: `NamespaceName`'s validation regex just never allowed `,` in the character class:

```java
Pattern.compile(String.format("^[a-zA-Z:;=/0-9_\\-\\.@+]{%d,%d}$", MIN_SIZE, MAX_SIZE));
```

Before adding it I checked whether `,` is used as a delimiter anywhere that a newly-permitted comma could break something — namespace lists in query params, `NodeId`-style token splitting, that sort of thing. `NodeId` only parses on `:` and `#`, and there's no comma-delimited namespace query param anywhere, so there's nothing for this to collide with.

```mermaid
flowchart TD
    In["Namespace value containing a comma
e.g. sqlserver://host;databaseName=DB1,DB2"]
    In --> Old["Old PATTERN: comma not in allowed char class"]
    Old --> OldOut["No match -> IllegalArgumentException"]
    In --> New["New PATTERN: comma added to allowed char class"]
    New --> NewOut["Match -> NamespaceName created"]
```

**Fix:** added `,` to `NamespaceName.PATTERN` and updated the validation error message to mention commas as an allowed character.

**Testing:** added three regression cases to `NamespaceNameTest#testValidNamespaceName` — a comma-containing name, a bare `,`, and a realistic SQL Server connection-string namespace. Confirmed all three throw before the fix and pass after:

```
./gradlew :api:test --tests "marquez.common.models.NamespaceNameTest"
```

3 failures before the fix, 15/15 passing after.

Fixes #3110.

### DIFF
--- a/api/src/main/java/marquez/common/models/NamespaceName.java
+++ b/api/src/main/java/marquez/common/models/NamespaceName.java
@@ -24,7 +24,7 @@ public final class NamespaceName {
   private static final int MIN_SIZE = 1;
   private static final int MAX_SIZE = 1024;
   private static final Pattern PATTERN =
-      Pattern.compile(String.format("^[a-zA-Z:;=/0-9_\\-\\.@+]{%d,%d}$", MIN_SIZE, MAX_SIZE));
+      Pattern.compile(String.format("^[a-zA-Z:;=/0-9_\\-\\.@+,]{%d,%d}$", MIN_SIZE, MAX_SIZE));
 
   @Getter private final String value;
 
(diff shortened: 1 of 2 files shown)


kalra-mohit:

PR #3113, opened 2026-07-22
Fix dataset_fields duplication for NULL type causing column_lineage explosion

### LINKED ISSUE #3083: [Bug] `dataset_fields` with NULL type causes explosion in `column_lineage`
## Description
I have identified a critical bug where `dataset_fields` are duplicated indefinitely if the field `type` is null (or omitted). This behavior triggers a cascading, exponential explosion of rows in the `column_lineage` table, leading to severe database performance degradation.

## Root Cause
The issue stems from the `INSERT` statement for `dataset_fields` within the PostgreSQL `ON CONFLICT` clause:

```sql
INSERT INTO dataset_fields ...
ON CONFLICT(dataset_uuid, name, type) DO UPDATE ...
```

In SQL standards, `NULL != NULL`. Therefore, if a field's `type` is provided as `NULL`, the Unique Constraint fails to identify the existing row. As a result, Marquez treats every incoming event for such fields as a "new" entry and performs an `INSERT` instead of an `UPDATE`.

## Impact: Exponential Data Explosion
This bug turns a linear data ingestion process into an exponential one.

1.  **Field Duplication:**
    - A single field (e.g., `modified_at`) can create **thousands of duplicate rows** in the `dataset_fields` table over time if the type is consistently null.

2.  **Lineage Explosion (The Multiplier Effect):**
    - When a job runs to link this duplicated input field to an output field, Marquez generates lineage edges for **ALL** duplicates.
    - **Expected:** 1 Input Field UUID → 1 Output Field UUID = **1 row** in `column_lineage`.
    - **Actual:** N Input Field UUIDs → 1 Output Field UUID = **N rows** in `column_lineage`.

3.  **Consequence:**
    - In my environment, I observed that the `column_lineage` table contained **over 1,000x more rows than necessary**.
    - Nearly **99.9% of the lineage data was garbage** caused by this duplication.
    - This massive overhead causes API timeouts when querying lineage and significantly increases database storage usage.

## Steps to Reproduce
1.  Emit an OpenLineage event for a dataset where a field's `type` is explicitly `null`.
2.  Emit the same event multiple times (e.g., 10 times).
3.  Check the `dataset_fields` table; you will see 10 duplicate rows for the same field name.
4.  Emit a `columnLineage` event involving that field.
5.  Check the `column_lineage` table; you will see 10 lineage rows created for what should be a single relationship.

## Suggested Fix
Modify the `INSERT` query in the DAO layer to handle `NULL` types explicitly.

Using **`COALESCE(:type, 'UNKNOWN')`** (or a similar default value) ensures that the `ON CONFLICT` clause correctly matches existing rows and performs an update instead of a duplicate insert.
https://github.com/MarquezProject/marquez/blob/a89b89cad27f87c90dd964c40dae8dfb3f64031f/api/src/main/java/marquez/db/DatasetFieldDao.java#L276

kalra-mohit:
`DatasetFieldDao.upsert` relies on the `(dataset_uuid, name, type)` unique constraint to turn a repeat upsert of the same field into an `UPDATE`:

```sql
INSERT INTO dataset_fields (uuid, type, created_at, updated_at, dataset_uuid, name, description)
VALUES (:uuid, :type, :now, :now, :datasetUuid, :name, :description)
ON CONFLICT(dataset_uuid, name, type) DO UPDATE SET ...
```

Problem is standard SQL never treats `NULL = NULL` as true for uniqueness/`ON CONFLICT` purposes. Plenty of OpenLineage producers don't report a column type, so any field upserted with `type = NULL` never matches its own previously-inserted row — a fresh row with a new UUID gets inserted every time. Each of those rows independently accumulates its own `column_lineage` edges, so what should be one relationship turns into N, and N grows unbounded with every event. That's the "explosion" from #3083 — the reporter saw 1000x+ row bloat.

```mermaid
flowchart TD
    U["upsert dataset_fields
(dataset_uuid, name, type = NULL)"]
    U --> C{"ON CONFLICT (dataset_uuid, name, type)"}
    C --> Old["Before: type stored as raw NULL"]
    Old --> N1["NULL <> NULL, conflict never matches"]
    N1 --> R1["New row inserted every upsert
N rows for 1 field -> N x column_lineage rows"]
    C --> New["After: type stored as COALESCE(:type, 'UNKNOWN')"]
    New --> N2["'UNKNOWN' = 'UNKNOWN', conflict matches"]
    N2 --> R2["Existing row updated
single dataset_fields row retained"]
```

**Fix:** coalesce a `NULL` type to the sentinel `'UNKNOWN'` before insert. Repeated upserts of an unknown-type field now reliably collide with the row already there. No schema/migration change, no risk to rows that already have a real type — this is the fix suggested in the issue itself. One visible side effect: a field's `type` now shows up as the literal string `"UNKNOWN"` via the API instead of `null` when nothing was reported. Updated `MarquezAppIntegrationTest#testDatasetWithUnknownFieldType` to match.

**Verification:** ran the actual `INSERT ... ON CONFLICT` SQL by hand against a real Postgres 16 instance via `psql`, outside Gradle/Testcontainers (see below for why). Two upserts of the same `(dataset_uuid, name)` with `type = NULL`: before the fix, `count = 2`; after, `count = 1`, and the returned `uuid` was identical across both calls, confirming the second call updated rather than inserted.

Correction on that verification, since I got it wrong in an earlier draft of this description and want to correct it plainly rather than leave it: I'd said the Postgres 16 instance I tested against matched "the version pinned by this repo's own tests." That's not accurate. I went and checked `api/src/test/java/marquez/PostgresContainer.java` (which backs the Testcontainers extension this PR's own new `DatasetFieldDaoTest` uses), plus `docker-compose.yml` and `docker-compose.db.yml` — they all still pin Postgres 14. There's an incomplete, unmerged `postgres-16` branch on the remote, and only two unrelated test files (`DbRetentionTest`, `StatsTest`) touch Postgres 16 today. So my manual verification used Postgres 16 while this repo's actual pinned test infrastructure runs Postgres 14 — a real version mismatch on my part. It doesn't change the fix's correctness though: the relevant SQL semantics here (`NULL` never equals `NULL` for `ON CONFLICT`/unique-constraint purposes) are identical across Postgres 14 and 16.

**Tests:**
- `DatasetFieldDaoTest` (new) — DAO-level: repeated upserts of a null-typed field resolve to the same row/UUID, the stored type becomes `"UNKNOWN"`, and real (non-null) types are unaffected.
- `MarquezAppIntegrationTest#testDatasetWithUnknownFieldType_repeatedUpsertsDoNotDuplicateField` (new) — emits the same dataset with a null-typed field 5 times, mirroring the issue's repro steps, and asserts only one field comes back, not five.
- `testDatasetWithUnknownFieldType` — updated to expect the `"UNKNOWN"` sentinel.

Couldn't run these locally through Gradle — same Testcontainers/Docker API mismatch I keep hitting in this sandbox (bundled `docker-java` defaults to API v1.32, local engine needs >= v1.40; confirmed it's pre-existing by hitting the identical failure on unmodified `RunDaoTest`). `./gradlew :api:compileTestJava` and `:api:testUnit` (118 tests) both pass. Given the local DB test gap, the manual `psql` verification above is what I'm actually leaning on to confirm correctness — would appreciate CI running the full DB-backed suite on this one.

Fixes #3083.

### DIFF
--- a/api/src/main/java/marquez/db/DatasetFieldDao.java
+++ b/api/src/main/java/marquez/db/DatasetFieldDao.java
@@ -251,6 +251,24 @@ public interface DatasetFieldDao extends BaseDao {
           """)
   List<InputFieldData> findInputFieldsDataAssociatedWithRun(UUID runUuid);
 
+  /**
+   * Upserts a {@code dataset_fields} row.
+   *
+   * <p>The {@code type} column participates in the {@code (dataset_uuid, name, type)} unique
+   * constraint used as the {@code ON CONFLICT} target below. In standard SQL, {@code NULL} is
+   * never considered equal to another {@code NULL} for uniqueness purposes, so a field with an
+   * unknown/omitted {@code type} (a legitimate, common case for OpenLineage events that don't
+   * report column types) would never match an existing row and a brand new {@code dataset_fields}
+   * row - with a new {@code uuid} - would be inserted on every single upsert. Since each distinct
+   * {@code dataset_fields} row independently accumulates its own {@code column_lineage} edges,
+   * this caused an unbounded, combinatorial explosion of duplicate rows in both {@code
+   * dataset_fields} and {@code column_lineage} for any field with a null type (see #3083).
+   *
+   * <p>To fix this without weakening the constraint, a null {@code type} is normalized to the
+   * literal sentinel value {@code 'UNKNOWN'} before insertion, so that repeated upserts of a
+   * field with no known type reliably collide with the previously-inserted row (which also stored
+   * {@code 'UNKNOWN'}) and are treated as an update instead of a fresh insert.
+   */
   @SqlQuery(
       "INSERT INTO dataset_fields ("
           + "uuid, "
@@ -262,7 +280,7 @@ public interface DatasetFieldDao extends BaseDao {
           + "description"
           + ") VALUES ("
           + ":uuid, "
-          + ":type, "
+          + "COALESCE(:type, 'UNKNOWN'), "
           + ":now, "
           + ":now, "
           + ":datasetUuid, "
--- a/api/src/test/java/marquez/MarquezAppIntegrationTest.java
+++ b/api/src/test/java/marquez/MarquezAppIntegrationTest.java
@@ -191,6 +191,11 @@ public class MarquezAppIntegrationTest extends BaseIntegrationTest {
     // (3) Create db table with invalid field type
     final Field field0 = Field.builder().name("field0").type(newFieldType()).build();
     final Field field1 = Field.builder().name("field1").type(null).build();
+    // A null/unknown field type is normalized to "UNKNOWN" on write so that repeated upserts of
+    // the same field reliably collide on the (dataset_uuid, name, type) unique constraint instead
+    // of creating a new dataset_fields row (and therefore a new column_lineage edge) every time
+    // (see #3083).
+    final Field expectedField1 = Field.builder().name("field1").type("UNKNOWN").build();
     final DatasetName datasetName = newDatasetName();
     final DbTableMeta dbTableMeta =
         DbTableMeta.builder()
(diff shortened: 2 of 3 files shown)


kalra-mohit:

PR #3114, opened 2026-07-22
Fix column lineage transformation details always appearing as null

### LINKED ISSUE #3100: Transformation details in column lineage always appear as null
Hi, 

We notice that, although, the JSON OpenLineage file has transformation information (type, subtype and some fields also description), in Marquez the transformation type and description always is null.

Let see the following example: 

> "outputs": [

        {
            "namespace": "bookstore2",
            "name": "customers_concat",
            "facets": {
                "columnLineage": {
                    "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
                    "_schemaURL": "https://openlineage.io/spec/facets/1-2-0/ColumnLineageDatasetFacet.json#/$defs/ColumnLineageDatasetFacet",
                    "fields": {
                        "id": {
                            "inputFields": [
                                {
                                    "namespace": "bookstore2",
                                    "name": "customers",
                                    "field": "id",
                                    "transformations": [
                                        {
                                            "type": "DIRECT",
                                            "subtype": "IDENTITY"
                                        }
                                    ]
                                }
                            ]
                        },
                        "customer_email": {
                            "inputFields": [
                                {
                                    "namespace": "bookstore2",
                                    "name": "customers",
                                    "field": "customer_email",
                                    "transformations": [
                                        {
                                            "type": "DIRECT",
                                            "subtype": "IDENTITY"
                                        }
                                    ]
                                }
                            ]
                        },
                        "customer_full": {
                            "inputFields": [
                                {
                                    "namespace": "bookstore2",
                                    "name": "customers",
                                    "field": "customer_email",
                                    "transformations": [
                                        {
                                            "type": "DIRECT",
                                            "subtype": "TRANSFORMATION",
                                            "description": "concat(customers.customer_name, ' - ', customers.customer_email)"
                                        }
                                    ]
                                },
                                {
                                    "namespace": "bookstore2",
                                    "name": "customers",
                                    "field": "customer_name",
                                    "transformations": [
                                        {
                                            "type": "DIRECT",
                                            "subtype": "TRANSFORMATION",
                                            "description": "concat(customers.customer_name, ' - ', customers.customer_email)"
                                        }
                                    ]
                                }
                            ]
                        },
                        "customer_name": {
                            "inputFields": [
                                {
                                    "namespace": "bookstore2",
                                    "name": "customers",
                                    "field": "customer_name",
                                    "transformations": [
                                        {
                                            "type": "DIRECT",
                                            "subtype": "IDENTITY"
                                        }
                                    ]
                                }
                            ]
                        }
                    }
                }, 
The customer_full field has transformation type, subtype and description but this is what we see in Marquez:

image

It is a bug?

_Originally posted by @irua84 in https://github.com/MarquezProject/marquez/discussions/3099_

kalra-mohit:
Marquez's `LineageEvent.ColumnLineageInputField` (used to deserialize the OpenLineage `ColumnLineageDatasetFacet`) only ever captured `namespace`/`name`/`field` — no `transformations` property existed on it at all. Meanwhile the write path, `OpenLineageDao.upsertColumnLineage`, only read `transformationDescription`/`transformationType` off the *output column* (`ColumnLineageOutputColumn`), which the [current OpenLineage spec](https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/ColumnLineageDatasetFacet.json) marks `deprecated: true`.

Modern producers report transformation details per input field instead, nested under `inputFields[].transformations[]` — exactly what's in the issue's example:

```json
"customer_full": {
  "inputFields": [
    {
      "namespace": "bookstore2", "name": "customers", "field": "customer_email",
      "transformations": [{ "type": "DIRECT", "subtype": "TRANSFORMATION", "description": "concat(...)" }]
    }
  ]
}
```

Since the model had nowhere to deserialize that array into, and the write path never looked at the deprecated output-column fields for producers that only emit the new shape, `transformation_description`/`transformation_type` landed as `null` for anyone using the current facet format — reproducing the bug as reported (a second user hit the same thing in the issue thread).

```mermaid
flowchart TD
    E["OpenLineage event:
inputFields[].transformations[] (current spec)"]
    E --> Old["Before: ColumnLineageInputField has no transformations field"]
    Old --> O1["Write path only reads deprecated
outputColumn.transformationDescription/Type"]
    O1 --> R1["null/null persisted whenever producer
only emits the new per-field format"]
    E --> New["After: transformationOf(inputField, outputColumn)"]
    New --> O2{"inputField.transformations present?"}
    O2 -->|yes| R2["Use that field's own transformation
description/type"]
    O2 -->|no - legacy producer| R3["Fall back to deprecated
outputColumn description/type"]
```

**Fix:**
- Added `transformations` (`List<Transformation>`) to `ColumnLineageInputField`, with a new `Transformation` class (`type`/`subtype`/`description`/`masking`) mirroring the spec. Kept the existing 3-arg constructor working unchanged so no call site needed to move — verified by compiling all 8 existing usages untouched.
- Added `OpenLineageDao.transformationOf(inputField, outputColumn)` — a small pure function that prefers the first entry of that specific input field's own `transformations` list, and falls back to the deprecated output-column-level fields for producers still on the old format.
- `upsertColumnLineage` now resolves the transformation per matched input field via `transformationOf` and groups input fields by their resolved `(description, type)` before calling the existing batch `ColumnLineageDao.upsertColumnLineageRow(...)` once per group. That persists a distinct transformation per input/output edge, which `column_lineage` and the read path already supported — without touching `ColumnLineageDao`'s public signature, so `ColumnLineageDaoTest` didn't need to change either.

**Testing:** new `OpenLineageDaoColumnLineageTransformationTest` (unit, no DB) exercises `transformationOf` directly against the issue's exact scenario, plus multi-field, fallback-to-deprecated, both-formats-absent, and multiple-transformations-on-one-field cases. Did the fail-then-pass check by hand: temporarily reverted `transformationOf` to always read the deprecated fields, saw 3 of 5 tests fail with null description/type, restored the fix, all 5 passed.

Also added `OpenLineageDaoTest#testUpdateMarquezModelDatasetWithColumnLineageFacet_usesNewTransformationsFormat`, a full DB-backed test that builds a facet using only the new `inputFields[].transformations[]` format (deprecated fields deliberately left unset, matching real modern producers) and checks the persisted row has a non-null description/type.

`./gradlew :api:testUnit` passes (123 tests, including the 5 new ones). Couldn't run the Postgres-backed `OpenLineageDaoTest` locally — same Testcontainers/Docker API mismatch I keep hitting in this sandbox (bundled `docker-java` wants API v1.32, the local engine needs >= v1.40; confirmed pre-existing by hitting it on unmodified `RunDaoTest` too). `:api:compileTestJava` passes, so the new integration test at least compiles clean against the real classes. Would appreciate CI running the DB suite here.

## Known gap I'm leaving in (marked with a comment in the code)

While working on this I found a real edge case in `upsertColumnLineage` that's worth surfacing outside a code comment. When a single output column's input fields span 2+ distinct transformation groups, `upsertColumnLineageRow(...)` gets called once per group — but that method's existing implementation always returns *all* `column_lineage` rows currently persisted for that `(outputDatasetVersionUuid, outputDatasetFieldUuid)` pair, not just the rows the current call wrote. So across multiple groups, the same row can show up more than once in the `List<ColumnLineageRow>` this method returns.

It has no production impact right now: the only consumer of that return value, `DatasetRecord#getColumnLineageRows()`, isn't read anywhere in production code — only in tests, and this PR's own new test only exercises the single-group case, so it doesn't trigger the duplication. The rows actually persisted in `column_lineage` are correct either way; this only affects the in-memory list handed back up the call stack.

I left a `// NOTE:` comment right above the relevant code explaining this, in case someone adds a real consumer of `getColumnLineageRows()` later — they'd want to either dedupe or change `upsertColumnLineageRow`'s contract to return only newly-written rows. Not fixing it here since it's inert today and out of scope for the null-transformation bug this PR is actually about.

Fixes #3100.

### DIFF
--- a/api/src/main/java/marquez/db/OpenLineageDao.java
+++ b/api/src/main/java/marquez/db/OpenLineageDao.java
@@ -1094,6 +1094,23 @@ public interface OpenLineageDao extends BaseDao {
                   datasetVersionRow.getUuid(),
                   inputFieldsByTransformation);
 
+              // NOTE: when this output field has 2+ distinct transformation groups (its input
+              // fields resolve to more than one distinct (description, type) pair via
+              // transformationOf(...) above), upsertColumnLineageRow(...) below is invoked once
+              // per group. That method's implementation always returns ALL column_lineage rows
+              // currently persisted for this (outputDatasetVersionUuid, outputDatasetFieldUuid)
+              // pair - not just the rows written by the current call - via
+              // ColumnLineageDao#findColumnLineageByDatasetVersionColumnAndOutputDatasetField.
+              // So across multiple groups, the same underlying row can appear more than once in
+              // the aggregate List<ColumnLineageRow> this method returns for a given output
+              // field.
+              //
+              // As of this writing this has no production impact: the only consumer of this
+              // return value, DatasetRecord#getColumnLineageRows(), is not read anywhere in
+              // production code today (only in tests, and only the single-transformation-group
+              // case is currently exercised). If a real consumer is added later, either dedupe
+              // the combined list here or change upsertColumnLineageRow(...)/ColumnLineageDao to
+              // return only the rows written by that specific call.
               return inputFieldsByTransformation.entrySet().stream()
                   .flatMap(
                       entry ->
--- a/api/src/main/java/marquez/db/OpenLineageDao.java
+++ b/api/src/main/java/marquez/db/OpenLineageDao.java
@@ -1049,10 +1049,23 @@ public interface OpenLineageDao extends BaseDao {
                 return Stream.empty();
               }
 
-              // get field uuids of input columns related to this run
-              List<Pair<UUID, UUID>> inputFields =
+              // Match each field associated with this run to the OpenLineage input field it
+              // corresponds to (if any), along with the transformation description/type that
+              // applies to that specific input -> output edge. Prefer the (non-deprecated)
+              // per-input-field InputField#transformations entry reported by the producer;
+              // fall back to the deprecated, whole-output-column
+              // transformationDescription/transformationType for producers that only report the
+              // old format. Without this fallback/lookup, transformation details reported via the
+              // modern `transformations` array were silently dropped and always appeared as null
+              // (see #3100).
+              //
+              // Input fields are then grouped by the transformation (description, type) that
+              // applies to each of them, so that each distinct transformation is persisted with
+              // its own value while still reusing the existing batch upsertColumnLineageRow(...)
+              // API (which applies a single description/type to every input field passed to it).
+              Map<Pair<String, String>, List<Pair<UUID, UUID>>> inputFieldsByTransformation =
                   runFields.stream()
-                      .filter(
+                      .flatMap(
                           fieldData ->
                               columnLineage.getInputFields().stream()
                                   .filter(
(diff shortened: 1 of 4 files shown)


kalra-mohit:

